PR #297, opened 2026-07-30
feat(ext): metadata-retrieve over bridge manifest kinds (B-4)

scoobydrew83:
Implements backlog item **B-4** — PR-4 of the approved mini-plan [`docs/design/visual-manifest-builder.md`](https://github.com/scoobydrew83/sfdt/blob/develop/docs/design/visual-manifest-builder.md) (§PR-4), delivering the Chrome Extension Execution Plan's **P5-4** (package.xml builder) and **P5-5** (destructiveChanges.xml builder) acceptance criteria. Depends on B-3 (#295, bridge kinds `manifest.discover`/`manifest.render`, protocol 1.3), already on `develop`.

## What changed

**`extension/features/metadata-retrieve.ts`** — refactored into the extension's manifest builder with two data paths behind one UI:

- **Bridge-connected**: on open, the feature probes the bridge with `manifest.discover` (no `type` → org type list). When it answers, per-type member expansion goes through `manifest.discover` (typed), and *every* preview render goes through `manifest.render` with `{items, mode, apiVersion}`. A member-discovery failure surfaces as an error and collapses the node — never a fabricated empty tree. A slow render can't overwrite a newer selection (monotonic sequence guard).
- **Offline fallback**: when no bridge answers (offline / unauthorized / older CLI), the original worker-proxied SOAP path — `describeMetadata`/`listMetadata` + the private XML writer — is used **unchanged**, per PR #295's delegated decision. A status line always shows which path is active.
- **Destructive mode (P5-5)**: an Additive | Destructive toggle. Destructive previews the `destructiveChanges.xml` + empty `package.xml` pair (both textareas), shows a prominent `role="alert"` warning ("deploying this pair DELETES…") with a link to the `SFDT_DESTRUCTIVE_TIMING` docs (same URL the GUI builder uses), downloads both files together, and disables **Retrieve Zip** (a destructive manifest is deploy-time input). Works on both paths — bridge via `mode: 'destructive'`, offline via the kept private writer.
- **Per-org persistence (P5-4 AC-3)**: selections (wildcards + explicit members) persist in `chrome.storage.local` keyed by Salesforce host, restored on reopen (members seed the tree exactly like a package.xml import), plus a **Clear all** button that also wipes the stored key.
- **Live preview (P5-4 AC-2)**: re-renders on every tick; whole-type tick → `*` wildcard, any child ticks → explicit members.
[...]

PR #298, opened 2026-07-30
fix(ext): member-restore truncation + redundant renders in metadata-retrieve

scoobydrew83:
Follow-up to #297 (B-4), fixing two findings the **CI code-review bot posted on #297 about 90 seconds after the squash merge** (`0efec92`), plus two non-blocking notes from this PR's own review (#298 review at `36ce3c3`).

## Fix 1 — restoring a member-level selection permanently truncated that type's member list (functional defect)

`restoreSelections()` seeds the persisted member into `match.childXmlNames`, but `toggleExpand()`'s fetch guard was `if (anyMeta.expanded && (!anyMeta.childXmlNames || anyMeta.childXmlNames.length === 0))` — a check for *emptiness*, not *completeness*. After reopening the tool, that type's tree therefore showed **only** the restored member, permanently: collapse/re-expand never re-fetched, and there is no refresh affordance, so the user could not see or select any other member of that type without dropping the saved selection. (Wildcard restores were never affected. The imported-`package.xml` seeding path had the same defect and is fixed with it.)

- New `membersLoaded` flag, set **only** by a successful bridge (`manifest.discover`) or SOAP (`listMetadata`) member fetch. The guard is `expanded && !membersLoaded`, so a seeded-from-storage node still fetches its real member list while a genuinely loaded node does not re-fetch.
- The restored ticks **survive the fetch**: `mergeRestoredSelection()` re-applies the pre-fetch selected member names onto the fetched list, and a previously-selected member the org no longer returns is **kept** (with a log line naming it) rather than silently discarded — not losing the user's selection is the entire point of persistence, it matches B-1's GUI builder, and silent dropping is the dangerous direction in destructive mode.
- On fetch failure `membersLoaded` stays false, so a retry re-fetches instead of leaving a partial tree; seeded children and ticks are untouched.
- Fixed on **both** the bridge and offline SOAP paths.

**Review round 2 (#298 note 1):** `restoreSelections()` no longer sets `expanded = true`. An expanded seeded node *presented* a one-member list as the type's complete membership, and the affordance that unlocked the rest was an undiscoverable collapse-then-expand — the flag was right but the UX still lied. Restored types are now **collapsed**, so the *first* expand does the real fetch. The selection stays visible on the collapsed row: the type checkbox renders `indeterminate` (partial tick, `aria-label` "…(some members selected)") whenever members are selected without the whole-type tick. Clear-all clears it; a wildcard restore is a full tick, never partial.

## Fix 2 — expand/collapse cost a `manifest.render` round-trip (efficiency)

`toggleExpand()` ended with an unconditional `onSelectionChanged()`, so every browse click re-rendered the manifest over the bridge and rewrote the persisted selection even though no selection had changed. The effective manifest selection is now signature-compared across the toggle (`selectionSignature()`, backed by the same `collectSelectedItems()` that feeds the preview and persistence), and the persist + re-render fires only when it actually moved — which the seeded-member merge from Fix 1 can legitimately do. The "live preview updates on every selection change" AC is unchanged and explicitly re-tested.

**Review round 2 (#298 note 3a):** the bridge branch set `membersLoaded = true` unconditionally after coercing a non-array `members` payload to `[]`, so a malformed `ok: true` response produced a fabricated empty tree that then **never re-fetched** — exactly what the "never a fabricated empty tree" invariant exists to prevent. A members-less response is now a failure: surfaced as an error, node collapsed, still re-fetchable, seeded children and ticks intact. The offline SOAP path has no equivalent hole (`apiSoap` throws on every transport error and SOAP fault, so a falsy `listMetadata` result genuinely means zero members) — now documented in place rather than left implicit.

## Mutation verification

Each fix was verified by reverting it in place, confirming the new tests fail, then restoring (`grep -c MUTATION` → 0 after each):

| Mutation | Result |
| --- | --- |
| Fetch guard reverted to `childXmlNames.length === 0` | **3 failed / 46 passed** — the seeded-restore tests: bridge path, offline path, "keeps a restored member the org no longer lists" |
| `onSelectionChanged()` made unconditional again | **3 failed / 46 passed** — sticky-wildcard-on-expand, "fires no manifest.render for a plain expand/collapse, even with a live selection", "still renders on every real selection change (AC preserved)" |
| `restoreSelections()` reverted to `expanded = true` | **5 failed / 50 passed** — the two seeded-restore fetch tests, the ghost-member test, "renders a restored type collapsed…", "keeps a seeded selection intact through a members-less response" |
| Members-less payload coerced to `[]` instead of failing | **2 failed / 53 passed** — "never marks the node loaded, so it stays re-fetchable (no permanent empty tree)", "keeps a seeded selection intact through a members-less response" |

_(The first row corrects the earlier body, which said 45 passed — the reviewer's count of 46 is right; the file has 49 tests at that revision.)_ Note the dedicated no-redundant-render test was deliberately strengthened to hold a live selection on another type — with nothing selected, an unconditional re-render short-circuits on the empty-items branch and the mutation escapes.

## Tests (13 new)

Round 1 (7): seeded-member restore fetches the full list and preserves the tick (bridge **and** offline), org-no-longer-lists-member is kept, a genuinely loaded list is not re-fetched, no render/storage write on a plain expand/collapse (with and without a live selection), renders still fire on every real selection change.

Round 2 (6): restored type renders collapsed and the *first* expand fetches exactly once; partial tick present with the right `aria-label` and cleared by a whole-type tick; clear-all removes the partial tick and the stored entry; a wildcard restore is a full tick, not partial; a members-less discover surfaces an error and the next expand re-fetches; a seeded selection survives a members-less response.

## Not regressed

Sticky whole-type wildcard semantics, the `renderSeq` stale-render guard tests, the two-host persistence test, single-writer sentinel assertions on the bridge path, the offline SOAP fallback, the feature manifest — **no permission changes**, no new dependencies, no catalog drift.

## Out of scope (backlog)

The reviewer's note 2 (a kept ghost member is visually indistinguishable and announced only at `info` level), plus render-error staleness hint, `URL.revokeObjectURL`, deploy-panel label association, render-vs-discover timeout asymmetry, `JSON.stringify` for `selectionSignature()`, and a test seam for the detached import file input.

## Gates (at `bbac6e6`)

| Gate | Result |
| --- | --- |
| `extension/` `npx vitest run` | 84 files, **1232 tests passed** |
| `extension/` `npx tsc --noEmit` | clean |
| `extension/` `npx eslint .` | 0 errors, 0 warnings |
| `extension/` `npx wxt build` | OK — manifest `0.9.1` == `extension/package.json` |
| root `npm test` | 284 files, **4486 tests passed** |
| root `npm run check:all-contracts` | exit 0 |
| root `npm run lint` | exit 0 |
| `npm run generate:catalogs` | no drift |

Still pending human verification: unpacked-build smoke against a real org with `sfdt ui` running — specifically that a restored member selection shows as a partial tick on a collapsed type and that one expand yields the full member list with the tick preserved.


scoobydrew83 (commented):
## verdict: APPROVE

Adversarial review of `36ce3c3f1e91d59ca1464442d15fb4291f92c43a` against merge-base `0efec92`, in a detached worktree, with the explicit assumption of self-review bias (same agent wrote B-4 and this fix). I re-ran every gate, independently reproduced both mutation checks without reading the PR's table first, and wrote 8 of my own probes covering the paths the PR asserts but does not test. Both defects are genuinely fixed, not flag-fixed. Three non-blocking items below, one of which I'd like an answer on before the next extension PR.

---

### Gates (all re-run here, real numbers)

| Gate | Result | Matches PR claim |
| --- | --- | --- |
| `extension/` `npx vitest run` | **84 files / 1226 tests passed**, 0 failed (24.85s) | yes |
| `extension/` `npx tsc --noEmit` | exit 0, no output | yes |
| `extension/` `npx eslint .` | exit 0, 0 errors / 0 warnings | yes |
| `extension/` `npx wxt build` | exit 0, `.output/chrome-mv3` Σ 1.2 MB; built manifest `"version":"0.9.1"` == `extension/package.json` `0.9.1` | yes |
| root `npm test` | **284 files / 4480 tests passed**, 0 failed (65.31s) | yes |
| root `npm run check:all-contracts` | exit 0 (licenses 7 manifests / node `>=22.15.0` / auth-docs 53 files / FEATURES.json 1 entry / package-paths 117 files) | yes |
| root `npm run lint` | exit 0 | yes |
| `npm run generate:catalogs` | ran, `git status --porcelain` **empty** — zero drift | yes |

Single-file baseline for the mutation work: `test/metadata-retrieve.test.ts` = **49 tests passed**.

### Independent mutation checks (reproduced, not trusted)

**Mutation 1 — fetch guard reverted to `expanded && (!childXmlNames || childXmlNames.length === 0)`**

`Tests 3 failed | 46 passed (49)`. The three are exactly:
- `seeded member restore … > fetches the full member list on expand and keeps the restored tick (bridge path)`
- `seeded member restore … > fetches the full member list on expand and keeps the restored tick (offline path)`
- `seeded member restore … > keeps a restored member the org no longer lists rather than dropping the selection`

Failure text is the real symptom, not a proxy: the rendered tree reads `▼ApexClass` **`Gone`** and nothing else — the truncation, verbatim.

**Mutation 2 — `if (selectionSignature() !== signatureBefore) onSelectionChanged();` → unconditional `onSelectionChanged();`**

`Tests 3 failed | 46 passed (49)`. The three are exactly:
- `bridge-connected path > keeps the whole-type wildcard sticky when the type is expanded (bridge path)`
- `expand/collapse is browsing … > fires no manifest.render for a plain expand/collapse, even with a live selection`
- `expand/collapse is browsing … > still renders on every real selection change (AC preserved)`

Both mutations reverted with `git checkout --`; `git status --porcelain` empty after each, `grep -c MUTATION` → 0.

**Correction to the PR's table:** mutation 1 is **3 failed / 46 passed (49)**, not "3 failed / 45 passed". The identities of the failures match the claim exactly; only the passed count is off by one (the PR's totals imply 48 and 49 tests for the two runs — the file has 49 in both cases). Bookkeeping, not substance, but the table is wrong as written.

**Confirmed weak test, as the PR admits:** `writes no storage entry for a plain expand/collapse with nothing selected` does **not** fail under mutation 2 — with nothing selected, `persistSelections()` calls `storageRemove` on an absent key, which is indistinguishable from not writing. So that test is mutation-blind and proves almost nothing; the load-bearing one is the `even with a live selection` variant. The PR discloses this honestly, which I'll credit, but see nit 3.

### #7 — was the old sticky-wildcard test relying on the redundant render? Yes, verified

I restored `0efec92`'s version of the test file over the fixed source and ran only that test:

```
× keeps the whole-type wildcard sticky when the type is expanded (bridge path)
AssertionError: expected 0 to be greater than 0
```

The old test's *only* evidence of stickiness was the payload of the redundant render (`expect(renderCalls.length).toBeGreaterThan(0)` + per-call `items` assertion). With the redundant render gone there were no calls to inspect, so the assertion was vacuous-by-collapse. The rewrite replaces it with three independent proofs — children render ticked, the *next real* selection change carries exactly `[{ApexClass,'*'},{CustomObject,'*'}]` on an exact call count of 1, and unticking a member still narrows to `[{ApexClass,'Alpha'}]`. That is strictly stronger, and mutation 2 kills it. Claim upheld.

### What I exercised hands-on (8 probes I wrote, all pass on head; deleted afterwards)

- **package.xml-import seeding path** — the PR claims this shares the defect and is fixed, but ships **no test** for it (the repo documents why: the `<input type="file">` is created detached, `extension/test/metadata-retrieve.test.ts:393`). I got a seam by spying `document.createElement` to capture the detached input, then dispatched a real `File`/`change`. Result: import seeds `Alpha` only → collapse + expand fires exactly **1** `manifest.discover` → `Beta` appears, `Alpha` still ticked, `Beta` not. The claim is true; it is just unverified in-repo.
- **Failure semantics (#6, first half)** — bridge `manifest.discover` returning `ok:false` on the first attempt: error logged, node self-collapses, seeded child + tick intact, `membersLoaded` still false, and a single re-expand click issues discover **#2** and yields the full list with the tick preserved. Retry works.
- **Both restore paths end-to-end** — reproduced the PR's bridge and offline scenarios from storage independently, plus one it doesn't cover: *offline* org-no-longer-lists-member. `<members>Gone</members>` survives into the offline preview and `<members>Alpha</members>` stays out. Good — the merge is symmetric across paths.
- **`selectionSignature()` attacks** — see #5.
- **Offline preview freshness** — with a live selection, a plain expand leaves the locally computed XML byte-identical, and the *next* real tick still updates it. Skipping `onSelectionChanged()` leaves nothing stale on the path where the preview is computed in-page.
- Regression surface: `renderSeq` stale-render test, two-host persistence (`winForHost`), sticky-wildcard semantics, and the `SERVER-ADDITIVE-XML` / `SERVER-DESTRUCTIVE-XML` single-writer sentinel assertions (10 sites) all still present and green. `grep -c innerHTML extension/features/metadata-retrieve.ts` → **0**. Diff is 4 files: two CHANGELOGs + feature + test — no `package.json`, no `wxt.config.ts`, no `FEATURES.json`, no `generated/*`. Both CHANGELOGs place the entries under `## [Unreleased]` → `### Fixed`. Offline SOAP request construction is untouched by the diff.

### #5 — attacking `selectionSignature()`: no false negative is reachable from `toggleExpand`

The structural argument is what convinced me: `selectionSignature()` is `collectSelectedItems()` — the *same* function that feeds both `updatePreview()` and `persistSelections()`. So the gate cannot skip a render whose input differs, only one whose *string encoding* collides. I worked each case you'd worry about:

| Attack | Outcome |
| --- | --- |
| Ordering instability (merge appends a no-longer-listed member after the fetched ones) | **False positive, not negative.** Probed: storage order `Zeta,Alpha`, org returns `Alpha,Beta` → signature changes order-only → render *does* fire, with `items` exactly `[Alpha, Zeta]`, and the re-persisted storage matches the rendered payload byte-for-byte (no preview/storage drift). |
| Wildcard vs explicit collapsing to the same string | Unreachable. `*` is not a legal Salesforce `fullName`, and `restoreSelections()` routes a literal `'*'` down the wildcard branch before it can ever become a child. |
| Member added **and** removed in one toggle | Impossible by construction: `mergeRestoredSelection()` only sets `selected = true` and only pushes; it never unticks and never drops. |
| Type ticked while its only child unticked | Can't coexist — the child `change` handler sets `obj.selected = false` on untick. Even if forced, `collectSelectedItems()` short-circuits on `meta.selected`, so signature *and* manifest are both `T:*` before and after — consistent, not stale. Probed the benign version: expanding a wildcard-ticked type fires 0 renders and leaves the preview byte-identical. |
| Duplicate member names across types | Type-prefixed, and `toggleExpand` mutates exactly one type's children — a cross-type collision cannot be constructed in a single toggle. |
| Empty vs cleared | Both `''` → skip is correct; `clearAllSelections()` calls `updatePreview()` on its own path and does not route through the gate. |
| Delimiter injection — a `fullName` containing `\|` or `:` (Layout names can) | Theoretically collapsible (`A:x\|B:y` from one item vs two), but it needs a *cross-type* swap in one toggle, which `toggleExpand` cannot do. Safe here; fragile if this helper is ever reused for a broader diff. See nit 3. |

I could not construct a missed re-render. The gate is placed correctly (signature captured before `expanded` flips; `collectSelectedItems` doesn't read `expanded`).

### #4 — keeping a selected member the org no longer lists: **correct**, and better than the precedent

I checked what the rest of the system does in exactly this situation:

- **GUI builder (B-1, `gui/src/pages/ManifestBuilder.jsx`)** keeps `sel.members` (a `Map<type, Set<member>>`, persisted to `localStorage`) **completely independent** of the fetched `membersByType`. `selectionToItems()` emits every persisted member regardless of whether the org still lists it — and does so **silently, with no log line and no marker**. So the GUI already keeps ghosts; this PR matches that behaviour and additionally tells the user. Divergence would have been the defect.
- **`renderPackageXml` (`src/lib/metadata-mapper.js:181`)** is a pure emitter — sort, print, no existence validation. The single writer will happily emit a member that doesn't exist, so dropping it in the extension would make the extension's manifest *differ* from `sfdt manifest` / the GUI for the same selection. That's a single-writer-parity violation, which is the more serious of the two failure modes.
- **Does it break `retrieve` / `deploy validate`?** Retrieve is tolerant (missing components come back as warnings). A source deploy against a manifest naming a member with no local file needs `allowMissingFiles` / ignore-warnings — and the extension's deploy panel already exposes **Allow Missing Files** and **Ignore Warnings** checkboxes, so this is an anticipated, user-controllable state rather than a dead end.
- **Destructive mode moves the calculus further toward keeping.** Silently dropping a member in destructive mode is the actively dangerous direction: the user would believe they had queued a deletion that isn't in the pair. Keeping it (and it failing loudly at deploy time as "not found") is the safe failure.

Verdict: **correct**, on the strongest available grounds — it preserves the persistence guarantee, matches B-1, and preserves single-writer parity. The residual weakness is presentational, see nit 2.

---

### Non-blocking items

**1. The reopened tree still *presents* a truncated list as complete.** `restoreSelections()` sets `match.expanded = true`, so after reopening the tool the type renders as `▼ ApexClass / Alpha` — one member, fully expanded, no hint the list is partial. The PR's own tests assert this state (`toContain('Alpha')` / `not.toContain('Beta')` right after `onActivate`). The permanent truncation is genuinely gone — collapse-then-expand now fetches, which I verified on both paths — but the affordance that unlocks it is *hidden*: the user has to collapse a node that already looks complete. I don't think this blocks (the "forever" clause of the reported defect is fixed, and eagerly fetching every restored type at startup is a real cost the PR is right to avoid), but the cheapest fix is one line — have `restoreSelections()` leave `expanded = false`, since the restored ticks are already announced by the `Restored N saved selection(s)` log line and already visible in the preview XML. Then the first expand click fetches, and no misleading state exists. Worth a follow-up.

**2. A kept ghost member is visually indistinguishable from a real one, and the log line repeats.** `mergeRestoredSelection()` pushes it into the tree with the same shape as a fetched member, so the only signal is an `addLog('info', …)` line in the Execution Log — the quietest level available (`addLog` has `error` / `success` / `working` / else-info; there is no warning level). It also re-fires on every expand of that type. Given `docs/design/visual-manifest-builder.md` §"Invariants" insists degraded org behaviour must reach "a visible error state", `info` is arguably too quiet for "this component no longer exists in your org". A `(not in org)` suffix on the row, or an `error`-level line, would land better.

**3. Two small robustness/testing points on the new code:**
- `membersLoaded = true` is set **unconditionally** in the bridge branch, after `const members = Array.isArray(data.members) ? data.members : []`. So an `ok:true` response missing `members` fabricates an empty tree that now **never re-fetches** — I probed this and confirmed the discover count stays at 1 across collapse/expand, where the pre-PR guard would have retried. That needs a CLI contract violation to reach, so it's low risk, but it's exactly the "never a fabricated empty tree" line in the design doc, and the fix is to gate the flag on `Array.isArray(data.members)`. (I also probed the offline analogue — a falsy `listMetadata` result — and there the sticky-empty state is *correct*: `apiSoap` throws on every transport error and SOAP fault, so a falsy return really does mean "zero members".)
- `selectionSignature()` joining on `:`/`|` is safe here only because of a property of the call site, not of the function. `JSON.stringify(collectSelectedItems())` costs nothing and removes the class of bug entirely.
- The **package.xml-import fix is untested** (a testing nit, not a correctness one — I verified it by hand as described above). The repo's existing note says there's no seam; a `createElement` spy to capture the detached file input is a viable one if you want the coverage. Also, the persist half of fix 2 is only observable through `manifest.render` on the bridge path — there's no test that a plain expand on the **offline** path doesn't rewrite storage.

### Nits

- PR table: mutation 1 is `3 failed / 46 passed`, not `3 failed / 45 passed`.
- `loadFromPackageXml()`'s reset now clears `membersLoaded` for *every* type, including ones genuinely loaded earlier in the session, so a post-import expand re-fetches them. That's consistent (children are cleared too) and correct — just noting it's an intentional extra round-trip, not an oversight.
- `restoreSelections()` silently drops stored entries whose type isn't in the current type list (`if (!match) continue`), and with fix 2 those entries now survive in `chrome.storage` until the next *real* selection change instead of being pruned by the first browse click. No user-visible effect (the manifest is built from the in-memory model), and arguably better — degraded discovery no longer eats saved selections.

Scope is clean: `extension/features/metadata-retrieve.ts`, `extension/test/metadata-retrieve.test.ts`, and the two CHANGELOGs. Nothing else. Human smoke against a real org with `sfdt ui` remains the one gate I can't run, and it's correctly still flagged as pending — specifically worth confirming nit 1 there, since it's a look-and-feel question a test can't settle.


---

PR #303, opened 2026-07-30
feat(ext): SOSL mode in soql-runner (P4-3)

scoobydrew83:
> **This PR also fixes a shipped break.** The runner on `develop` typed the REST Search reply as an array and spread it — the real reply is `{searchRecords: […]}`, so `[...envelope.records]` throws `TypeError`. **Every `FIND` query in the shipped SOQL Runner is a hard failure today.** Reviewer independently executed the old code path on the real reply shape and confirmed the throw. So this is "SOSL works at all", not just "SOSL gets a mode".

## What changed

The extension's SOQL Runner gains a **SOQL | SOSL** mode toggle, and SOSL results render **grouped per returned sObject**, each group carrying the existing shared result toolbar.

**`extension/features/soql-runner.ts`**
- New exported pure helpers: `isSoslQuery` / `detectQueryLang`, `searchRecordsFrom`, `groupSearchRecords`, `soslRowCount`, `entryLang`, plus `QueryLang` / `SoslGroup` types.
- `runSearch(api, sosl)` — worker-proxied `GET /services/data/vXX.X/search?q=…` → `SoslGroup[]`.
- The language toggle is a real **radiogroup**: `role="radiogroup"` + `aria-label="Query language"`, two `role="radio"` buttons with `aria-checked`, roving `tabindex`, and arrow-key selection.
- **One toolbar definition.** `createResultActions({getRecords, filePrefix, scope})` builds Copy CSV / Export CSV / Copy JSON / Copy for Excel and is used **twice**: once for the flat SOQL footer (replacing the four previously hand-rolled buttons and their handlers) and once per SOSL group. The P1-3 serialisers `recordsToCsv` / `recordsToJson` / `recordsToTsv` remain the only ones in either path. The result table builder (`buildRecordsTable`) is shared the same way, so record-Id cells and the 200-char clamp behave identically in both views.
- History / bookmarks / the Saved SOQL hand-off all carry `lang`, and restoring one restores the mode.
- **Transport is preserved across a SOSL round trip** and **an explicit toggle click wins over auto-detect** — see "Review follow-ups" below.

**`extension/features/saved-soql.ts`** — each row is badged `SOQL`/`SOSL` and stages the language with the pending query.

**`extension/CHANGELOG.md`** — entry under `[Unreleased]`.

**Docs site** — companion PR on `scoobydrew83/sfdt-site` branch `docs/p4-3-sosl-mode` (see below).

## Review follow-ups (commit `9c92d87`)

Both review findings were "silent wrongness" rather than rough edges, so both are fixed here rather than deferred.

**A — the Tooling transport was silently spent by a SOSL round trip.** `setLang('sosl')` used to call `setMode('rest')`, and nothing remembered the prior choice. Because `syncLangFromText()` runs on every keystroke, a Tooling user who merely *pasted* a `FIND` query to look at it came back to SOQL sitting on REST with no indication.

The fix separates *choice* from *transport in effect*:
- `mode` is now strictly the user's **SOQL transport choice** and is never written by the language toggle.
- `effectiveMode()` (`lang === 'sosl' ? 'rest' : mode`) is what runs, and is what history/bookmarks record — so a SOSL entry honestly says `api: 'rest'` because that is what executed, without that being mistaken for a preference.
- In SOSL the **whole** transport control reads as unavailable: both REST and Tooling are genuinely `disabled` (not hidden), titled *"Not available in SOSL mode — the Search resource is REST-only. Your SOQL choice is restored when you switch back."* The control paints REST while there, because REST **is** the transport in effect — then comes back to Tooling on the way out.
- `setMode()` ignores activations while SOSL is selected, rather than recording a change the user can't see take effect.
- Restoring a stored entry now sets the **language first** (which re-enables the control) and the transport **only for SOQL entries** — so restoring a SOSL entry can't clobber a pending Tooling choice.

**B — auto-detect could revert a deliberate click.** With a leftover `SELECT` in the editor, choosing SOSL survived only until the next keystroke.

`setLang()` now latches an explicit choice (`langExplicit`) together with **what the text detected as at that moment** (`langChoiceBaseline`). `syncLangFromText()` re-detects only when the query's own language has moved *off that baseline*:

| State | Next event | Result |
| --- | --- | --- |
| `SELECT …` in editor, user clicks **SOSL** | edits that same SELECT | **stays SOSL** — the human's signal outranks inference |
| same | pastes a real `FIND {…}` | stays SOSL, latch releases (detection now agrees on its own merits) |
| latch released | types a `SELECT` | back to SOQL automatically |
| any | text is empty / `FIN` / `SELEC` | **unchanged** — ambiguous text settles nothing |

The detection function itself is untouched, so the regex parity with `validateLocal()` and the GUI's `isSosl` that the review verified character-for-character is unchanged; only *when it is consulted* changed.

**Nits also taken:** the arrow-key handler computes its target from the **current selection** rather than the focused button's own value (correct if a third language is ever added or focus is moved programmatically); the per-group export filename sanitises the org-supplied sObject name (`[^A-Za-z0-9_]+` → `_`) — the one place org data reached a filesystem name; and the transport buttons gained `aria-pressed`, since their selected state was conveyed by colour alone (partially addressing finding C — the fuller retrofit stays P6-3's job).

**9 new tests** pin all of it, including that a SOSL run really goes over `/search` while a Tooling choice is pending (`toolingQuery` not called), that the subsequent SOQL run goes over Tooling without a re-click, and that restoring a Tooling SOQL entry *from* SOSL mode works.

Nothing the review verified was disturbed: the footer's CSV bytes and `soql-<stamp>.csv` filename are untouched (that code path did not change), per-group scoping is untouched, and Tooling/Explain are still genuinely `disabled` rather than hidden — now with REST disabled alongside Tooling, and asserted as still visible.

## Design decision: the toggle is explicit **and** auto-detects `FIND {`

Stated per the task: **both**, matching the GUI console rather than diverging from it — with the explicit choice given precedence (finding B above).

`gui/src/pages/SoqlConsole.jsx` has no toggle at all — `isSosl = /^\s*find\b/i` routes a FIND query to the SOSL endpoint without asking. A toggle that ignored the text would mean the same query behaves differently in the two surfaces, which is exactly the "parallel model" to avoid. So the toggle is the explicit control the AC asks for (and the thing history records), and typing/pasting a query whose first keyword is `FIND` flips it to SOSL, `SELECT` flips it back — unless the user has said otherwise about *this* query.

## How it matches the CLI / GUI semantics

| Concern | CLI (`src/lib/soql-runner.js`) | GUI (`gui/src/pages/SoqlConsole.jsx`) | This PR |
| --- | --- | --- | --- |
| Detecting SOSL | `validateLocal()`: `kind='sosl'` on `/^find\b/i` (after trim) | `isSosl = /^\s*find\b/i` | `isSoslQuery = /^\s*find\b/i` — word-boundary, so `Finder__c` is not SOSL |
| Transport | `sf data search --query` → `result.searchRecords` | posts to the CLI's `soqlSosl` endpoint (same runner) | `sfApiFetch`-proxied `GET …/search?q=` — the REST resource the CLI's `sf data search` wraps. Semantic parity, not code sharing (as scoped). |
| Row shape | `searchRecords.map(stripAttributes)` | server-side, same | `groupSearchRecords` strips the `attributes` envelope per row |
| Plan / EXPLAIN | `explainQuery` is SOQL-only; `validateQuery` warns "org round-trip validation covers SOQL only" | **Plan button disabled** when `isSosl(query)` | Explain disabled in SOSL mode |
| Tooling variant | n/a — `runSearch` takes no `--use-tooling-api` | n/a | Transport control disabled in SOSL; the user's SOQL choice preserved |

**Deliberate divergence (one):** the CLI returns SOSL rows **flat** (`records: [...]`) because it prints an envelope; this PR groups them because the AC requires grouping and a table needs headings. The grouping is derived from `attributes.type`, which the CLI discards *after* it has served its purpose — no new notion of "object" is invented.

**Bound/limit handling:** the CLI clamps rows via `config.soql.defaultLimit`/`maxLimit` because it is a scriptable, unattended surface. The extension has no such config and never had one for SOQL either — its bound is the interactive `PAGE_CAP`/`Load more` model, which does not apply to SOSL (the Search resource has no `queryMore`). A user's explicit trailing `LIMIT n` in the SOSL text is passed through untouched, same as SOQL. Not a divergence in semantics — a surface that has no bound to inherit.

**The shipped break:** `runQuery`'s pre-existing FIND branch typed the `/search` reply as `Record<string,unknown>[]` and returned it as `records` directly. The real reply is `{searchRecords: […]}`, so `[...envelope.records]` throws `TypeError` — every `FIND` query in the shipped runner fails hard. It now routes through `groupSearchRecords` and flattens — one SOSL mapping, two views of it — and the branch condition tightens from `startsWith('find')` to the word-boundary rule.

## AC-by-AC evidence

**AC1 — "mode toggle SOQL/SOSL"**
- `soql-runner.ts` toolbar: `langGroup` radiogroup + `setLang()`; auto-detect via `syncLangFromText()` on `input`/`keyup`, with the explicit-choice latch above.
- Tests: `exposes the language toggle as a labelled radiogroup, SOQL selected by default`, `switches language by click and by arrow key`, `auto-detects the language from the query text, matching the GUI console`, `disables Explain and the whole transport toggle in SOSL mode`, `an explicit toggle click survives further edits of the same query`, `hands control back to detection once the query itself changes language`, `an explicit SOQL choice survives edits of a leftover FIND query`, `leaves the mode alone for text that settles nothing`, `moves the selection from the current radio, whichever one has focus`, `preserves and restores the Tooling transport across a SOSL round trip`.

**AC2 — "SOSL results grouped per object with the shared result toolbar (copy/export applies per group)"**
- `renderGroupedResults()` emits one `<section aria-label="Account results">` per group; each head gets `createResultActions({scope: group.sobject})` — the same factory the footer uses.
- Tests: `runs SOSL through the search resource and renders one section per object` (section order, per-group row counts, and that Account's table does not contain Contact's row), `gives every group its own copy of the shared result toolbar` (2 of each of the four buttons, per-object `aria-label`s), `copies and exports per group, not the whole result set` (exact CSV/TSV/JSON payload for the Contact group only, plus the `sosl-Contact-…csv` filename), `hides the SOQL-only footer actions while showing SOSL results`.

**AC3 — "history/saved entries record the mode"**
- `HistoryEntry.lang`, `SavedQuery.lang`, `PendingQuery.lang`; `entryLang()` back-fills by detection for pre-existing entries; both menus badge `SOQL`/`SOSL` and restore language-then-transport.
- Tests: `records the language on history entries and dedupes per language`, `round-trips the language through the saved-query and pending-query stores`, `records the SOSL mode in history and restores it from the history menu`, `restores the SOSL mode from a bookmark and from a staged pending query`, `restores a Tooling SOQL entry from SOSL mode, and a SOSL entry leaves the transport choice alone`, `restores SOSL for a pre-P4-3 entry that has no recorded language`, plus the two `saved-soql` tests.

**AC4 — "vitest for SOSL result mapping"**
- `soql-runner — SOSL result mapping`: multi-object grouping + first-seen order, attributes stripping, **empty** result → zero groups (never a group with an empty record list), malformed/`null`/non-array replies → zero groups, bare-array input, `Unknown` bucket for rows with no usable `attributes.type`, row counting, `runSearch` call shape, and `runQuery` flattening the same groups.

## Global DoD

| Item | Status |
| --- | --- |
| Feature-registry feature | Extension of the existing `soql-runner` feature — manifest (`id`, `name`, `contexts`, `settingsSchema`) unchanged, so registry / kill switch / options toggle / catalogs are untouched. No new opt-in toggle: this is a mode inside an existing tool, on by default like the transport toggle beside it. |
| Zero `innerHTML` | MET — `createElement` + `textContent` throughout; no `innerHTML`/`outerHTML`/`insertAdjacentHTML` in either changed file. |
| No new runtime dependencies | MET — no `package.json` change. |
| Vitest (happy-dom), per-feature pattern | MET — counts below. |
| A11y (P0-8 checklist) | Items 1–7 **N/A** (no new overlay; the runner's `present-view` host is untouched). Item 8 keyboard path: MET — radiogroup with roving tabindex and arrow keys that move selection *and* focus; every group toolbar button is a native `<button>`. Item 9: MET. Item 10: MET — group `aria-label`, per-radio `title`, per-object toolbar `aria-label`s, per-group `<section aria-label>`. Item 11: MET — `aria-checked` on the radios, and `aria-pressed` now on the transport buttons (previously colour-only); the unavailable state is a real `disabled` + explanatory `title`, and headings carry the row count as text. Item 12: MET. Item 13: unchanged. |
| Both themes | MET — every new colour is a `var(--sfdt-color-*)` token; no hex added. |
| CHANGELOG under `[Unreleased]` | MET |
| Docs-site MDX | MET — companion PR (below), including the two wording corrections from the review. |
| Telemetry invariant | Untouched — `test/telemetry.test.ts` green. |
| Secrets / SID invariant | Untouched — SOSL goes through `api.apiGet` (thin client); `test/sid-never-leaves-worker.test.ts` green. |
| Permission ledger | **No permission change.** Built manifest: `["storage","clipboardWrite","cookies","contextMenus","sidePanel"]` + 10 host patterns, identical to the ledger; the manifest source isn't in the diff. |

## Gates

Re-run in full at **`9c92d87`** from a clean worktree off `origin/develop` (`e683431`), after `npm install` + `npm run build -w packages/flow-core`.

| Gate | Command | Result |
| --- | --- | --- |
| Extension tests | `npx vitest run` (from `extension/`) | **84 files / 1269 tests passed**, 0 failed |
| Extension types | `npx tsc --noEmit` | **exit 0**, 0 errors |
| Extension lint | `npx eslint .` | **exit 0** |
| Extension build | `npx wxt build` | **success**, 1.2 MB; manifest `0.10.0` == `extension/package.json` `0.10.0` ✅ |
| Root tests | `npm test` | **284 files / 4523 tests passed**, 0 failed |
| Root contracts | `npm run check:all-contracts` | **exit 0** |
| Root lint | `npm run lint` | **exit 0** |
| Catalogs | `npm run generate:catalogs` | **zero drift** — `git status --porcelain` afterwards lists only the 3 files this commit edits |
| Docs site | `npm run build` (sfdt-site) | **success**, 64 pages indexed |

Test delta vs `develop`: `test/soql-runner.test.ts` 90 → **125** (+35), `test/saved-soql.test.ts` 8 → **10** (+2) — 37 new tests, no files added. Extension total 1232 → **1269**; root total 4486 → **4523**.

One pre-existing test was updated mechanically (`saves a bookmark when a name is supplied` now expects `lang: 'soql'` — AC3 landing), and one was renamed and widened (`disables Explain and the Tooling toggle in SOSL mode` → `…and the whole transport toggle…`, now also asserting REST is disabled, the explanatory title, and that the control is disabled rather than hidden).

## Manual steps I could not run

- **Loading `.output/chrome-mv3` unpacked against a real org.** No browser and no Salesforce org here, so the following are asserted in happy-dom only and want a human pass:
  - A real multi-object SOSL (`FIND {Acme} IN ALL FIELDS RETURNING Account(Id,Name), Contact(Id,Name), Lead(Id,Company)`) against a live org — groups match what Salesforce returns, an object with zero hits is simply absent.
  - **Copy for Excel** actually pasting as columns into a spreadsheet from a group toolbar (clipboard write is stubbed in tests).
  - **Export CSV** from a group actually downloading `sosl-Contact-<stamp>.csv` (the download anchor's `click()` is neutered in tests so happy-dom does not navigate).
  - Dark-theme rendering of the group headers, the toggle's selected/unselected states, and the **disabled** transport buttons (a disabled control still needs to read as disabled in dark).
  - Screen-reader read-out of the radiogroup and of the transport buttons' new `aria-pressed`.
- **Docs site visual check** — the site builds, but the rendered Highlights section and the `#soql-console` deep link into `/cli/dashboard` were not eyeballed in a browser.

## Companion docs PR

`scoobydrew83/sfdt-site` → branch `docs/p4-3-sosl-mode` (head `77b2b7b`): updates the `soql-runner` rows in `content/chrome-extension/features.mdx` and `content/chrome-extension/workspace.mdx` and adds a **"SOSL mode in the SOQL Runner"** Highlights section. The review's two corrections are in: the transport parenthetical now attaches to SOQL only (SOSL is called out as REST-only), and the SOQL-only LangGraph generator is no longer in the same clause as SOSL. It also picks up the precedence and transport-preservation behaviour.

Not merged, per instructions. `FEATURES.json` untouched.


scoobydrew83 (commented):
## verdict: APPROVE

Independent adversarial review. Clean detached worktree at `b4e06db` (merge-base with `develop` = `e683431`, so the diff below *is* the PR), `npm install` + `npm run build -w packages/flow-core`, every gate re-run from scratch. I did not take a single claim in the description on trust — I wrote a throwaway 45-assertion probe suite against the built feature and deleted it before the scope check.

Everything the AC asks for is there and holds under attack. Two behaviour findings below are real and reproducible but neither breaks an AC or the Global DoD, so they are judgment items, not blockers.

### Gates (my numbers, not the PR's)

| Gate | Command | Result |
| --- | --- | --- |
| Extension tests | `npx vitest run` (in `extension/`) | **84 files / 1262 tests passed**, 0 failed |
| Extension types | `npx tsc --noEmit` | **exit 0**, 0 errors |
| Extension lint | `npx eslint .` | **exit 0** |
| Extension build | `npx wxt build` | **success**, 1.2 MB; built manifest `0.10.0` == `extension/package.json` `0.10.0` ✅ |
| Root tests | `npm test` | **284 files / 4516 tests passed**, 0 failed |
| Root contracts | `npm run check:all-contracts` | **exit 0** (node policy, auth docs 53 files, FEATURES.json 1 entry, package paths 117 files) |
| Root lint | `npm run lint` | **exit 0** |
| Catalogs | `npm run generate:catalogs` | **zero drift** — `git status --porcelain` empty afterwards |

Every number in the PR description matches what I measured. (My first root run showed 285/4554 — that was my own probe file being picked up by the root vitest config; with it removed the count is exactly 284/4516.)

Scope, diffed against the merge-base: 5 files — `extension/CHANGELOG.md`, `extension/features/{soql-runner,saved-soql}.ts`, `extension/test/{soql-runner,saved-soql}.test.ts`. Nothing under `generated/`, no `FEATURES.json`, no `wxt.config.ts`, no manifest source. The "generated/ is untouched" claim is correct rather than a missed regeneration — I regenerated and got a byte-clean tree.

---

## What I exercised hands-on

### 1. "One toolbar, one set of serialisers" — verified, including the SOQL regression surface

Grepped `features/` and `lib/` for any second CSV/TSV/JSON assembly. The only `join(',')` / `join('\t')` sites are the pre-existing P1-3 `recordsToCsv`/`recordsToJson`/`recordsToTsv`, the P1-4 streaming export-all helper, and the unrelated `comparison-exporter.ts`. This PR adds none.

The footer replacement is the part that actually worried me, so I drove it: ran a SOQL query returning `{Id:'a1', Name:'Acme, Inc'}` (comma deliberately, to exercise quoting) and asserted the footer **Copy CSV** writes exactly `Id,Name\na1,"Acme, Inc"` and **Export CSV**'s download anchor is `soql-<ISO stamp>.csv` — matching `^soql-\d{4}-\d{2}-\d{2}T[\d-]+Z\.csv$`. Identical bytes, identical filename. The deltas I found are all benign or improvements:

- the four footer buttons now carry `aria-label`s and `type="button"` (they had neither);
- **Copy CSV**/**Export CSV** pick up an explicit `color: var(--sfdt-color-text-strong)` that **Copy JSON**/**Copy for Excel** already had — a consistency fix, not a regression;
- toasts singularise ("Copied 1 row as CSV" vs the old "1 rows");
- a zero-row result now hides the toolbar, where before it left stale buttons visible.

### 2. Auto-detect parity — verified against the real CLI and GUI source, then attacked

- CLI `src/lib/soql-runner.js:130` — `else if (/^find\b/i.test(text))` where `text = String(query ?? '').trim()`.
- GUI `gui/src/pages/SoqlConsole.jsx:43` — `const isSosl = (q) => /^\s*find\b/i.test(q ?? '')`, **character-for-character** what `isSoslQuery` uses. Line 152 routes `FIND` to the SOSL endpoint with no toggle; line 385 disables **Plan** on `isSosl(query)`. Both claims accurate.

17 adversarial inputs, all cross-checked against the CLI's `validateLocal` kind:

| Input | → | |
| --- | --- | --- |
| `   FIND {Acme}` / `\n\t FIND {Acme}` | SOSL | leading whitespace, tabs, newline |
| `find {acme}` / `FiNd {acme}` | SOSL | case-insensitive |
| `SELECT Id FROM Account WHERE Name = 'FIND'` | SOQL | `FIND` in a string literal doesn't fire |
| `SELECT Find__c FROM Account`, `SELECT Id FROM Finder__c`, `Find__c = 1`, `FINDER {x}` | SOQL | `\b` doesn't fire before `_` or a letter |
| `\n SELECT Id FROM Account` | SOQL | leading-newline SOQL |
| `` (empty) | SOQL | safe default |

One shared limitation: a leading `/* comment */ FIND {x}` is not detected — but *identically* in the CLI (`kind: 'unknown'`) and the GUI, and the extension is the only one of the three with a manual override, so it's strictly better than parity rather than a divergence.

**Does typing flip the mode destructively mid-edit?** Yes, in one narrow way — see finding B.

### 3. Backward compatibility — tested with genuine pre-change stored shapes

Wrote `{q: 'FIND {Legacy}', api: 'rest', ts: 1}` (no `lang` key at all) straight into history, opened the runner: the entry is badged `SOSL` in the menu and clicking it puts the toggle in SOSL. A no-`lang` **SOQL** entry with `api: 'tooling'` restores SOQL *and* keeps Tooling selected — `setLang('soql')` doesn't clobber the mode, only `setLang('sosl')` forces REST. A pre-change pending query with no `lang` opens in the detected mode. And the malformed cases: `entryLang(null)`, `entryLang(undefined)`, `entryLang({})`, `entryLang({q: null})`, `entryLang({q: 12345})` all return `'soql'` without throwing.

Dedupe is `entryLang`-aware in both directions: a pre-P4-3 `FIND {x}` and a new `{lang:'sosl'} FIND {x}` collapse to one entry (not two), while the same text under different languages stays two.

### 4. Grouping — probed every shape you'd actually hit

Record with no `attributes` → `Unknown`. `attributes: {}`, `type: ''`, `type: 42` → all `Unknown`. Empty `searchRecords`, `[]`, `null`, an error body `{message, errorCode}`, and a `{records: […]}` decoy → **zero groups**, never a fabricated empty group. Single-object result → one group, `soslRowCount` correct. Multi-object where Account and Contact **both** have a `Name` field → two groups in first-seen order with independent column sets.

Per-group export really is scoped: with an Account row and a Contact row on screen, the *Contact* group's **Copy CSV** wrote exactly `Id,Name,Email\nc1,Ann,a@x.com` — no Account row anywhere in it — and its **Copy for Excel** / **Copy JSON** matched `recordsToTsv`/`recordsToJson` of the Contact rows only. Account's `<table>` contains none of the Contact data. Headings render `Account · 1 row` (pluralised correctly), sections carry `aria-label="Account results"`, and the per-object accessible names are right (`Copy Account rows as CSV`, `Download Account rows as a CSV file`, …). Empty search → `No matches.` and zero `<section>` elements.

### 5. The latent bug — real on `develop`, and the fix is correct

Confirmed at `e683431:extension/features/soql-runner.ts`:

```ts
const results = await api.apiGet<Record<string, unknown>[]>(`…/search`, { q: soql });
return { records: results, done: true };
```

The live REST Search reply is `{searchRecords: […]}`, and `run()` then does `records = [...envelope.records]`. I executed the spread on the actual old shape: `TypeError`. So any user typing a `FIND` query into the shipped runner got a hard throw. The fix routes through `groupSearchRecords` and flattens — I verified `runQuery` on a real-shaped reply returns `[{Id:'a1'},{Id:'c1'}]` and calls `/services/data/v64.0/search` with `{q: …}`. As a bonus the branch condition tightens from `startsWith('find')` to the word-boundary rule.

### 6. Disable vs hide — genuinely disabled

`toolingBtn.disabled === true` and `explainBtn.disabled === true` in SOSL mode; a real `.click()` on Explain issues **zero** API calls. Both are `disabled === false` again after switching back. `setBusy()` also ANDs in `lang === 'sosl'`, so a busy-cycle can't re-enable Explain behind the toggle's back. The autocomplete box goes `flex → none → flex` and `runAutocomplete()` short-circuits with `autocompleteState = ''` — and restoring it to `flex` unconditionally is correct here, because that box is an always-present hint panel ("Enter query to see suggestions…"), not a popup.

### 7. a11y (P0-8), item by item

Items 1–7 (overlay/Esc/focus-trap/dialog semantics) — **N/A**, no new overlay; the runner's existing `present-view` host is untouched.

- **8 full keyboard path** — MET, verified: `role="radiogroup"` with `aria-label="Query language"`, two `role="radio"` `<button>`s, roving tabindex `[0, -1]`, and ArrowRight/ArrowLeft/Up/Down each move the selection **and** the focus (`document.activeElement` follows). Every group toolbar button is a native `<button>` in DOM order. `Home`/`End` aren't handled — correctly, they aren't part of the WAI-ARIA radiogroup pattern.
- **9 native controls first** — MET, real `<button>`s throughout.
- **10 every control labelled** — MET; group `aria-label`, per-radio `title`, per-object toolbar `aria-label`s, per-group `<section aria-label="… results">`.
- **11 state exposed not styled** — MET; `aria-checked` kept in sync in `setLang()`, heading text carries the row count (`Account · 2 rows`) rather than colour.
- **12 zero innerHTML** — MET; `grep -n "innerHTML\|outerHTML\|insertAdjacentHTML"` over both changed files → no hits.
- **13 shadow-root mounting** — unchanged.
- **14/15 tokens, both themes** — MET; zero hex literals in the changed files, and every token the new code uses (`brand`, `on-accent`, `text-strong`, `surface`, `surface-alt`, `border`, `text-weak`, `text-icon`, `surface-shade-3`) has **2** definitions in `lib/tokens.ts` — light and dark.

### 8. Global DoD

No new runtime dependencies (no `package.json` change). Permissions verified from the **built** manifest, not the description: `["storage","clipboardWrite","cookies","contextMenus","sidePanel"]` + 10 host patterns — identical to the ledger, and the manifest source isn't in the diff so it couldn't have changed. CHANGELOG entry sits under `[Unreleased]` in Keep-a-Changelog form. `test/telemetry.test.ts`, `test/sid-never-leaves-worker.test.ts` and `test/tokens.test.ts` are all inside the green 84/84. `FEATURES.json` untouched and `check-features-edits.mjs` passes.

---

## Judgment items (non-blocking)

**A. The Tooling transport selection is silently lost across a SOSL round trip.** `setLang('sosl')` does `if (sosl && mode !== 'rest') setMode('rest')`, but nothing remembers the prior mode. Reproduced: select **Tooling** → switch to SOSL → switch back to SOQL. The Tooling button is correctly re-enabled, but the selection sits on REST with no indication anything moved. Because `syncLangFromText()` runs on every keystroke, this fires for a Tooling user who merely *pastes* a `FIND` query to look at it. A `let priorMode` captured on the way into SOSL and restored on the way out is a couple of lines. The AC doesn't cover it, so I'm not blocking — but "switching back restores them" is only true of the *button*, not the *choice*.

**B. Auto-detect can override a deliberate click.** With a leftover `SELECT …` in the editor, clicking **SOSL** sticks — until the next keystroke anywhere in the textarea, at which point `syncLangFromText()` flips it back to SOQL. Reproduced hands-on. It never mis-routes a valid query (no valid SOQL begins with `FIND` and no valid SOSL begins with `SELECT`), so this is friction rather than a correctness bug, and the design decision to match the GUI is well argued. But the description's "the toggle stays the source of truth for anything the text does not settle" undersells it: an explicit user action can be reverted by text the user did not just type. Worth a follow-up thought about latching the toggle once it's been clicked, or only syncing when the leading keyword actually *changes*.

**C. Two adjacent segmented controls, only one of which has ARIA.** The new language toggle is a proper radiogroup; the REST/Tooling pair right beside it is still two bare `<button>`s with no `role`/`aria-pressed`, so a screen reader hears "SOQL radio, checked / SOSL radio / REST button / Tooling button" for what look like two identical widgets. Pre-existing and squarely P6-3's job — flagging only because this PR is what makes them neighbours. Relatedly, given finding B, an auto-flip of the mode isn't announced anywhere; a `role="status"` line would help.

**D. Companion docs PR `sfdt-site#42`** — I checked its content against the surface I just drove, not against the description. Accurate on all the load-bearing details: the toggle, the shared detection rule, per-object grouping, the four per-group toolbar buttons, the per-object export filename, mode-aware history/bookmarks with badges, and Tooling+Explain being disabled. Two small inaccuracies to fix before merge:
- `workspace.mdx` now reads "**SOQL or SOSL** (REST or Tooling)" — Tooling is disabled in SOSL, so the parenthetical reads as if it applies to both.
- `features.mdx`'s `soql-runner` row now advertises SOSL in the same sentence as the LangGraph node generator, which is SOQL-only (hidden in SOSL mode).

## Nits

- `filePrefix: \`sosl-${group.sobject}\`` interpolates a server-supplied string into a download filename. sObject API names are `[A-Za-z0-9_]` so it's safe in practice, and the `Unknown` bucket catches the malformed case — but it's the one place org data reaches a filesystem name unsanitised.
- The radio `keydown` handler computes its target from the button's own `value` (`value === 'soql' ? 'sosl' : 'soql'`) rather than from the current selection. Unreachable today — roving tabindex means Tab only lands on the selected radio and a click selects the one clicked — but it would misbehave if a third language were ever added, or if focus were moved programmatically to the unselected radio.
- `renderGroupedResults()` leaves stale group sections on screen when the user toggles back to SOQL without re-running. Harmless (the same is true of stale SOQL results today) and arguably desirable.

---

PR #304, opened 2026-07-30
fix(ext): make enabledByDefault authoritative at runtime

scoobydrew83:
## What was dead, and why it mattered

`FeatureManifest.enabledByDefault` was **declared, emitted, and never consulted**:

- `extension/lib/feature-registry.ts` declared `enabledByDefault?: boolean`
- `extension/lib/settings.ts` only *mentioned* it in a comment
- `extension/test/feature-manifests.test.ts` encoded the intended semantic (`manifest.enabledByDefault ?? true`) into the checked-in JSON and `generated/chrome-features.json`
- but `isFeatureEnabled()` ended in a bare `return true`, and `extension/test/settings.test.ts` pinned that as "defaults unknown ids to enabled"

So the flag round-tripped through the manifest, the catalog, and the parity test without ever reaching a runtime decision. **A feature could only ship on.** Declaring `enabledByDefault: false` had literally no effect, which blocks any capability that must ship off — e.g. the guarded `record-delete` in the P4-1 mini-plan. This PR is that prerequisite only; it does **not** add `record-delete`.

## How it's threaded

New `extension/lib/feature-defaults.ts` owns the `enabledByDefault ?? true` resolution, fed by two sources that the existing parity test already keeps in agreement:

1. **Seeded at import time from `extension/lib/feature-manifests.json`** — the checked-in, browser-runtime-free copy of every manifest. This is what makes the default correct in surfaces that never build a `FeatureRegistry`: notably `entrypoints/background.ts`, which gates the right-click "Inspect this record" menu through `isFeatureEnabled` with no registry in sight. A registration-only mechanism would have silently returned the wrong answer there.
2. **`FeatureRegistry.register()`** publishes each live manifest's default, so runtime-constructed features (`ui/workspace-host.ts` builds its tools from factories) and any newly added feature are honoured before the JSON is regenerated.

**Real call-site count: 12 production call sites across 6 files** (the mini-plan estimated ~14):

| File | Call sites |
|---|---|
| `extension/features/missing-description-flags.ts` | 4 |
| `extension/entrypoints/content.ts` | 3 (registry gate, palette, side-button menu) |
| `extension/features/setup-tabs.ts` | 2 |
| `extension/features/org-release-badge.ts` | 1 |
| `extension/features/context-menu-inspect.ts` | 1 (service worker) |
| `extension/entrypoints/options/main.ts` | 1 |

All 12 go through `isFeatureEnabled()`, so **none of them needed to change** — the fix lands at the single fallback point rather than being special-cased per caller. `entrypoints/options/main.ts` also writes every row's checkbox state on save, so an off-by-default feature renders unchecked and persists as an explicit `false`, which is correct.

## All 44 existing features still resolve enabled

Re-verified against `develop` before starting: all 44 entries in `generated/chrome-features.json` are `"enabledByDefault": true`, and `grep -rn "enabledByDefault:\s*false" extension/ --include=*.ts` returns 0. **No existing feature's effective default changed.**

That is now a checked invariant, not an assumption — `extension/test/feature-manifests.test.ts`, suite **"enabledByDefault is authoritative without flipping any shipped feature off"**, over the real factory manifests:

- `has no shipped feature declaring enabledByDefault: false`
- `resolves every shipped feature to enabled for a user with no stored preferences`
- `covers every feature in the checked-in manifest JSON, not just a subset`
- `seeds a default for every shipped feature from the manifest JSON`
- `still lets a user disable any shipped feature`

## Test coverage of the seed — three mutations

The suite is validated by mutation, not just by being green.

| Mutation | Effect | Result |
|---|---|---|
| **A** — `isFeatureEnabled` fallback reverted to `return true` | wiring dead | **2 tests fail** |
| **B** — JSON seed stubbed to `[]` | seeding dead | **1 test fails** |
| **C** — seed mapping changed to a constant `true` (all 44 ids still registered, flag ignored) | seed *value* dead | **2 tests fail** |

**Mutation C was found in review and originally survived the entire suite (1264/1264 passing).** The membership test asserts ids, never values, and every value-sensitive test reached the map through `registerFeatureDefault()` rather than through the JSON. Because all 44 shipped entries are `true`, **no assertion over the real manifests can distinguish a correct mapping from a constant** — so the one path this PR exists to make authoritative (the registry-free service worker, and any future ship-off feature gated from it) was invisible to the tests.

Closed by `extension/test/feature-defaults.test.ts`: a `vi.mock`'d manifest carrying an `enabledByDefault: false` entry — the one input the real corpus cannot supply. It lives in its own file because the mock is module-scoped and `feature-manifests.test.ts` needs the real seed. Re-verified on the final tree: fails under Mutation C (`honours enabledByDefault: false from the JSON seed`, `a seeded enabledByDefault:false resolves to disabled through isFeatureEnabled`), passes unmodified.

## Stored preferences and the kill switch

Precedence, highest first:

1. Explicit stored preference under the canonical kebab-case id
2. Explicit stored preference under a legacy camelCase id (`LEGACY_FEATURE_ID_MAP`)
3. The manifest's `enabledByDefault` (→ `true` when omitted, or when the id is unknown)

**A user's explicit choice always wins over the default, in both directions** — the main regression risk, pinned in `extension/test/settings.test.ts`: a stored `true` turns on a feature that ships off, a stored `false` keeps off a feature that ships on, both legacy camelCase directions, and the canonical/legacy collision case (canonical wins deterministically — `hasOwnProperty` short-circuits before the legacy loop).

**The remote kill switch is unchanged and still outranks everything.** It is applied by the `InitGate` in `initForCurrentRoute` (`!disabledRemote.has(id) && gate.isUserEnabled(id)`), one layer above `isFeatureEnabled`. `extension/test/feature-registry.test.ts` pins that a kill-switched feature stays disabled both when the user explicitly opted into an off-by-default feature and when an on-by-default feature has no stored preference.

An id with **no** manifest entry still resolves to enabled — deliberate: an unknown id is a stale/legacy storage key, not a feature asking to ship off. Failing *open* is safe only because the parity test catches manifest/JSON skew; the two are coupled, and that coupling is now stated in the `UNKNOWN_FEATURE_DEFAULT` comment. `extension/test/settings.test.ts:150` was retitled to that narrower semantic with its assertion body unchanged.

## Recorded decision: ~9 kB/bundle for the single-source seed

Importing the whole manifest array to read one boolean per feature inlines it into every entrypoint reaching settings: **+9 kB each to `background.js` (74.3 → 83 kB) and `content.js` (544 → 552.74 kB)**; total 1.20 → 1.22 MB.

Measured alternative: a derived id→boolean map is **1,002 B vs the 11,778 B source — ~8 kB/bundle recoverable**, and the gap grows with each feature added.

**Kept as-is deliberately.** That saving is 0.65% of the extension for no user-perceptible effect (8 kB of JSON parse is sub-millisecond even on an MV3 worker wake), and the alternative is a *second* generated artifact deriving from the first — another regenerate-both step and another drift guard to maintain, purely for bytes. Revisit if the manifest grows materially or a bundle budget lands, at which point the compact map should be generated from `collectEntries()` in the same `SFDT_WRITE_MANIFESTS` pass and asserted in the same parity describe, so authoring stays single-source. This rationale is recorded in `feature-defaults.ts` rather than only here.

Explicitly rejected: seeding only the ids that are `false`. That list is empty today, which would make a completely dead seed indistinguishable from a correct one — reintroducing the Mutation B hole.

## Global DoD

- **Zero `innerHTML`** — MET (no DOM code in this change)
- **No new runtime dependencies** — MET (the JSON import is an in-repo file; `package.json` / lockfile untouched)
- **No new permissions** — MET (`wxt.config.ts` and the manifest untouched; emitted manifest byte-identical to base)
- **Vitest coverage** — MET (see suites above)
- **CHANGELOG under `[Unreleased]`** — MET
- **Telemetry invariant** — untouched; no-network-egress test green
- **Secrets/SID invariant** — untouched; no code path changed around the worker boundary
- **A11y checklist / both themes** — N/A, no UI surface in this change
- **Docs-site MDX** — N/A, no user-visible behaviour change (all 44 features resolve exactly as before)
- **`generated/*` never hand-edited** — MET; regenerated via `npm run generate:catalogs`, zero drift

## Gates

Re-run in full at `0ef5eee`:

| Gate | Result |
|---|---|
| `npx vitest run` (extension) | **85 files, 1255 tests passed** |
| `npx tsc --noEmit` (extension) | **clean, exit 0** |
| `npx eslint .` (extension) | **clean, exit 0** |
| `npx wxt build` (extension) | **success, Σ 1.22 MB, finished in 2.963 s** |
| `npm test` (root) | **285 files, 4509 tests passed** |
| `npm run check:all-contracts` (root) | **all green** (license, node policy, auth docs, FEATURES.json, package paths) |
| `npm run lint` (root) | **clean** |
| `npm run generate:catalogs` | **zero drift** |

Bundle sizes unchanged from the reviewed commit (83 kB / 552.74 kB / 1.22 MB) — the follow-up is tests and comments only. Seed re-verified present in the emitted worker bundle: **44/44 ids + the `enabledByDefault` key in `.output/chrome-mv3/background.js`**.

`FEATURES.json` untouched. No `record-delete` feature added.


scoobydrew83 (commented):
## verdict: REQUEST_CHANGES

Reviewed at `f8b7e3da03e4a367a53013568254a7e8aa7e7837` in a detached worktree, `npm install` + `npm run build -w packages/flow-core` from scratch. I did not write this code.

**The implementation is correct.** I could not make it change any existing user's effective enablement, in any direction, on any surface. Every number in the PR description reproduced exactly. The one thing I am blocking on is a **coverage hole in the new suite, not a defect in the code**: the JSON-seeded path — the path that exists specifically because the service worker has no registry, and the reason this PR is a prerequisite for a ship-off feature — passes the entire suite with `enabledByDefault` ignored. One fixture closes it; details and the exact test are in §Mutation C.

---

## Gates — re-run, exact counts

| Gate | Claimed | Measured | |
|---|---|---|---|
| `npx vitest run` (extension) | 84 files, 1248 tests | **84 files, 1248 tests passed** | ✅ exact |
| `npx tsc --noEmit` (extension) | clean, exit 0 | **exit 0, no output** | ✅ |
| `npx eslint .` (extension) | clean, exit 0 | **exit 0, no output** | ✅ |
| `npx wxt build` (extension) | success, 1.22 MB, 2.199 s | **success, Σ 1.22 MB, finished in 2.934 s** | ✅ |
| `npm test` (root) | 284 files, 4502 tests | **284 files, 4502 tests passed** | ✅ exact |
| `npm run check:all-contracts` | all green | **node policy OK · auth docs OK (53 files) · FEATURES.json OK (1 entry) · package paths OK (117 files)** | ✅ |
| `npm run lint` (root) | clean | **clean** | ✅ |
| `npm run generate:catalogs` | zero drift | **`git status --porcelain` empty for `generated/` and for the whole tree** | ✅ |

Preconditions independently re-measured on this SHA: `generated/chrome-features.json` = **44 entries, 44 `enabledByDefault: true`, 0 `false`**; `lib/feature-manifests.json` = **44 entries, 0 `false`, 0 omitted, 0 duplicate ids**.

---

## Independent mutation results

### Mutation A — revert the fallback to `return true`
Replaced `return isEnabledByDefault(featureId);` with `return true;` in `lib/settings.ts:133`, ran the full extension suite.

**2 PR-authored tests fail** — matches the claim exactly:
- `test/feature-registry.test.ts` › `does not init an enabledByDefault:false feature when the user has no stored preference`
- `test/settings.test.ts` › `resolves an enabledByDefault:false feature to disabled with no stored preference`

(3 of my own scratch tests also failed; excluded from the count.) `git diff` clean after restore.

### Mutation B — stub the JSON seeding to `[]`
Replaced the seed source with an empty array (`manifests` still imported, so no unused-import noise).

**1 PR-authored test fails** — matches the claim exactly:
- `test/feature-manifests.test.ts` › `seeds a default for every shipped feature from the manifest JSON`

Notably, `resolves every shipped feature to enabled for a user with no stored preferences` **still passed** under this mutation — which is precisely the hole the PR says that test closes. The reasoning in the description is sound and the test is genuinely load-bearing. `git diff` clean after restore.

### Mutation C — the one that survives ⚠️
Changed the seed mapping from `m.enabledByDefault ?? UNKNOWN_FEATURE_DEFAULT` to a constant `true`, i.e. the JSON seeding still runs, still registers all 44 ids, but **ignores the flag it exists to read**.

```
Test Files  85 passed (85)
     Tests  1264 passed (1264)
```

**Nothing fails.** Not the new suite, not the parity test, not `tsc`, not `eslint`, not the catalog drift check.

The `seeds a default for every shipped feature` test asserts *membership* (`_knownFeatureDefaultIdsForTests()` contains every id) but never the *value*. The value-sensitive paths that do exist (`settings.test.ts`, `feature-registry.test.ts`) all reach the map via `registerFeatureDefault()`, never via the JSON. Since all 44 shipped entries are `true`, no assertion over the real manifests can distinguish a correct mapping from a constant — I tried the obvious `isEnabledByDefault(id) === (entry.enabledByDefault ?? true)` assertion and it passes under Mutation C for exactly that reason.

Why this matters rather than being pedantry: the PR's own design memo says the JSON seed exists because `entrypoints/background.ts` gates through `isFeatureEnabled` with no registry — I confirmed that (below). So for **any** future off-by-default feature gated from a registry-free surface, correctness rests entirely on the seeded value, and that is the one thing the suite cannot see. This PR's stated purpose is to unblock exactly such a feature (`record-delete`, P4-1).

**The fix is a fixture, and the implementation already passes it.** I wrote it and confirmed it fails under Mutation C and passes on unmodified HEAD:

```ts
vi.mock('../lib/feature-manifests.json', () => ({
  default: [
    { id: 'fixture-ships-off', name: 'x', contexts: [], enabledByDefault: false },
    { id: 'fixture-ships-on',  name: 'y', contexts: [], enabledByDefault: true },
    { id: 'fixture-omitted',   name: 'z', contexts: [] },
  ],
}));

it('the JSON seed honours enabledByDefault: false', async () => {
  const { isEnabledByDefault } = await import('../lib/feature-defaults.js');
  const { isFeatureEnabled, SettingsSchema } = await import('../lib/settings.js');
  expect(isEnabledByDefault('fixture-ships-off')).toBe(false);
  expect(isEnabledByDefault('fixture-ships-on')).toBe(true);
  expect(isEnabledByDefault('fixture-omitted')).toBe(true);   // omitted ⇒ true
  const fresh = SettingsSchema.parse({});
  expect(isFeatureEnabled(fresh, 'fixture-ships-off')).toBe(false);
});
```

Result on this SHA: **passes**. Under Mutation C: `AssertionError: expected true to be false`. That is the whole ask — the code is right, the guard is missing. It also directly covers the "nothing ships `false` today, so this path is otherwise untested" gap.

---

## What I exercised hands-on

**1 · "No existing feature's effective default changed" — confirmed, including the fresh-profile upgrade case.**
The PR's own test uses `SettingsSchema.parse({})`. I went further and drove the real storage layer with **no `sfdt.settings` key written at all** — the actual state of a profile upgrading into this build:

```
loadSettings() → Object.keys(s.features) === []
44/44 shipped ids → isFeatureEnabled === true   (0 resolved off)
```

Also confirmed 44/44 resolve enabled through the *import-time seed with zero `register()` calls anywhere in the test module* — i.e. the background/service-worker shape, not the content-script shape.

**2 · The load-bearing design decision — the background path really would have answered from nowhere.**
`grep -rn createFeatureRegistry` over non-test sources returns exactly three construction sites: `ui/workspace-host.ts:247`, `entrypoints/options/main.ts:201`, `entrypoints/content.ts:105`. **`entrypoints/background.ts` builds no registry** and reaches `isFeatureEnabled` at `background.ts:138 → isInspectMenuEnabled → context-menu-inspect.ts:85`. Under a registration-only design that lookup would hit an empty map for every id in the worker. (Precision on the PR's wording: today it would still *coincidentally* answer `true`, because `context-menu-inspect` omits the flag and unknown ⇒ enabled — so no shipped feature is affected. The claim holds as a design argument for any off-by-default worker-gated feature, not as a present-day bug.)

**Import-time seeding verified in the real service-worker bundle**, not just in vitest: I inspected `.output/chrome-mv3/background.js` and all **44/44 feature ids and the `enabledByDefault` key are present** in the emitted worker bundle. No dynamic `import()`, no top-level `await` — a plain module-scope `Map` construction that re-runs on each worker wake, which is the correct shape for MV3.

**Circular-import / init-ordering:** `lib/feature-defaults.ts` imports *only* `./feature-manifests.json` — it is a leaf module with no TS imports at all, so it cannot participate in a cycle. Both new edges (`settings.ts → feature-defaults`, `feature-registry.ts → feature-defaults`) point into that leaf. No TDZ or partial-module hazard is reachable. Clean.

**3 · The hole-closing test.** Reasoning verified sound (Mutation B above), and I confirmed the specific asymmetry it protects against: with seeding dead, `resolves every shipped feature to enabled` still passes. Good test. Its blind spot is value fidelity (Mutation C).

**4 · Both mutation checks reproduced** — see above; counts match the PR exactly (A: 2, B: 1). `git diff` verified clean after each restore.

**5 · Precedence — exercised in both directions, plus collision and kill switch.** All of the following I wrote and ran myself:
- stored `false` beats an on-by-default manifest ✅
- stored `true` beats an off-by-default manifest ✅
- **all 44 real shipped ids**: a stored `false` keeps each one off ✅
- legacy camelCase `setupTabs: true` beats a forced-false `setup-tabs` default ✅ **and** `setupTabs: false` beats a `true` default ✅ (the PR only pinned the first direction)
- **canonical/legacy collision**: `{'setup-tabs': false, setupTabs: true}` → `false`, and the mirror → `true`. Canonical wins, deterministically, because the `hasOwnProperty` check on the canonical id short-circuits before the legacy loop. Not order-dependent on `Object.entries`. ✅
- **kill switch outranks everything**: `InitGate` is `!disabledRemote.has(id) && gate.isUserEnabled(id)` (`feature-registry.ts:167`). A kill-switched, off-by-default feature the user explicitly opted *into* does not init; removing the kill switch on the same registry then inits it. ✅

**7 · "12 call sites, none needed changing" — confirmed, count is exact.**
`missing-description-flags.ts` 4 (`:247,:252,:253,:260`) · `content.ts` 3 (`:233,:255,:280`) · `setup-tabs.ts` 2 (`:292,:350`) · `org-release-badge.ts` 1 (`:124`) · `context-menu-inspect.ts` 1 (`:85`) · `options/main.ts` 1 (`:365`) = **12**. No bypass exists: the only direct reads of `settings.features` outside `settings.ts` are three *writes* (`missing-description-flags.ts:261`, `setup-tabs.ts:351`, `options/main.ts:525`) and two comments. `lib/palette-sources.ts` takes an injected `isEnabled` which `content.ts:255/:280` supplies from `isFeatureEnabled`. No independent default logic anywhere. ✅

**8 · Adversarial cases — all pass on this SHA:**
- registry-present / JSON-absent, `enabledByDefault: false` → resolves **off** ✅
- **a feature rejected by the permission gate still publishes its default** — `register()` calls `registerFeatureDefault` *before* the `missing.length > 0` early return. `reg.has(id) === false` but `isEnabledByDefault(id) === false`. The ordering comment in the diff is accurate and this is the right call: a feature skipped for a missing permission must not silently answer "enabled" to the options page. ✅
- JSON-present / never-registered → resolves from the seed ✅
- **corrupt stored settings** (`features: {'setup-tabs': 'yes-please'}, theme: 42`) → `safeParse` fails → `defaultSettings()` → `features === {}` → every id falls to its manifest default, *not* to blanket-on. An off-by-default feature stays off through a corrupt blob. Fail-safe in the right direction. ✅
- `__proto__` / `constructor` / `toString` as feature ids → `true`, no prototype leakage (`hasOwnProperty.call` guard + `Map` lookup) ✅
- `enabledByDefault: false` honoured end-to-end via the JSON seed → ✅ **works**, but is the untested path (Mutation C).

**9 · Global DoD + scope:**

| Item | Result |
|---|---|
| Zero `innerHTML` | ✅ `git diff \| grep -c '^+.*innerHTML'` → **0**; no DOM code in the change |
| No new runtime deps | ✅ no `package.json` / lockfile in the diff; the import is an in-repo JSON |
| No new permissions | ✅ `wxt.config.ts` untouched. I built the **base commit** and byte-compared the emitted manifests: **IDENTICAL** (`storage, clipboardWrite, cookies, contextMenus, sidePanel`; host_permissions and version unchanged) |
| CHANGELOG under `[Unreleased]` | ✅ `### Fixed`, first section, above `## [0.10.0]` |
| Telemetry / SID invariants | ✅ untouched; full extension suite (incl. no-network-egress + SID-boundary tests) green |
| `FEATURES.json` untouched | ✅ not in the diff; `check:features` green |
| No `record-delete` added | ✅ `grep -rn record-delete extension/ --include=*.ts` → 0; not in `generated/chrome-features.json` |
| Diff scope | ✅ 7 files, all `extension/` (5 source/test + CHANGELOG), +330/−10 |
| A11y checklist / both themes / docs MDX | N/A agreed — no UI surface, and no user-visible behaviour change to document (44/44 resolve exactly as before, verified) |

---

## Judgment calls

**6 · The `settings.test.ts:150` rewrite is NOT a weakening — and I want to be explicit about that, because the PR undersells it.** The diff changes the `it(...)` title from `defaults unknown ids to enabled (enabledByDefault semantics)` to `defaults an id with no manifest entry to enabled` and adds a two-line comment. **The assertion body is byte-identical**: same storage setup, same `loadSettings()`, same `expect(isFeatureEnabled(s, 'never-toggled')).toBe(true)`. Nothing was relaxed to fit the implementation — the behaviour under test is unchanged and the rename makes the *reason* legible. Correct handling.

On the underlying semantic (unknown id ⇒ enabled): I agree it is right, with one named residual. It is fail-safe for the dominant case (a stale storage key, a legacy id, a feature registered before the JSON is regenerated) — the alternative, fail-closed, would silently disable working features on any id-map skew, which is far worse. The residual: a feature that declares `enabledByDefault: false` in code, is gated from a registry-free surface, and whose entry has not yet been regenerated into `feature-manifests.json` would ship **on**. The guard against that is the parity test in `feature-manifests.test.ts`, which runs in the standard suite and fails on any manifest/JSON skew — I confirmed it is not conditional or opt-in. That's an adequate guard; worth a one-line note in the `UNKNOWN_FEATURE_DEFAULT` comment that the parity test is what makes fail-open safe, since the two are coupled but only one of them says so.

**On `register()` publishing into module-global state.** Acceptable, and consistent with the existing `_clearSettingsCacheForTests` / `_resetSettingsShapesForTests` house pattern. Each surface is its own JS realm, so the options page's registry cannot perturb the content script's map, and vitest isolates per file. No cross-talk found.

---

## Nits (non-blocking)

1. **Test-isolation flake vector.** `settings.test.ts` › `a legacy camelCase stored preference still beats the manifest default` does `registerFeatureDefault('setup-tabs', false)` and restores it to `true` on the **last line of the test body**. If the `expect` above it ever fails, the restore never runs and `setup-tabs` stays seeded `false` for every later test in the file — a failure that cascades into confusing secondary failures. The file already has an `afterEach` for `OFF_ID`/`ON_ID`; move the `setup-tabs` restore into it (or use `try/finally`).

2. **~+9 kB in every bundle for one boolean per feature.** Measured against a base build: `background.js` 74.33 → 83 kB, `content-scripts/content.js` 543.98 → 552.74 kB (total 1.20 → 1.22 MB). The whole manifest array — `name`, `contexts`, `workspace`, `sideButton`, `bridgeRequired` — is inlined into every entrypoint so that `enabledByDefault` can be read. Seeding from a minimal generated `Record<id, boolean>` (or only the ids that are `false`, defaulting the rest) would be a few hundred bytes. ~1.6% of the content script is not worth blocking over, and the single-source-of-truth simplicity has real value — but it is a cost the PR description doesn't mention, and it grows with every feature added.

3. The `_knownFeatureDefaultIdsForTests()` name promises more than the test using it delivers — it enumerates ids, and the test reads as if it verifies seeding. Renaming it, or adding the Mutation-C fixture above, removes the ambiguity.

---

**Summary:** the wiring is live (Mutation A), the seeding is live (Mutation B), precedence is correct in both directions including the legacy path, the kill switch still outranks everything, corrupt storage fails safe, all 44 features resolve identically for a genuinely fresh profile, and no permission, dependency, or generated artefact moved. The single blocker is that the seed's *value* mapping — the mechanism this PR exists to make authoritative on the one surface that has no registry — is invisible to the test suite. The fixture above is written, verified against both the mutant and HEAD, and is the only change I'm asking for.


---

PR #307, opened 2026-07-30
feat(ext): editability model + typed API errors — P4-1 PR-1 (no user-visible change)

scoobydrew83:
**PR-1 of 4** in the Record Edit / Clone / Delete chain planned in `docs/design/record-edit-clone.md` (#302), which satisfies Chrome Extension Execution Plan item **P4-1**. Dependency order is strict: `ext/fix-write-timeout` (#305) + `ext/enabled-by-default-authoritative` (#304) → **PR-1** → PR-2 → {PR-3, PR-4}. Both prerequisites are merged; this branches off the resulting `develop` (`df870e6`).

**There is no user-visible change in this PR.** It ships an unused-but-tested model — the same contract-first shape as the manifest-builder plan's PR-3 — so that the edit path built on top of it in PR-2 has one contract to obey rather than several. **It touches no feature file.**

## Why a model PR at all

`features/inspect-record.ts` already ships a rudimentary edit path, and that is the real starting position rather than a blank page. Three things about it are why the model comes first:

- **Two independent dirty-diff implementations that can disagree.** The save-bar visibility loop (`:399`) and the PATCH-body loop (`:520`) each walk `originalRecordData` comparing with raw `!==`. They happen to agree today because they are copy-paste siblings; nothing makes them agree.
- **Raw `!==` is the wrong comparison.** `10` vs `'10'` from a text input is a change. `''` vs `null` is a change. Two spellings of the same instant are a change. All three produce phantom dirty state and put fields into a PATCH body that did not need to be there.
- **One text input for every non-boolean type** (`:319`), so the question "may this field be written, and in what shape" has never had an answer anywhere.

### 1. `lib/salesforce-api.ts` — `SalesforceRestError`

A Salesforce REST rejection body names the field the org refused:

```jsonc
[{ "message": "Value too long", "errorCode": "STRING_TOO_LONG", "fields": ["Name"] }]
```

`buildRequestError()` collapsed all of that into one prose string via `extractErrorDetail()`, and `fields` / `errorCode` were dropped on the floor. The plan's **AC-1** — "server field errors render on the exact field" — is not implementable without them, and nothing else in the tree reads those attributes, which is why this one piece of API-layer work lives in P4-1 rather than being its own concern.

`buildRequestError()` now returns a `SalesforceRestError` carrying the parsed records plus `status`:

```ts
export interface SalesforceRestErrorDetail {
  message: string;
  errorCode: string;  // '' when absent — never undefined
  fields: string[];   // [] for object-level failures
}

export class SalesforceRestError extends Error {
  readonly sfdtKind: SfApiErrorKind = 'http-error';
  readonly status: number;
  readonly details: SalesforceRestErrorDetail[];
}
```

**Strictly additive.** `.message` is byte-identical, `sfdtKind` is still `'http-error'`, `.status` is unchanged, and **no call site moves** — everything that renders only `err.message` is unaffected, which is asserted by a test. `parseRestErrorDetails()` is total: a non-JSON body (a 503 HTML page), a bare object instead of an array, a record with no `message`, or a non-string entry inside `fields` all degrade to an empty or filtered list rather than throwing, so a malformed error body cannot break an error path.

**This is the only change to `salesforce-api.ts`.** The write-timeout work an earlier draft of the plan put here arrived from #305 instead; see [Prerequisite assertion](#prerequisite-assertion) below.

### 2. `lib/describe-cache.ts` — the edit-relevant describe attributes

`updateable`, `createable`, `autoNumber`, `htmlFormatted`, `encrypted`, `restrictedPicklist`, `dependentPicklist`, `controllerName` are declared on the shared `FieldDescribe`. **Types only** — the cache stores `data` wholesale (`describe-cache.ts:12`), so these already arrive on every field object; they were simply undeclared, which is precisely why `inspect-record.ts:14` keeps a private `FieldDescribe` interface in order to see one of them.

They are optional, for the same reason as the P2-1 block above them: existing consumers and test fixtures built against the base shape must keep compiling. **That makes the polarity load-bearing, and it is documented at the declaration** — absence is not permission, so consumers read them as `=== true`, never `!== false`. The model does exactly that, and a test pins it.

The behavioural claim this rests on ("the cache passes the payload through") is now itself pinned by a test, so a future mapping refactor cannot quietly drop the attributes and leave the types lying.

### 3. `lib/record-edit.ts` — the model (new, pure, zero DOM)

No DOM, no `chrome.*`, no I/O. Every export is a total function of its arguments.

#### `EDITABLE_TYPES` / `classifyFieldEditability()`

Editable types are chosen by **a rule, not a taste list** (plan decision 4): a field is editable in v1 iff the DOM offers a native lossless control for it **and** the wire format is unambiguous. That admits `multipicklist` and `time`, which are a superset of the types the AC enumerates — they are in scope because the rule puts them there, and a principle that never changes an answer is not a principle.

`classifyFieldEditability(field, mode)` is the only place the question is answered. `mode` is `'update'` (edit) or `'create'` (clone stages a create form, plan decision 6), which is the same rule set asking `createable` instead of `updateable`. Order matters, so the most specific *why* wins:

| Condition | Reason | What the user will be told |
|---|---|---|
| `calculated: true` | `formula` | Calculated server-side (formula or roll-up summary) |
| `autoNumber: true` | `auto-number` | Salesforce assigns this value |
| platform-maintained name **and** neither `updateable` nor `createable` | `system` | Maintained by Salesforce and not writable |
| type not in the editable set, or a rich-text `textarea`, or `encrypted` | `unsupported-type` | wording names the type / the specific reason |
| the residual case | `no-permission` | "not editable for you (field-level security or object permissions)" |

Two details worth review attention:

- **`system` requires both halves**, not just membership of the name list. An org with Set Audit Fields enabled reports `CreatedDate` as `createable`, and then it is a genuine create-time input rather than a field to grey out. Describe is authoritative over the hardcoded list. Pinned by a test.
- **The `no-permission` wording is non-committal on purpose.** Describe is evaluated per running user, so a field *this user* cannot edit is indistinguishable from one nobody can edit. We do not assert a cause we cannot verify — and the test asserts the wording, not just the reason code.

Nothing is ever filtered out of the *view*; only out of the payload. Read-only rows carry their reason as a chip in PR-2.

#### `formatForInput()` / `coerceForWire()`

Per-type round trips. Two are worth calling out:

- **`date` is handled as text and never parsed through a `Date`.** Salesforce sends a bare `YYYY-MM-DD` with no zone; a round trip through the browser's zone shifts it a day for everyone west of UTC. This is the off-by-one-day bug class, and the fix is to never invoke the machinery that causes it.
- **Unparseable numeric text is handed to the org rather than guessed at.** `coerceForWire` never throws and never rejects a value. A number field given `"twelve"` passes it through so the *org* rejects it — which lands as a field-level error on that exact field, now that (1) makes that possible. Guessing locally would trade a precise server error for a silent local one.

`coerceForWire` is **idempotent**, and that is load-bearing rather than incidental — see below.

#### `buildDirtyDiff(describe, original, edited) → { patchBody, changedFieldNames }`

The **single source of truth** for "what changed". The save bar and the PATCH body become one computation, so they cannot disagree. Three filters, in order:

1. **Only describe-known fields classified editable.** A read-only field cannot become dirty, so a stale entry in the edited map for one can never reach the wire.

2. **Only fields present in the original GET payload.** This is the security-relevant one. A field hidden from the running user by field-level security is *simply absent* from the record JSON, so it reads back as `undefined` — indistinguishable, to a naive comparison, from a field the user just cleared. Including it would `PATCH` `null` over a value the user was never allowed to see. `hasOwnProperty` is the test, **never** truthiness, because a legitimately null field IS present and must stay editable. Three tests pin this: the hidden field is omitted, it is omitted *even when the edited value is explicitly `null`* (the dangerous direction), and a present-but-null field still edits normally.

3. **Only values that actually differ in canonical wire form.** Both sides are run through `coerceForWire` before comparison — which is why its idempotence matters: the diff applies it to the untouched GET value and the edited control value alike, and a non-idempotent normaliser would manufacture a phantom change on the second pass. This is what makes `3` vs `'3'`, `''` vs `null`, and `2026-07-30T14:35:00.000+0000` vs `2026-07-30T14:35:00Z` all correctly *no change*.

#### `mapSaveErrors(details, renderedFieldNames)`

Routes each named field independently:

| Case | Destination |
|---|---|
| non-empty `fields`, field **is** rendered | inline error on that field's row |
| non-empty `fields`, field is **not** rendered | form-level banner that **names the field**: `Foo__c: Value too long` |
| empty `fields` (validation rule with no field binding, `UNABLE_TO_LOCK_ROW`, trigger `addError()` on the record) | form-level banner carrying the `errorCode` |

The banner is not a fallback for ugly errors — it is the guarantee that **no error is lost**. An error naming a field the user cannot currently see (filtered out, hidden by the show-nulls toggle, or not rendered at all) would otherwise vanish with the row, which is exactly the silent failure AC-1 exists to prevent. One record naming both a visible and a hidden field produces **both** an inline error and a banner line; a test asserts the total count so nothing can be dropped.

<a id="prerequisite-assertion"></a>
## Prerequisite assertion (not reimplementation)

The plan (line 155-157) has PR-1 **assert** #305's guarantee rather than reimplement it, and flags the risk that #305 might deliver only the timeout bump and not the distinguishability — in which case the narrow distinguishing piece would move into PR-2.

**Verified on this branch's base: it delivered both halves.** `apiRequest` routes through `sendProxied` with `WRITE_MESSAGE_TIMEOUT_MS = 120_000` (`salesforce-api.ts:39`, `:375-386`), so writes are not capped at 5 s; and `sfApiErrorKind(err)` returns a stable `'timeout' | 'no-session' | 'http-error'` discriminant, with `mutating` set on the timeout case, so a transport failure is distinguishable at the call site from a server rejection. The plan's three-state save contract is implementable in PR-2 as written, with no additional plumbing. **No scope moves.**

`SalesforceRestError` composes with that rather than competing: it sets the same `sfdtKind: 'http-error'` tag, so `sfApiErrorKind` stays total over this client.

## What this deliberately does not touch

- **No feature file.** `inspect-record.ts` is untouched — including its private `FieldDescribe` interface. The plan says two things about that interface: the describe-cache declaration is "how inspect-record stops keeping" it (line 273), and "PR-1 deletes inspect-record's private `FieldDescribe`" (line 345). Those conflict with PR-1's own scope boundary, "touches no feature file" (line 276). **I followed the scope boundary**, which is the more specific and more emphatic of the two, and the deletion is a one-line change that belongs with PR-2 — which rewrites that file wholesale anyway, and would otherwise have to review the same edit twice. Flagging it rather than deciding it silently; happy to move it here if you'd rather.
- **No write-timeout or settings-layer change** — both belong to the merged prerequisites.
- **No new permission, no manifest change, no `PRIVACY.md` change.** Plan decision 7: if any PR in this chain finds itself reaching for a permission, that is a stop-and-escalate event. None did.
- **No new runtime dependency** — `package.json` / lockfile untouched.
- **No DOM code**, so the zero-`innerHTML` invariant holds trivially. No SID-adjacent code: everything stays above the `sfApiFetch` proxy boundary.
- **No docs-site MDX.** Nothing user-facing changes; the plan scopes docs-site duty to the user-facing PRs (2, 3, 4).
- **`FEATURES.json` untouched** — nothing here flips a `passes`.

## What PR-2 will consume

For reviewer orientation only — none of this is in this PR:

- per-type editors driven by `classifyFieldEditability` + `formatForInput`, with reason chips on read-only rows;
- `buildDirtyDiff` replacing **both** loops at `:399` and `:520`, retiring the raw `!==` comparison;
- the three-state outcome contract (Saved / **No changes were saved** / **Save outcome unknown**) branching on `sfApiErrorKind`, with a post-success re-GET retiring the optimistic `originalRecordData = { ...editedRecordData }` at `:540`;
- `SalesforceRestError.details` → `mapSaveErrors` → inline field errors and the form banner, with the field filter cleared and show-nulls forced on so an error can never land on a hidden row.

## Tests

New file `extension/test/record-edit.test.ts` — **42 cases**: the editable-type set and its exclusions; the full classification table including the writable-audit-field case, the rich-text `textarea`, the ordinary-looking `encrypted` field, the create-vs-update split, and the absence-is-not-permission default; per-type coercion round trips for all 17 editable types plus an explicit idempotence sweep; the diff builder including the three FLS cases, the three normalisation cases, and malformed-describe tolerance; and the error mapping including the not-rendered-field case, the no-field case, the mixed record, and a nothing-is-lost count assertion.

Extended `extension/test/salesforce-api.test.ts` (42 → **48 cases**): the structured details are carried, `.message` is asserted byte-for-byte (`Salesforce PATCH request failed (HTTP 400): Value too long`) and `sfdtKind` still reads `'http-error'`, a record with no `fields[]` normalises to `[]` rather than `undefined`, multi-record bodies are kept whole, a non-JSON 503 body degrades to an empty list, and `parseRestErrorDetails` is exercised across every shape the wire can produce.

Extended `extension/test/describe-cache.test.ts` (6 → **7 cases**): the cache passes the whole field payload through untouched, which is the claim the types-only change rests on.

## Gates

Run from this branch head (`b430d32`), on a clean isolated install:

| Gate | Result |
|---|---|
| `extension/` · `npx vitest run` | **86 files, 1360 tests passed** (was 85 / 1311) |
| `extension/` · `npx tsc --noEmit` | **clean, exit 0** |
| `extension/` · `npx eslint .` | **clean, exit 0** |
| `extension/` · `npx wxt build` | **success, chrome-mv3, 1.23 MB** |
| root · `npm test` | **286 files, 4614 tests passed** (was 285 / 4565) |
| root · `npm run lint` | **clean, exit 0** |
| root · `npm run check:all-contracts` | **all green** — catalogs up to date (11 files), licenses, node policy, auth docs, FEATURES.json, package paths |
| root · `npm run generate:catalogs` | **zero drift** — `git status --porcelain generated/` empty |

---


scoobydrew83 (commented):
## Verdict: REQUEST_CHANGES (one blocking finding)

Independent review from a clean isolated worktree at `b430d32`, fresh `npm ci` at the repo root (795 packages; `extension/` is a workspace so its deps hoist — `extension/node_modules` legitimately holds only `typescript`) plus `packages/flow-core` built so `@sfdt/flow-core/bridge-contract` resolves. **All eight gates reproduce, to the test.** Findings below come from executing the real module under vitest, not from reading the tests.

Proportion up front: the FLS guard, the error parser, the prerequisite assertion and the "and only that" claim on `salesforce-api.ts` all survived deliberate attack. The blocking item is a hole in one of the three normalisation guarantees the PR body enumerates, and it is exactly the kind of thing a contract-first PR exists to catch before a UI binds to it.

---

## Gates — reproduced

| Gate | Author claimed | I measured |
|---|---|---|
| `extension/` · `npx vitest run` | 86 files / 1360 tests | **86 files / 1360 tests passed** ✅ |
| `extension/` · `npx tsc --noEmit` | clean | **exit 0** ✅ |
| `extension/` · `npx eslint .` | clean | **exit 0** ✅ |
| `extension/` · `npx wxt build` | success, 1.23 MB | **success, chrome-mv3, Σ 1.23 MB** ✅ |
| root · `npm test` | 286 files / 4614 tests | **286 files / 4614 tests passed** ✅ |
| root · `npm run lint` | clean | **exit 0** ✅ |
| root · `npm run check:all-contracts` | all green | **all green** — catalogs up to date (11 files), licenses (7 manifests), node policy, auth docs (54 files), FEATURES.json, package paths (117 files) ✅ |
| root · `npm run generate:catalogs` | zero drift | **`git status --porcelain generated/` empty** ✅ |

Per-file counts also check out: `salesforce-api.test.ts` + `describe-cache.test.ts` = 55 (48 + 7), matching the claimed 42→48 and 6→7.

Side-claims, all confirmed:
- **No new runtime dependency** — `package.json` / `package-lock.json` are not in the diff at all.
- **No manifest permission change** — `wxt.config.ts` untouched; `PRIVACY.md` untouched.
- **`FEATURES.json` untouched.** `check-features-edits.mjs` passes.
- **Diff is 7 files, all under `extension/`**, +1117/−3. No feature file. Scope boundary held.
- **The model really is unused**: `classifyFieldEditability` / `EDITABLE_TYPES` appear nowhere in `.output/chrome-mv3/` — tree-shaken out, which is the strongest available evidence for "no user-visible change".

(Methodology note, since a related trap bit the builder: my first root run reported 287/4643. That was my own 29-case probe file sitting in `extension/test/` — root vitest loads the extension project as one of its `test.projects`. Removed it, re-ran, got 286/4614 exactly. Flagging so nobody chases a phantom.)

---

## Blocking

### B1 — `datetime` (and `time`) read→write round trip destroys seconds, producing a false-dirty PATCH

`formatForInput` truncates a datetime to minute precision (`record-edit.ts:203-210`), and `coerceForWire` re-expands it with `:00.000`. The round trip is therefore **not value-preserving**, which breaks the property `buildDirtyDiff` is built on.

Executed against the real module (`TZ=America/Los_Angeles`):

```
field:  { name: 'SlaExitDate__c', type: 'datetime' }
GET:    "2026-07-30T14:35:45.000Z"

formatForInput(f, wire)              -> "2026-07-30T07:35"          // :45 already gone
coerceForWire(f, "2026-07-30T07:35") -> "2026-07-30T14:35:00.000Z"
coerceForWire(f, wire)               -> "2026-07-30T14:35:45.000Z"

buildDirtyDiff(describe,
  { SlaExitDate__c: "2026-07-30T14:35:45.000Z" },   // original, from the GET
  { SlaExitDate__c: "2026-07-30T07:35" })           // what the control holds
=> patchBody { "SlaExitDate__c": "2026-07-30T14:35:00.000Z" }
```

**Wrong outcome, twice over.** The save bar lights up for a field nobody edited, and the PATCH permanently overwrites `:45` with `:00`. Same for sub-second precision — `"2026-07-30T14:35:00.123Z"` diffs to `{"When__c":"2026-07-30T14:35:00.000Z"}`. `time` has the identical shape: `formatForInput(f, '09:30:00.5')` → `'09:30:00'`, and an unstepped `<input type=time>` truncates further to `'09:30'`, which coerces to `'09:30:00.000Z'`.

Two reasons I'm calling this blocking rather than should-fix:

1. **The source comment asserts the opposite guarantee.** `record-edit.ts:200-202` says seconds are dropped because "a save must not silently rewrite them — `coerceForWire` round-trips whatever the control gives back." That reasoning doesn't hold: `coerceForWire` can only round-trip what `formatForInput` put into the control, and the seconds were destroyed one step earlier. A comment documenting a safety property the code doesn't have is worse than no comment.

2. **The test that should have caught it is fixture-lucky.** `record-edit.test.ts:163-171` asserts the round trip via `getTime()` equality — the right assertion — but with `'2026-07-30T14:35:00.000Z'`. Change the fixture to `...14:35:45.000Z` and that existing test fails as written. Nothing about the assertion needs to change.

**Reachability, stated honestly.** Nothing ships broken today — no feature file consumes this. Under today's `inspect-record.ts` wiring (`editedRecordData` seeded from `originalRecordData` at `:513`, control values written only on `change`) an *untouched* field wouldn't hit it. But PR-2 replaces that wiring wholesale with per-type editors, and the natural shape — `input.value = formatForInput(...)`, `edited[name] = input.value` — walks straight into it. Even under the most conservative wiring, a user who nudges only the *date* half of a datetime silently loses the seconds. This is the moment to fix it: the whole argument for a contract-first PR is that the contract is right before anything binds to it, and fixing it after PR-2 means re-reviewing the UI to prove the model change didn't shift behaviour.

Cheapest fix: have `formatForInput` emit `YYYY-MM-DDTHH:mm:ss` (and `HH:mm:ss` for `time`), with the control carrying `step="1"` in PR-2. The round trip becomes lossless, and both existing tests keep their assertions.

---

## Should-fix

### S1 — `multipicklist` reorders an untouched value

`formatForInput` preserves the record's stored order; a `<select multiple>` reads back in **DOM order** (picklist-definition order), and `coerceForWire` joins in whatever order it receives. Verified end-to-end in happy-dom:

```
stored:                            "C;A"
formatForInput                  -> ['C','A']
select.selectedOptions readback -> ['A','C']       // DOM order wins
buildDirtyDiff({Tags__c:"C;A"}, {Tags__c:['A','C']})
=> patchBody { "Tags__c": "A;C" }
```

Untouched field, unrequested write. Not destructive — same set — but it fires on *every* save for any record whose multipicklist order differs from the picklist definition, so the "phantom dirty state" the PR body sets out to eliminate survives for this one type. Same root cause as B1 (round trip not value-preserving); the fix is a set comparison for `multipicklist` in `buildDirtyDiff`, or canonical ordering in `coerceForWire`. No test currently covers ordering.

### S2 — Set-Audit-Fields audit field gets a wrong, falsifiable reason in update mode

The `system` classification correctly requires both halves (`record-edit.ts:149`) — verified, and the Set-Audit-Fields case is genuinely handled. But only in `create` mode, which is the only mode the test at `record-edit.test.ts:85` exercises. In `update` mode:

```
classifyFieldEditability({name:'CreatedDate', type:'datetime', updateable:false, createable:true}, 'update')
=> { editable:false, reason:'no-permission',
     message:'Not editable for you (field-level security or object permissions).' }
```

`CreatedDate` is never updateable in any org under any FLS configuration. The PR body defends the non-committal wording with "we do not assert a cause we cannot verify" — I agree with that principle — but here the cause *is* verifiable (the name is in `SYSTEM_FIELD_NAMES`) and the code asserts a wrong one. A user in a Set-Audit-Fields org sees an FLS explanation for a platform restriction and may go hunting through profiles for a permission that doesn't exist. Suggest a distinct reason for "platform-maintained, not writable in this mode" that reuses the `system` wording without claiming FLS.

### S3 — "every export is a total function of its arguments" is not quite true

A `null` entry inside `describe.fields` throws:

```
buildDirtyDiff({fields:[null, fld('Name')]}, {Name:'a'}, {Name:'b'})
=> TypeError: Cannot read properties of null (reading 'calculated')
classifyFieldEditability(null) => same
```

`mapSaveErrors` likewise throws on a `null` element, a non-string inside `fields`, a non-string in `renderedFieldNames`, or a non-array `details`. All are excluded by the TypeScript signatures and none is producible by `parseRestErrorDetails`, so this is not a live risk — but the header comment at `record-edit.ts:3` makes an unqualified totality claim, and the malformed-describe test (`:344`) only covers `null` / `undefined` / `{fields:[]}`. Either guard the element or soften the claim.

---

## What I attacked and could not break

**`buildDirtyDiff`'s FLS filter is sound.** Every payload shape I could construct stays out of `patchBody`:

| Attack | Result |
|---|---|
| key absent from `original`, present in `edited` | `{}` ✅ |
| absent from `original`, `edited` value explicitly `null` (the dangerous direction) | `{}` ✅ |
| key only on `original`'s **prototype chain** | `{}` ✅ (`hasOwnProperty`, not `in`) |
| key only on `edited`'s prototype chain | `{}` ✅ |
| `original` is a null-prototype object | no throw, correct diff ✅ |
| `original` has the key with value `null` | stays editable → `{Secret__c:'v'}` ✅ |
| `original` has the key with value `''` | stays editable ✅ |
| empty `original`, `edited` nulls three editable fields | `{}` ✅ |

The `hasOwnProperty`-not-truthiness claim is exactly right, and the "present but null stays editable" half — the part a naive `!= null` guard would break — genuinely works. This is the security-relevant behaviour in the PR and it holds.

**`parseRestErrorDetails` is total.** Twenty hostile bodies, none throws: `''`, `'null'`, `'0'`, `'false'`, a bare JSON string, `[]`, `{}`, nested `[[{...}]]`, `<html>503</html>`, `[{"message":null}]`, `fields:"NotAnArray"`, `fields:[1,null,"Ok__c"]` (→ `["Ok__c"]`), `errorCode:42` (→ `''`), a bare object instead of an array, a mixed array where the middle record is junk (→ the two good ones kept), `[1,2,3]`, `[null]`, `'undefined'`. A `{"__proto__":{...}}` body does **not** pollute `Object.prototype`. 5000 records parse fine. Claim verified.

**`SalesforceRestError` really is only that.** `salesforce-api.ts` is +80/−3; the three deleted lines are the `tagError(new Error(...))` call, nothing else. No write-timeout code is in the diff. `.message` is byte-identical, `sfdtKind === 'http-error'`, `.status` unchanged, `instanceof Error` true, no call site moved. I checked the shipped consumers directly — `data-import`, `field-creator`, `apex-anonymous`, and every other feature use the `err instanceof Error ? err.message : String(err)` idiom, so they take the `.message` branch and are unaffected.

**The prerequisite claim (#305) is correct — nothing moves into PR-2.** Verified against merged `origin/develop`, not against this branch: `WRITE_MESSAGE_TIMEOUT_MS = 120_000` at `develop:39`, applied in `sendProxied` at `develop:302`, with `apiRequest` declaring `mutating: true` at `develop:385`; `sfApiErrorKind()` at `develop:91` returning `'timeout' | 'no-session' | 'http-error' | null`. Both halves are there. The three-state save contract is implementable in PR-2 as written.

**`date` is genuinely timezone-proof.** No `Date` constructor on any `date` path — string slicing only. Identical output under `TZ=UTC` and `TZ=America/Los_Angeles`, including when the source value is a full datetime (`'2026-01-01T23:00:00.000Z'` → `'2026-01-01'` in both zones, where a `Date` round trip would have given `2025-12-31` in LA). The off-by-one-day bug class is structurally absent, not merely untested.

**Polarity claim verified.** `!== false` appears nowhere in the code — the only occurrence in the diff is the doc comment forbidding it (`describe-cache.ts:38`). `updateable: 'true'` (truthy non-boolean) correctly classifies as `no-permission`, i.e. the default is closed.

**Idempotence of `coerceForWire` holds** for every editable type across the edge shapes I threw at it — `'1e3'`, `'0x10'`, `' 12 '`, `'12,50'`, `'Infinity'`, `'-0'`, `'00123'`, `'1_000'`, `'9:30'`, `'09:30:00.5'`, `'23:59:59.999Z'`, `' x '`, `[]`, `null`. Note B1/S1 are *not* idempotence failures — `coerceForWire` is idempotent as claimed; the loss happens in `formatForInput`, upstream of it. The PR body's framing ("idempotence is what lets `buildDirtyDiff` canonicalise both sides") is correct as far as it goes, but idempotence alone doesn't buy round-trip fidelity, and that gap is where B1 lives.

---

## Nits

- **`.name` did change**, from `'Error'` to `'SalesforceRestError'`, so `String(err)` now prefixes `SalesforceRestError: `. The PR body enumerates `.message`/`sfdtKind`/`.status` as unchanged and doesn't mention it. Harmless in fact — every consumer takes the `err.message` branch — but worth a word, since "strictly additive" is doing load-bearing work in the argument.
- **The polarity rule in `describe-cache.ts:37-38` is stated for all nine attributes, but only four are permission flags.** For the deny-flags (`autoNumber`, `htmlFormatted`, `encrypted`, and the pre-existing `calculated`), `=== true` is the *permissive* default, not the closed one: `encrypted: 'true'` (string) classifies as **editable**. Unreachable from a real describe, which always sends booleans, but the comment implies one uniform fail-safe direction where there are two.
- **A field literally named `__proto__`** yields `changedFieldNames: ['__proto__']` with `patchBody: {}` — the save bar and the PATCH body disagree, which is the one thing `buildDirtyDiff` exists to prevent. No prototype pollution (verified). Unreachable, since Salesforce API names are `[A-Za-z0-9_]+`, but `Object.create(null)` for `patchBody` closes it for free.
- **An own key whose value is explicitly `undefined`** defeats filter 2: `buildDirtyDiff(D, {Name:'x', Secret__c: undefined}, {Name:'x', Secret__c:'pwn'})` → `{"Secret__c":"pwn"}`. `JSON.parse` cannot produce that, so it is not a live FLS bypass — but if PR-2 ever builds `original` by key-copying rather than using the parsed record directly, the guard silently stops guarding.
- **`SYSTEM_FIELD_SET` is case-sensitive** while `mapSaveErrors` case-folds field names. Two conventions in one module; describe always sends canonical casing, so nothing breaks today.
- **`buildDirtyDiff` hardcodes `'update'`** (`:364`) with no mode parameter, so PR-3's clone POST body can't reuse it. The plan only promises clone reuses the *classification*, so this is consistent with the design — noting it so PR-3 isn't surprised.
- The plan's attribute list included `nillable`; it was already declared (required, `describe-cache.ts:19`). Not a gap.

---

## On the scope question: the two `FieldDescribe` definitions

**I agree with the author's call, and I think it was right to surface it rather than decide silently.**

Line 276 ("touches no feature file") is a scope *boundary* — a hard edge that makes the PR reviewable as "does this change behaviour? provably not." Line 345 ("PR-1 deletes inspect-record's private `FieldDescribe`") is a *consequence* the plan sketched for a chain that had not yet been split the way it ended up being split. When a boundary and a consequence collide, the boundary should win, because it is the thing that makes the "no user-visible change" claim mechanically checkable — and it is: the model tree-shakes clean out of the build. Deleting the private interface would have pulled `inspect-record.ts` into the diff, and a reviewer's first question becomes "prove this didn't change the inspector", for a line PR-2 deletes anyway.

**Is leaving both in the tree a real hazard, or just untidy?** I checked the specific failure mode you named — PR-2 silently binding to the stale private one. It is close to untidy, with one asymmetry worth knowing:

- The private interface (`inspect-record.ts:14-21`) declares `updateable: boolean` — **required**. The shared one declares `updateable?: boolean` — **optional**. They are not interchangeable, and the private one is the *stricter* shape.
- The direction that matters is safe. A mixed state does compile (the private interface is structurally assignable to the shared one, so `classifyFieldEditability` would accept it), so `tsc` will not catch the omission — but the file's own reads are `!== true`-equivalent (`field.updateable` at `:319`, `chk.disabled = !field.updateable` at `:312`), so swapping to the optional shape cannot flip a default open. The "PR-2 quietly gets different permission semantics" failure does not materialise.
- The residual risk is the boring one: PR-2 rewrites the file and simply forgets the deletion, leaving a dead private interface indefinitely. Nothing mechanical catches that.

So: untidy, not dangerous. I'd ask only that PR-2 lead with the deletion as its first commit so it can't be lost in the rewrite — a line in the PR-2 description is sufficient. I would not move it into this PR; that trades a real reviewability property for a cosmetic one.

---

## Summary

Genuinely good work, and unusually honest about its own seams — the prerequisite is verified rather than asserted, the FLS filter does what its comment says, the plan contradiction was flagged instead of papered over, and every gate number in the body reproduces exactly. The one thing I'd hold it for is B1: the model's read→write round trip loses data on `datetime`/`time`, the comment above it claims the opposite, and the existing test would catch it with a one-character fixture change. Fix that (and ideally S1, same root cause) and this is a clean merge.

---

scoobydrew83 (commented):
## Verdict: APPROVE-equivalent — round 1's blocking finding is closed, no new blockers

Re-reviewed at `49670c0` in the same isolated worktree. **All four round-1 findings are genuinely fixed**, not papered over — I re-derived each by execution rather than reading the new tests. Two small things came *in* with the fixes; neither blocks, and I'd be happy to see them ride into PR-2 rather than spin a third round.

---

## Gates at `49670c0` — totals and deltas both reconcile

| Gate | Claimed | Measured | Δ from `b430d32` |
|---|---|---|---|
| `extension/` vitest | 86 / 1374 | **86 files / 1374 tests** ✅ | +0 files / **+14 tests** |
| root `npm test` | 286 / 4628 | **286 files / 4628 tests** ✅ | +0 files / **+14 tests** |
| `record-edit.test.ts` | — | **42 → 56** | **+14** |
| ext `tsc --noEmit` | clean | **exit 0** ✅ | |
| ext `eslint .` | clean | **exit 0** ✅ | |
| ext `wxt build` | 1.23 MB | **success, Σ 1.23 MB** ✅ | |
| root `lint` | clean | **exit 0** ✅ | |
| root `check:all-contracts` | green | **all green** (catalogs 11, licenses 7, node, auth-docs 54, FEATURES.json, paths 117) ✅ | |
| root `generate:catalogs` | zero drift | **`git status --porcelain` empty** ✅ | |

**The deltas close exactly**: +14 root tests = +14 extension tests = +14 `record-edit.test.ts` tests, with zero new test files. Nothing was removed elsewhere to make room, and no assertion was weakened — every assertion that *changed* got stricter (the round-trip regex went from `T\d{2}:\d{2}$` to `T\d{2}:\d{2}:\d{2}$`; every `datetime`/`time` fixture gained non-zero seconds; the idempotence sweep moved to `.345Z`). That was the failure mode worth watching for and it isn't present.

Still 7 files, all under `extension/`; `salesforce-api.ts` untouched in round 2, so the additive-only verification from round 1 carries. No feature file, `FEATURES.json` untouched, and `classifyFieldEditability` / `EDITABLE_TYPES` / `multipicklistKey` remain absent from `.output/chrome-mv3/` — "no user-visible change" is still mechanically true.

**Pushed history is coherent**: 8 commits off `develop`, four in round 2, each commit's file scope matching its subject (`7ce2151` datetime/time only; `e841ba3` multipicklist only; `93de461` reasons + guards + polarity docs; `49670c0` CHANGELOG only). No trace of the misleading commit — the reset worked.

---

## Findings I now consider closed

### B1 — datetime/time truncation: **CLOSED**

Round trip is exact across the whole spread I could construct, including year boundaries, 1-ms values and historical dates:

```
2026-07-30T14:35:45.000Z -> 2026-07-30T07:35:45     -> 2026-07-30T14:35:45.000Z  OK
2026-07-30T14:35:00.123Z -> 2026-07-30T07:35:00.123 -> 2026-07-30T14:35:00.123Z  OK
2026-01-01T00:00:00.001Z -> 2025-12-31T16:00:00.001 -> 2026-01-01T00:00:00.001Z  OK
2026-12-31T23:59:59.999Z -> 2026-12-31T15:59:59.999 -> 2026-12-31T23:59:59.999Z  OK
1999-02-28T05:06:07.008Z -> 1999-02-27T21:06:07.008 -> 1999-02-28T05:06:07.008Z  OK
```
7/7 exact for `datetime`, 5/5 for `time`. **TZ sweep re-run independently**: `UTC`, `America/Los_Angeles`, `Asia/Kolkata` (+05:30), `Pacific/Chatham` (+12:45) — 56/56 in each. The rewritten real-edit fixture that derives its edited value from `formatForInput` rather than string-substituting UTC digits is the right shape; that was a good catch on your side.

**Mutation check reproduces, and is stronger than claimed.** Reverting `toLocalDateTimeInput` to minute precision fails **5** tests, not 4 — `datetime: milliseconds are emitted only when non-zero` also goes red. Dropping the multipicklist set comparison fails 1. The fixtures are demonstrably capable of failing, which is the whole point.

**One risk the `step` discussion doesn't cover, which I checked and which passes.** PR-2 will read `input.value`, i.e. the **UA-sanitised** string, not what the model wrote. The HTML value-sanitisation algorithm drops zero seconds and strips trailing fraction zeros, so the readback genuinely differs from the emitted value:

```
wire 2026-07-30T14:35:00.000Z  set "20:05:00"     readBack "20:05"        -> CLEAN
wire 2026-07-30T14:35:45.250Z  set "20:05:45.250" readBack "20:05:45.25"  -> CLEAN
wire 2026-07-30T14:35:00.100Z  set "20:05:00.100" readBack "20:05:00.1"   -> CLEAN
wire 09:30:45.500Z             set "09:30:45.500" readBack "09:30:45.5"   -> CLEAN
```
All ten cases diff clean (probed through real `<input>` elements under `Asia/Kolkata`), because `coerceForWire` re-normalises the fraction to the right magnitude. **The fix is robust to the sanitiser, not just to the raw string** — worth knowing, since it's the part that would have bitten silently.

### S1 — multipicklist ordering: **CLOSED**

16 probe cases through `buildDirtyDiff`, all correct:

| Input | Result |
|---|---|
| reorder only, array `'C;A'` → `['A','C']` | CLEAN ✅ |
| reorder only, string `'C;A'` → `'A;C'` | CLEAN ✅ |
| empty entry either side, trailing `;` | CLEAN ✅ |
| `null` ↔ `[]` | CLEAN ✅ |
| case difference `'A;B'` → `['a','B']` | change ✅ (picklist values are case-sensitive) |
| whitespace `'A;B'` → `['A',' B']` | change ✅ |
| genuine add / remove / clear | change, **written in control order** (`'A;B;C'`, `'A'`, `null`) ✅ |

The "sorting in `coerceForWire` would rewrite the user's selection on every real save" reasoning is right, and the code matches it — `multipicklistKey` decides equality only and never touches the written value.

### S2 — audit-field reason: **CLOSED**, both directions

```
Set-Audit org, CreatedDate (updateable:false, createable:true)
  update -> { system, "Maintained by Salesforce and not editable." }        ✅ was the wrong FLS message
  create -> { editable: true, type: 'datetime' }                            ✅ describe still beats the name list
Ordinary org, CreatedDate (both false)
  update -> system "not editable"   |  create -> system "not settable on create"   ✅
Hypothetically-updateable system field (IsDeleted, updateable:true)
  update -> editable                                                        ✅ describe wins, per mode
```
The original intent survives the change. Non-system fields are unaffected (`Locked__c` still `no-permission`; `ExternalKey__c` still create-editable).

### S3 — totality: **CLOSED for the two array walkers**

Every hostile shape that threw in round 1 now returns: `null`/`undefined`/`42`/`'x'` describe elements, a `{name: 5}` element, `null` details, non-string `fields` entries, non-string/`null`/`undefined` `renderedFieldNames`, a non-array `details`. Nothing-is-lost still holds (3 records → 3 destinations). The narrowed header claim ("total over its declared argument types… the two that walk arrays additionally skip malformed elements") is now accurate rather than aspirational.

### Nits — 6 of 7 closed

`.name` change disclosed in the CHANGELOG (and disclosed *as* a difference, not buried); PERMIT/DENY polarity split is exactly the asymmetry I meant; `__proto__` closed via `Object.create(null)` with the serialised-body assertion, which is the right way to test it; explicit-`undefined` own key now rejected; system-list case-folding unified with `mapSaveErrors`.

### Round-1 conclusions re-verified, not assumed

I re-ran the FLS sweep since `buildDirtyDiff` was rewritten: absent key `{}`, `edited: null` `{}`, prototype-chain `{}`, explicit-`undefined` `{}`, present-null and present-empty both still editable. **No regression.**

---

## New in round 2

### N1 (should-fix) — `classifyFieldEditability` now throws on a field with no `name`

The case-folding fix introduced `field.name.toLowerCase()` with no guard, so:

```
classifyFieldEditability({ type: 'string' })       -> TypeError: Cannot read properties of undefined (reading 'toLowerCase')
classifyFieldEditability({})                       -> same
classifyFieldEditability({ name: 42, ... })        -> TypeError: field.name.toLowerCase is not a function
```

Round 1 returned `{ reason: 'no-permission' }` for these. It's *in contract* — `FieldDescribe.name` is required and the new header explicitly declares leaf functions non-defensive — and `buildDirtyDiff` guards its own path with `typeof field.name !== 'string'`. So this is not reachable from a real describe.

What makes it worth a line: `classifyFieldEditability` is the function **PR-2 calls per field in a render loop**, and the justification given for guarding the array walkers — "both sit on paths where a throw would cost the user their edits" — applies just as well to a half-built inspector. The guard is one character wider than the one already written: `String(field.name ?? '').toLowerCase()`. Your call whether that belongs here or in PR-2's loop.

### N2 (should-fix / or document as a known limit) — DST fall-back still shifts an untouched datetime by an hour

Pre-existing, not a round-2 regression — I missed it in round 1 — but it is now the *only* remaining round-trip infidelity, so it's worth naming:

```
TZ=America/Los_Angeles
2026-11-01T08:30:45.000Z (PDT) -> 2026-11-01T01:30:45 -> 2026-11-01T08:30:45.000Z  OK
2026-11-01T09:30:45.000Z (PST) -> 2026-11-01T01:30:45 -> 2026-11-01T08:30:45.000Z  *** SHIFTED 1h ***
```

01:30 local occurs twice on that date; `new Date('…T01:30:45')` resolves to the earlier one. So a datetime landing in the repeated hour reads back as a different instant — untouched field, false-dirty, and a PATCH that moves the value an hour. Once a year per DST zone, but it's the same failure *class* as B1.

This is inherent to a local-time control and cannot be fixed inside `coerceForWire`. The cheap fix lives in `buildDirtyDiff` and — importantly — does **not** reintroduce the control-precision coupling you rejected:

```ts
// The control still shows exactly what we rendered into it: the user did not touch it.
if (formatForInput(field, original[field.name]) === edited[field.name]) continue;
```

That's a pure identity check on the rendered representation, not a precision tolerance — it needs no knowledge of `step` and stays correct whatever PR-2 chooses. It would also make N3 below moot. Equally defensible to just document it; I'd not hold the PR either way.

### N3 (nit) — `patchBody` is null-prototype, undocumented

`Object.create(null)` closes the `__proto__` case, and the real PATCH path is safe (`salesforce-api.ts:455` does `JSON.stringify(body)`; `Object.keys`, spread, `in` and `JSON.stringify` all work). But two things now throw on a public return value:

```
patchBody.hasOwnProperty('Name')  -> TypeError: not a function
`${patchBody}` / String(patchBody) -> TypeError: Cannot convert object to primitive value
```

In a module whose new header promises the save path won't throw, that deserves one line on `DirtyDiff.patchBody` rather than being discovered by a future logging statement.

### N4 (nit) — duplicate multipicklist entries count as a change

`'A;B'` → `['A','A','B']` writes `'A;A;B'`; `'A;A;B'` → `['A','B']` writes `'A;B'`. Unreachable (Salesforce won't return duplicates, `<select multiple>` can't produce them) and harmless (the org dedupes). Raising it only because the framing is "compare as sets" and a set would dedupe — `[...new Set(parts)]` in `multipicklistKey` if you want the name to be literally true.

---

## Rulings you asked for

### Is a doc comment enough to carry the `step="1"` obligation across a PR boundary?

**No — but the guard belongs in PR-2, not here.** I confirmed the failure returns verbatim: a step-less control handing back `HH:mm` produces `{"When__c":"2026-07-30T14:35:00.000Z"}` from an untouched field. The round-1 bug, through the front door, with the model still perfectly correct.

The cheapest *effective* guard is not a `step` assertion. It's the property test one level up: **render a record into real controls, read `input.value` back, run `buildDirtyDiff`, assert the diff is empty.** That catches a missing `step`, a UA-sanitisation surprise, a wrong formatter, and anything else nobody thought of — because it tests the property rather than one mechanism that happens to produce it. You already have exactly this test at model level (`a read->render->save cycle with no user edit is never dirty`); PR-2 needs its DOM-level twin. I'd make that an explicit acceptance criterion on PR-2 and consider the obligation discharged.

Optional and take-it-or-leave-it: export a `datetimeInputStep(value)` helper so the control *derives* its `step` from the value instead of remembering to set it. That is consistent with your own principle rather than in tension with it — "precision belongs to the value, display belongs to the control" is a one-way derivation, and a step hint read off the value is the value telling the control what it needs. It is not a second source of truth about comparison, which is the thing you were right to refuse. But the property test is the load-bearing guard; this is only belt-and-braces.

### The precision-preserving vs. diff-tolerance trade

**Agree, and it's not a coin flip.** The chosen fix makes the model's output independent of every UI decision — which is exactly why my UA-sanitisation probe passes without the model knowing the sanitiser exists. The rejected alternative would have had to encode the control's precision, and would have been wrong the moment PR-2 picked a different `step` for millisecond-bearing values than for second-bearing ones. Your reasoning ("a second source of truth about one question, the same failure shape as the two dirty-diff loops") is the right reason, and the result is demonstrably more robust, not just tidier.

### The `mode` param rejection

**Agree — correct to push back.** The plan has clone reuse the *classification*, not the diff. A `mode` parameter added now would be untested speculative surface on the one function that must not acquire ambiguity, and PR-3 can add it together with the tests that justify it, when it actually has a create-body to build. Rejecting a nit with a reason is the right instinct, and this reason holds.

---

## Bottom line

The blocking finding is closed with the stronger of the two available fixes, the fixtures that let it through are now provably capable of failing, and the three should-fixes landed with their original intent intact. N1 and N2 are worth a line of code each; N3 and N4 are a comment and a shrug. **None of them is worth a third round** — I'd take this as-is and fold them into PR-2, which touches all of this code anyway.

Thanks for taking the `inspect-record.ts` ask on PR-2's first commit; that closes the one loose end from round 1's scope discussion.

---

scoobydrew83 (commented):
## Confirmation pass on `49670c0..f8383fd` — clean to merge

Scoped to the delta. Everything established at `49670c0` stands and wasn't re-derived. **No findings.**

### Gates — deltas reconcile exactly

| | `49670c0` | `f8383fd` | Δ |
|---|---|---|---|
| ext vitest | 86 / 1374 | **86 / 1378** ✅ | +0 files / **+4** |
| root `npm test` | 286 / 4628 | **286 / 4632** ✅ | +0 files / **+4** |
| `record-edit.test.ts` | 56 | **60** ✅ | **+4** |

Three +4s, same four tests, one per finding, nothing deleted to make room. tsc exit 0 · eslint exit 0 · wxt 1.23 MB · root lint exit 0 · contracts all green · `generate:catalogs` zero drift. Delta touches only `lib/record-edit.ts` and `test/record-edit.test.ts`, so 7 files all under `extension/`, no feature file, `FEATURES.json` untouched — trivially. Model still absent from `.output/chrome-mv3/`.

**TZ sweep, full suite, re-run independently: 1378/1378 under each of `UTC`, `America/Los_Angeles`, `Asia/Kolkata` (+05:30), `Pacific/Chatham` (+12:45).**

### N2 — the three things you asked me to check specifically

**Placement: correct, and structurally safe beyond just being correct.** The identity check sits after all three presence guards (`hasOwnProperty(original)` → `original[name] === undefined` → `hasOwnProperty(edited)`), and it only ever `continue`s. So it cannot add a field to `patchBody` — the worst it could do from a bad position is skip one, which is the same non-write the FLS filter produces. Re-ran the full FLS sweep against the new loop anyway:

```
absent + edited value  {}     absent + edited null  {}     absent + edited ""  {}
own-undefined + edited {}     proto-only + edited   {}     non-editable        {}
present-null  -> {"Secret__c":"v"}   present-empty -> {"Secret__c":"v"}   clear -> {"Name":null}
```

No bypass, and no false-clean introduced anywhere else — I re-checked real edits across `string`/`datetime`/`boolean`/number/`date` (all still register) and confirmed the check is a no-op for `multipicklist` (`formatForInput` returns a fresh array, never `===` the input), so set logic still governs there. Mutation check backs this: disabling the line fails **exactly one** test, so it changes exactly one behaviour and masks nothing.

**Coupling: the distinction holds in the code as written.** The check calls `formatForInput` — a function in this same file — and tests for byte identity. It encodes no `step`, no precision, no UI-owned fact, and would stay correct if PR-2 chose any `step` at all. The rejected alternative needed "ignore differences below N", where N is a control property. That's a real difference, not a rhetorical one, and the comment draws it accurately.

**Residual: accurate, and understated in one direction.** Probed through real `<input type="datetime-local">` elements under `TZ=America/Los_Angeles`:

```
2026-11-01T09:30:45.000Z  rendered 01:30:45  UA readback 01:30:45  -> CLEAN
2026-11-01T09:30:00.000Z  rendered 01:30:00  UA readback 01:30     -> DIRTY {"When__c":"2026-11-01T08:30:00.000Z"}
2026-11-01T08:30:45.000Z  rendered 01:30:45  UA readback 01:30:45  -> CLEAN
```

Exactly what the comment claims — no more. Worth saying explicitly, since the comment doesn't: **the fix survives real UA sanitisation whenever seconds are non-zero.** The residual needs a DST zone *and* the repeated hour *and* zero seconds *and* zero ms simultaneously. Documented in place, in an unused model, with PR-2's DOM-level property test named as the closer. **Acceptable to ship.**

**The DST fixture is load-bearing**, which was the thing most worth confirming. With the identity check removed:

```
TZ=America/Los_Angeles  ×  datetime: a value inside a repeated DST hour is not dirty when untouched
  AssertionError: 2026-11-01T09:30:45.000Z: expected [ 'When__c' ] to deeply equal []
TZ=UTC                  60 passed
```

Fails on the ambiguous instant under LA, passes trivially under UTC — precisely as the fixture's own comment states. It can't regress silently, and the comment about why the TZ sweep matters for it is honest.

### Rulings

**N4 (duplicates collapse) — agree.** Both behaviours are unreachable, so this is a tie-break, and the tie-break goes to the module's stated purpose: the org treats `'A;B'` and `'A;A;B'` as the same value, so writing one over the other is a phantom write. Consistent with every other call in this file, and the name is now literally true. Recording the counter-argument rather than pretending it away is the right form. Confirmed the written value still comes from `coerceForWire`, not the dedup key, so a genuine change is unaffected.

**N1 and N3 — both fixed as intended.** `classifyFieldEditability` no longer throws on `{type:'string'}`, `{}`, or `{name: 42}`, and the header now names all three tolerant functions with the asymmetric-cost reasoning that justifies each. Rejecting your own "it's in contract" defence was the right call — the contract was defensible, the blast radius wasn't. `patchBody`'s null prototype is documented with the working substitutes and pinned by a test, which is what turns a sharp edge into a known one.

**Declining `datetimeInputStep` — agree**, same reasoning as the `mode` param, and applied consistently. An export with no caller is surface you have to defend without evidence. PR-2 can add it if it turns out to want it.

Noted on the mutation count correction (5, not 4) — appreciated, though it was in your favour.

**`f8383fd` is clean to merge.**

---

PR #308, opened 2026-07-30
fix(ext): stop a fallback host's 401 replacing the org's real error, and explain errors like Salesforce does

scoobydrew83:
## The mechanism

Reproduced first, theorised second. The reported string is byte-for-byte reproducible, and the cause is **not** a second 401 request, a refresh-and-retry, or anything at the proxy boundary. It is a dropped error in `sfApiFetch`'s session-cache fast path (`extension/lib/sf-api-proxy.ts`).

1. After **any** prior successful call, the per-host session cache holds which of your org's two host names served it (`acme.my.salesforce.com`). The next request goes straight there. This is the steady state, not an edge case.
2. The org answers the reported query with **HTTP 400 `MALFORMED_QUERY`** — "value of filter criterion for field `'Invoice_Status_Type__c'` must be of type double and should not be enclosed in quotes".
3. `fastPathFailure` captured that error, but it was only ever *returned* in the four "no alternate candidate remains" early-return branches. `acme.lightning.force.com` **is** an alternate candidate, so control fell through to full resolution and **re-ran the same request against the other host**.
4. `deriveBaseUrls` documents in its own comment that `lightning.force.com` "often 401s". It does — `[{"message":"Session expired or invalid","errorCode":"INVALID_SESSION_ID"}]`.
5. The retry test `has401 && !hasNon401` is computed over **that** result only — the fast-path 400 lives in a separate object — so the session-refresh retry fired and 401'd again.
6. `return result` returned **only** the 401. The 400 was discarded before it ever left the worker.
7. Page-side `buildRequestError` found no non-401 in the list, fell to `errors[0]`, and rendered the reported line.

Control that isolates it — same stubbed org, same query, only the cache differs:

| session cache | user saw |
|---|---|
| warm (normal) | `Salesforce GET request failed (HTTP 401): Session expired or invalid` |
| cold | `Salesforce GET request failed (HTTP 400): value of filter criterion for field 'Invoice_Status_Type__c' must be of type double…` |

Contained, not architectural — the proxy boundary carries status and body fine. And **not SOQL-specific**: it needed only a warm cache plus any non-401 4xx, which is why the same thing was reported on `data-import`'s CSV path and could hit any feature.

## The fix

A **non-401 4xx is the org answering the request**, not the transport failing around it. Both candidates are two names for the same org, so a second host cannot answer differently — only fail differently. Those now return verbatim. 401/5xx/network errors still fall back, but the fast-path failure is **merged** into the reported errors rather than replaced. Side effect: a write the org already rejected is no longer re-sent to the second host.

Then, since all 51 org-facing call sites render `err.message` and nothing else, the guidance had to live in that string to reach anyone. `SalesforceRestError` from #307 was **unused and tree-shaken out of the build**; this is its first consumer. The org's own message stays the first line byte-for-byte and is never replaced; below it go the `errorCode`, the `fields[]` the org named, and one "what to do" line for ~25 common codes. An unrecognised code still renders the org's message **and** the code in full — an unknown code means we have no advice to add, not that we substitute something generic. Nothing is inferred: session advice can only attach to an error the org itself tagged `INVALID_SESSION_ID`, which is the whole point of the bug above.

## Audit (phase 2)

Every surface that shows an org error, what it did, and what changed.

| Surface | Org returned | User saw (before) | Matched? | Now |
|---|---|---|---|---|
| SOQL runner — REST run | 400 `MALFORMED_QUERY` | 401 session error | **no** | org message + code + field + guidance |
| SOQL runner — Tooling run | same path | same defect | **no** | fixed (same choke point) |
| SOQL runner — Explain | `err.message` | org text, no code/field | partial | + code/field/guidance |
| SOQL runner — GraphQL | HTTP 200 + `errors[]` | joined messages | yes | unchanged |
| SOQL runner — autocomplete | describe rejection | "Loading X metadata failed. Click to retry." | **no** | + org reason |
| `data-import` — CSV batch | SOAP fault | 401 session error / fault text | **no** | fixed + faultcode guidance |
| `data-import` — per row | `statusCode`/`message`/`fields` | all three rendered | yes | unchanged |
| `data-import` — describe/sobject list | `err.message` | org text | partial | + code/field/guidance |
| `field-creator` — deploy | `err.message` | org text in row status | partial | + code/field/guidance |
| `field-creator` — SObject list | anything | **"Failed to load SObject list"** (error discarded) | **no** | + org reason |
| `apex-anonymous` — compile/runtime | 200 + `compileProblem`/`exceptionMessage` | both rendered w/ line+col | yes | unchanged |
| `apex-anonymous` — transport | `err.message` | org text | partial | + code/field/guidance |
| `inspect-record` — load | `err.message` | org text | partial | + code/field/guidance |
| `inspect-record` — lookup nav | anything | **"Navigation failed"** (`catch {}`, unbound) | **no** | + org reason |
| `schema-browser` — object list | describe rejection | "Failed to load objects." | **no** | + org reason |
| `schema-browser` — field table | describe rejection | "Failed to load object describe." | **no** | + org reason |
| `metadata-retrieve` (SOAP) | faultstring | fault text via `addLog` | partial | + faultcode guidance |
| `rest-explore`, `soap-explore`, `org-limits`, `apex-test-runner`, `code-coverage`, `trace-flags`, `debug-log-viewer`, `event-monitor`, +19 more | `err.message` | org text only | partial | + code/field/guidance |

Three distinct defects, three separate commits:

1. **`sf-api-proxy` dropped the org's error** — the reported bug.
2. **`describe-cache` recorded `status:'error'` and threw the error object away**, so `schema-browser` and the SOQL autocomplete could only ever say "failed" with a retry button and no reason to think retrying would help.
3. **Two catches discarded the error entirely** — `field-creator`'s SObject list, `inspect-record`'s lookup navigation.

**None of them warrants splitting out.** (2) and (3) are the same failure class as (1) — a real cause replaced by a confident, uninformative one — they are small, and they share the guidance layer. Splitting would put the audit's conclusion in a different PR from the audit.

## Tests

Pinned to the reported case specifically. **All of these fail on `develop`:**

- `test/sf-error-guidance.test.ts` — drives the real path (client → worker proxy → stubbed org) with the exact reported query and the body Salesforce actually returns. Asserts the `MALFORMED_QUERY` text and code survive, and that the message contains **no** `Session expired`, no `INVALID_SESSION_ID`, no `401`, no "log in again". 4 fail on develop.
- `test/sf-session-resolution.test.ts` — 6 proxy-level regression tests. 3 fail on develop (`expected 401 to be 400`, `expected [401] to include 503`, and the mutating re-send: `expected 1 call, got 3`).
- `test/field-creator.test.ts` — 1 fails on develop (`expected '…Failed to load SObject list' to contain 'API access is disabled'`).
- Plus guards that must **not** regress: a genuine all-candidate 401 is still reported as a session error with re-auth advice; 5xx still falls back to the other host.

One existing assertion was intentionally changed: `salesforce-api.test.ts` asserted `.message` was byte-identical, which #307 added to protect its own additivity. Enriching that string **is** this change, so it now asserts the org's headline is preserved as line 1 and the annotation follows.

## Gates

Clean `npm ci` at the repo root first (`extension` is a workspace — installing inside it shadows the hoisted deps and breaks `@sfdt/flow-core`; `packages/flow-core` also needs building in a fresh worktree).

| Gate | Result |
|---|---|
| `extension` vitest | **1490 passed**, 88 files |
| `extension` `tsc --noEmit` | exit 0 |
| `extension` `eslint .` | exit 0 |
| `extension` `wxt build` | exit 0, 1.29 MB |
| root `npm test` | **4774 passed**, 289 files |
| root `npm run lint` | exit 0 |
| root `npm run check:all-contracts` | exit 0 |
| root `npm run generate:catalogs` | **zero drift** |

No new runtime dependency, no new manifest permission, no `innerHTML`, colours via `var(--sfdt-color-*)`.

---


scoobydrew83 (commented):
## Verdict: mechanism verified, fix is correct, no blocking findings — 4 should-fix, 3 nits

Independent review. I did not take the PR body's word for anything load-bearing: I rebuilt from a clean root `npm ci`, built `packages/flow-core`, reproduced the warm/cold control myself, and re-ran every gate.

---

### The mechanism — independently reproduced ✅

I wrote my own probe (client → worker proxy → stubbed org, real `SalesforceApiClient`, real `sfApiFetch`, only the cache seeding differing) and ran it against **both** the PR tree and a tree with all eight production files reverted to `develop`.

Reverted (`develop`) production code:

```
WARM >>> Salesforce GET request failed (HTTP 401): Session expired or invalid
COLD >>> Salesforce GET request failed (HTTP 400): value of filter criterion for field
         'Invoice_Status_Type__c' must be of type double and should not be enclosed in quotes
```

PR tree — both states:

```
Salesforce GET request failed (HTTP 400): value of filter criterion for field
'Invoice_Status_Type__c' must be of type double and should not be enclosed in quotes
MALFORMED_QUERY — Check the SOQL syntax and the value types in the WHERE clause — text and id
values are quoted, but number, boolean and date values must not be.
```

That is the reported string, byte-for-byte, appearing and disappearing on nothing but cache warmth. The diagnosis is right: a dropped error in the `sfApiFetch` fast path — not a refresh, not a retry, not the proxy boundary. `deriveBaseUrls` returns `[…my.salesforce.com, …lightning.force.com]` in that order, and the second candidate is the one its own comment says "often 401s". Everything downstream rests on a sound foundation.

---

### The regression you asked me to hunt: non-401 4xx from the wrong host

**It is a real behaviour change, and I could not make it cost a user a working request.** Both halves matter, so here is the evidence.

The change is real. Warm cache on `…my.salesforce.com`, that host answers 404 (or a WAF-style HTML 403), the other candidate would have answered 200:

| | `develop` | this PR |
|---|---|---|
| 404 on cached host, alternate healthy | `ok:true` via `lightning.force.com`, 2 fetches | `ok:false` 404, **1 fetch** |
| 403 HTML (no SF body) on cached host, alternate healthy | `ok:true`, 2 fetches | `ok:false` 403, **1 fetch** |
| 404 on first host, **cold** cache | `ok:true`, 2 fetches | `ok:true`, 2 fetches (unchanged) |

So the loss of resiliency is confined to the warm fast path — `isDefinitiveOrgFailure` is only consulted at `sf-api-proxy.ts:299`, and full resolution (`runOnce` over all candidates) is untouched. Good scoping.

Now reachability, which decides severity. To reach the new early return you need **all** of:

1. a cache entry for this page host — and `deps.cache.set` is only ever called on `result.ok` (`sf-api-proxy.ts:342-345`), so the cached host *already served this org successfully* this browser session (`chrome.storage.session`);
2. the sid cookie still present on that host with a **matching org id** (`orgIdFromSid(sid) === cached.orgId`) — otherwise the entry is deleted at line 305;
3. that same host to then answer a non-401 4xx for a request the *other* host would have served.

I tried to build (3) and failed:

- **Endpoint asymmetry:** every endpoint in the extension is under `/services/…` (`/services/data/vNN/…`, `/services/Soap/{c,u,s,m,T}/…`). I enumerated them; there is no path one hostname serves and the other doesn't, so no "cached host succeeded on E1, 404s on E2" case exists.
- **My Domain rollout / enhanced domains:** the cache key is `pageHost`. A domain change changes the key. Unreachable.
- **Org switch / sandbox refresh:** org id changes → entry deleted at line 305. Unreachable.
- **Instance migration / maintenance:** 5xx, which still falls through (and there's a test for it).
- **Corporate proxy / WAF (zScaler, Defender):** the only shape I can construct. It requires the WAF to start 403-ing `…my.salesforce.com` *mid-session*, after it had already let a call through, while leaving `…lightning.force.com` open. Even then the pre-fix outcome was a fall-through to the host documented to "often 401" — usually a dead end with the *wrong* message rather than a working request.

**Ruling: should-fix, not blocking.** But the predicate is broader than the argument that justifies it. The comment at `sf-api-proxy.ts:217-228` reasons about *the org answering* — `MALFORMED_QUERY`, `INVALID_FIELD`, an FLS denial — yet the code tests only `status >= 400 && status < 500 && status !== 401`, which also matches a WAF HTML 403 and a routing 404 the org never saw. Tighten it to the thing you actually mean:

```ts
// only treat it as the org answering if the body IS a Salesforce error record
errors.every((e) => e.status >= 400 && e.status < 500 && e.status !== 401
                 && parseRestErrorDetails(e.errorText).length > 0)
```

(plus the `<faultcode>` equivalent for SOAP). That keeps 100% of the reported bug fixed — the `MALFORMED_QUERY` body parses — while an opaque 4xx from an intermediary keeps the old fall-through. It closes the only case I could construct, at the cost of one predicate. Worth doing before 0.11.0 given this reaches real users.

Everything else in the fast path checks out: the cache is correctly left intact on a definitive org failure (tested), `errors.length > 0` guards the vacuous-true `every()`, and the fast path only ever produces one error so the `every` can't be fooled by a mixed list.

---

### The other proxy behaviours you asked about ✅

| Scenario | Result |
|---|---|
| Genuine expired session (401 on every candidate) | `HTTP 401: Session expired or invalid` + `INVALID_SESSION_ID — … log in again` ✅ |
| Genuine dead host (503, alternate healthy) | falls back, succeeds, 2 fetches ✅ |
| Merged first failure (503 fast path + 401 fallback) | `HTTP 503: unavailable` + the 503 status guidance. **Not** doubled, **not** confusing — and strictly better than `develop`, which reported this as "Session expired" ✅ |
| Mutating call already rejected by the org | no longer re-POSTed to the second host ✅ |

`pickPrimary`'s tier-2 fallback is literally the old `find(e => e.status >= 400 && e.status !== 401)` rule, so nothing that used to be headlined stops being headlined. Verified.

---

### The guidance layer

**Line 1 preservation: confirmed.** `buildUserFacingMessage` only ever `push`es after `lines[0] = headline`, and `headline` is the pre-existing `Salesforce ${operation} failed (${summary})${detail}` string built the same way as before. Verified empirically across single-record, multi-record, empty `fields[]`, populated `fields[]`, unparseable HTML body, and a 260-character org message (no truncation — `extractErrorDetail` only truncates non-JSON bodies, so parsed org text survives in full).

**Unrecognised code loses nothing: confirmed.**
```
Salesforce GET request failed (HTTP 400): Something novel happened
TOTALLY_NEW_CODE
```
Message and code both present, no invented advice. Multi-record with a mix of known and unknown codes renders every record (`Also: …` + annotation), with the unknown one showing bare `BRAND_NEW_UNKNOWN_CODE`.

**All 24 code strings + 9 status strings read.** Nothing dangerous. Two would send a user down a false path:

- 🔧 **`REQUEST_LIMIT_EXCEEDED`** — asserts the 24-hour rolling allowance and points at *Setup › Company Information › API Requests, Last 24 Hours*. Salesforce returns this code for the **concurrent long-running request limit** too (`ConcurrentPerOrgLongTxn`). A user who hits the concurrent limit will open that page, see plenty of headroom, and conclude the extension is wrong. Hedge it: "…the org's rolling 24-hour API allowance, or its limit on concurrent long-running requests."
- 🔧 **`FIELD_CUSTOM_VALIDATION_EXCEPTION`** — "A validation rule on the object rejected the value … written by your admin." This code is *also* what Apex `addError()` on a field produces. Sending someone to Setup › Validation Rules when the text came from a trigger is exactly the wrong path. Note `CANNOT_INSERT_UPDATE_ACTIVATE_ENTITY` right below already says "a trigger or validation" — make this one consistent with it.

The other 22 are accurate, and the discipline is genuinely good: the `INVALID_SESSION_ID` advice is the *only* one that mentions re-authenticating, it can only attach to a record the org itself tagged, and there's a test (`never advises re-authenticating for a code that is not a session problem`) pinning that. `MALFORMED_QUERY`'s quoting rule is correct SOQL. `MALFORMED_ID`'s "15 or 18 characters" is correct.

🔧 **Newline handling is incomplete, and it's on the reported path.** The guidance embeds `\n`. The toast got `white-space: pre-line` and `schema-browser` got it — but `data-import.ts:715-717` puts the batch-level error into a `<td>` via `textContent` with `'padding: 6px 10px;'` and no white-space rule, so HTML collapses the newline and the guidance runs into the org's text as one sentence. `data-import`'s CSV batch path is one of the two paths users actually reported. One-line fix: add `white-space: pre-line;` to that cssText.

(Checked the rest: `soql-runner:966`, `rest-explore:155`, `soap-explore:218`, `apex-test-runner:142` already carry `pre-wrap`; `field-creator`'s `tdStatus.title` is a native tooltip, which renders newlines fine; and `csvSerialize` quotes every field unconditionally, so the failed-rows CSV export is still valid RFC-4180 with an embedded newline. No corruption there.)

**Design opinion — string-embedding vs structured rendering.** Right call *for this PR*; should not become the pattern.

Right for now, because: the parsed records are still on `SalesforceRestError.details`, so nothing is destroyed — the string is a projection, not the source of truth; line 1 is byte-stable so any prefix matching survives; and the alternative is touching ~60 call sites in the PR that also fixes a live user-facing bug, which is a far worse blast radius for a 0.11.0 fix. I count 63 `err instanceof Error ? err.message` sites, so the "51 org-facing" claim is credible.

But it does bake presentation into an error object, and the `data-import` `<td>` above is that tax coming due *within the same PR* — a `\n` in `.message` is now an implicit contract every current and future render site must honour, enforced by nothing. Worth noting that `lib/record-edit.ts:555-592` **already** consumes `.details` structurally to split field-level from banner-level errors, so the structured path isn't hypothetical, it's shipping. The end state I'd aim for is a shared `renderSfError(err, host)` that reads `.details` and lets each surface decide layout, with `.message` demoted to a fallback for the sites that never get converted. Not this PR — but I'd rather see that in the backlog than see the next feature add a 25th guidance string.

---

### The changed test — genuinely as strong, not a weakening ✅

The removed assertion was `expect(err.message).toBe('Salesforce PATCH request failed (HTTP 400): Value too long')`. The replacement:

```ts
const [headline, ...rest] = err.message.split('\n');
expect(headline).toBe('Salesforce PATCH request failed (HTTP 400): Value too long');
```

That is the **same `toBe`, on the same string**, scoped to line 1 — precisely the property #307 was protecting, and precisely the property that must survive enrichment. Plus three new `toContain`s on the annotation. Strictly stronger than what it replaced. `git diff --numstat` on `extension/test/*` gives `62/0, 18/0, 9/3, 284/0, 106/0` — the only three deleted lines are this one test's header and body. Nothing else removed, nothing else weakened.

---

### Do the regression tests actually fail without the fix? ✅ — and by more than claimed

This is the check that matters most, so I did it the hard way: reverted all eight production files to `origin/develop` (keeping the new `sf-error-guidance.ts` module so the unit tests could still resolve) and ran the PR's tests unchanged.

**13 failures**, including all three you named in `sf-session-resolution` verbatim:

```
sf-session-resolution  surfaces the org MALFORMED_QUERY…      expected 401 to be 400
sf-session-resolution  a mutating call … not re-sent          expected 1 time, got 3 times
sf-session-resolution  keeps the fast-path failure alongside  expected [ 401 ] to include 503
field-creator          surfaces the org reason…               expected '…Failed to load SObject list' to contain 'API access is disabled'
sf-error-guidance      × 5   (you said 4 — it's 5)
describe-cache         × 3
salesforce-api         × 1   (the changed assertion)
```

Every new test is load-bearing. Your count was conservative, not inflated.

---

### Gates — all reproduced exactly

Root `npm ci` + `packages/flow-core` build first (the workspace-shadowing trap in the PR body is real; a fresh worktree with no `node_modules` also silently resolves into the main checkout).

| Gate | Claimed | Measured |
|---|---|---|
| `extension` vitest | 1490 / 88 files | **1490 passed, 88 files** ✅ |
| `extension` `tsc --noEmit` | 0 | **0** ✅ |
| `extension` `eslint .` | 0 | **0** ✅ |
| `extension` `wxt build` | 0, 1.29 MB | **0, Σ 1.29 MB** ✅ |
| root `npm test` | 4774 / 289 files | **4774 passed, 289 files** ✅ |
| root `npm run lint` | 0 | **0** ✅ |
| root `check:all-contracts` | 0 | **0** ✅ |
| root `generate:catalogs` | zero drift | **zero drift** (`git status` clean after) ✅ |

Golden principles: no `innerHTML` added, colours via `var(--sfdt-color-*)` (`schema-browser:312`), no new manifest permission, no new runtime dependency, `generated/*` untouched, clean tree.

---

### Packaging of the two extra defects — agreed, keep them together

I'd have made the same call. `describe-cache` recording `status:'error'` with the error discarded, and the two bare catches (`field-creator`'s SObject list, `inspect-record`'s unbound `catch {}`), are not independent bugs — they're the *same* failure class as the 401, and more to the point they have **no user-visible value without the guidance layer in this PR**. `describe-cache.error` and `Navigation failed — ${err.message}` are only worth surfacing because the string they now carry has been annotated with the org's code and field. Split them out and you ship plumbing in one PR and its only consumer in another, with the audit that found them in a third. Separate commits with a shared audit table is the right granularity. Nothing here deserved its own PR.

---

### Summary

**Blocking:** none.

**Should-fix (I'd take these before 0.11.0):**
1. Narrow `isDefinitiveOrgFailure` to bodies that actually parse as Salesforce error records — the current status-range predicate is broader than the reasoning in its own comment, and dead-ends an opaque WAF 403 / routing 404 that used to fall back. *Failure scenario:* warm cache on `acme.my.salesforce.com`; a corporate proxy starts returning `403 <html>Blocked</html>` for that host while `acme.lightning.force.com` still works → previously the request succeeded on the second host, now it returns the 403 after a single fetch.
2. `REQUEST_LIMIT_EXCEEDED` guidance omits the concurrent-request limit and points at a page that will show headroom in that case. *Failure scenario:* user runs 11 concurrent long-running queries, gets `REQUEST_LIMIT_EXCEEDED`, is told to check *API Requests, Last 24 Hours*, sees 2% used, and concludes the tool is broken.
3. `FIELD_CUSTOM_VALIDATION_EXCEPTION` guidance attributes to a validation rule what is often Apex `addError()`. *Failure scenario:* a trigger rejects an import row; the user searches Setup › Validation Rules on that object, finds none, and has no next step.
4. `data-import.ts:717` — add `white-space: pre-line` so the guidance line doesn't run into the org's text in the failed-rows table. This is one of the two originally-reported paths.

**Nits:**
5. `STATUS_GUIDANCE` has no `400` entry, so a 400 with an unparseable body gets a headline and nothing else (verified). Deliberate-looking, but 400 is the most common status here.
6. SOAP path: an *unrecognised* `<faultcode>` yields no guidance **and** doesn't fall back to `guidanceForStatus`, unlike REST which at least still renders the code. Usually harmless because the faultstring normally leads with the code — but the two paths could state the same rule.
7. Confirm whether the `sfdt-site` staleness pass is owed for this before the 0.11.0 release, per the cross-repo duty in CLAUDE.md.

Good bug, correctly diagnosed, honestly evidenced, and the tests earn their place. My only real reservation is finding #1, and it's a one-predicate change.

---

scoobydrew83 (commented):
## Round 2 verdict: all seven round-1 items land. Clear to merge, with one 16th surface the sweep missed and one guard hole that hid it — both one-liners, neither blocking.

Round-1 findings (mechanism, reverted-tree failures, the strengthened `.message` assertion) stand and were not re-derived. New work only.

---

### 1. The module split — no cycle, no page-side code in the worker ✅

`lib/sf-error-body.ts` **imports nothing at all**. It is a true leaf, so a cycle is structurally impossible, not merely absent today. The split also *removed* an edge: `sf-error-guidance.ts` now takes its type from `sf-error-body.js` instead of `salesforce-api.js`.

Worker bundle, checked against the built artifact rather than the source:

| symbol in `.output/chrome-mv3/background.js` | count |
|---|---|
| `SalesforceApiClient`, `SalesforceRestError` | **0** |
| guidance strings (`Check the SOQL syntax`, `concurrent long-running`) | **0** |
| `DOMParser` | **0** |
| the SOAP probe regex (`fault(?:string\|code`) | **1** ✅ |

So the worker got exactly the primitive it needed and nothing else, and the "MV3 workers have no `DOMParser`" justification is confirmed by the bundle rather than asserted. Re-export preserves every existing path — `lib/record-edit.ts` and `test/record-edit.test.ts` still import `SalesforceRestErrorDetail` from `salesforce-api.js` unchanged.

**The regex probe survives attack.** I threw 24 bodies at `isSalesforceErrorBody`:

| attack | result |
|---|---|
| `<faultstringy>` (word-boundary evasion) | rejected ✅ |
| `&lt;faultstring&gt;` (HTML-escaped) | rejected ✅ |
| `<soap-env:` / `<soap.env:` prefixes | rejected — false *negative*, i.e. falls back as before, the safe direction ✅ |
| echoed SOAP **request** body | rejected (requests carry no fault elements) ✅ |
| Cloudflare / zScaler / nginx HTML | rejected ✅ |
| OAuth `{"error":…}`, empty, plain text, `null`, `42`, `["nope"]` | rejected ✅ |

Only two positives I could manufacture were `<FAULTCODE>` in caps (SOAP is case-sensitive; nothing emits it) and an HTML page containing a literal unescaped `<faultstring>`. Neither is realistic. **The regex is not the weak point.**

**The residual false positive is the JSON path, not the regex.** `toErrorDetail` requires only a `message` string, so an intermediary that answers in JSON classifies as the org:

```
ORG      {"message":"Forbidden"}                     ← AWS API Gateway 403
ORG      {"message":"Missing Authentication Token"}  ← AWS API Gateway 404
ORG      {"statusCode":403,"message":"Access denied…"} ← Azure APIM
not-org  <html>…Blocked by corporate proxy…</html>   ← fixed ✅
not-org  <html>…404 Not Found…</html>                ← fixed ✅
not-org  (empty body)                                ← fixed ✅
```

End-to-end, warm cache, alternate host healthy: WAF HTML 403 / nginx 404 / empty 4xx now **fall back and succeed in 2 fetches** — the exact scenario I named in round 1 is closed. `{"message":"Forbidden"}` still dead-ends in 1. That residual is narrower than the one I already called non-blocking, and TLS-inspecting proxies in front of `*.my.salesforce.com` emit HTML, not JSON. **Nit, not should-fix:** if you want it exact, require a non-empty `errorCode` for *classification* only, leaving `parseRestErrorDetails` permissive for *rendering* — every documented Salesforce rejection body carries one, and every fixture in this PR would still pass.

**Load-bearing check reproduced.** Stubbed `isSalesforceErrorBody` to `return true`, ran the full suite: `Tests 3 failed | 1502 passed`. Exactly the three opaque-body tests. The other two new proxy tests (Salesforce-explained 404, SOAP fault) correctly still pass under the stub because they assert the dead-end that `return true` also produces — so "exactly 3" is the right expectation, not an undercount. The 404 pair — same status, opposite handling — is the sharpest test in the PR; it pins body-not-status directly.

---

### 2. Is the fix global, or one funnel? — **one funnel, and it is the right one**

Every org-bound request in the extension is constructed in `lib/salesforce-api.ts`. There are exactly **three** `action:'sfApiFetch'` sends and exactly **one** raw `fetch()` in the entire source outside the proxy (`background.ts:440`, the localhost bridge ping — not an org route).

| Route | Through the fixed fast path | Gets the guidance |
|---|---|---|
| REST reads — `proxyText` (`apiGet`, `query`, `toolingQuery`, describes) | **yes** | **yes** — `buildRequestError` |
| REST writes — `apiRequest` (POST/PATCH/PUT/DELETE) | **yes** | **yes** |
| SOAP — `apiSoap` (Partner, Metadata, user-typed via `soap-explore`) | **yes** (`kind:'soap'`) | **yes** — faultstring + `guidanceForErrorCode(faultcode)` |
| Metadata retrieve/deploy | **yes** — all six calls go through `api.apiSoap`; **no private SOAP writer exists** | **yes** |
| Streaming — `sfApiStream` (event-monitor CometD) | **no** — separate Port, own `fetchImpl`, one resolved baseUrl, no session cache | **no** — `HTTP Error ${status}: ${statusText}`, body discarded |
| CLI bridge / native messaging | **no** — localhost/native, not an org route | n/a |
| Feature making its own `fetch` | **none exist** | — |

**So "this fixes everything" is true for the bug and 6-of-7 for the guidance.** The one org route outside the fast path is the CometD stream, and it *structurally* never had this bug — no dual-host fallback, no cache, so there was no second host to displace the first. It also gets no guidance, but that is pre-existing and independent: `sf-stream-worker.ts:199` throws away the response body entirely, so it would show a bare status line with or without this PR. **Correctly out of scope** — fixing it means reworking the Bayeux client's error construction, and bundling that here would widen a release fix for no user in the reported path. Name it as follow-up.

**On SOAP specifically:** yes, SOAP shares the fast path, so SOAP faults were being dropped identically — and a fault is not a JSON record, which is exactly why the predicate needs the regex arm. Proven, not reasoned (below).

**Catch-and-reformat sweep: clean.** 38 unbound `catch {` in `features/` — all are either clipboard failures (`navigator.clipboard`, a browser API, not an org error) or deliberate best-effort fallback chains (`trace-flags` userinfo→chatter, `api-version-audit`/`org-release-badge` degrading to a default release, `export-for-prompt` optional describes), every one commented as such. The only bound catch that substitutes copy is `field-impact.ts:805` — and there `status.textContent = 'Failed'` is a *status label*, with the real error going to an adjacent panel that already carries `white-space: pre-line`. The audit's three were the three.

---

### 3. CSV import path, driven end to end (not inferred) ✅

Real `SalesforceApiClient` → real `sfApiFetch` → stubbed org, warm cache on `…my.salesforce.com`, org answers the SOAP POST with HTTP 400 `sf:INVALID_FIELD`, other host answers 401. Clicked through paste → SObject → **Start Import**:

```
--- failed-rows cell ---
INVALID_FIELD: No such column 'Emial' on entity 'Account'. If you are attempting to use a
custom field, be sure to append the '__c' after the custom field name.
That field is not on the object, or your user cannot see it. Check the API name (custom
fields end in __c) and your field-level security.

cssText: padding: 6px 10px; white-space: pre-line; color: var(--sfdt-color-error-text); …
SOAP fetches to my.salesforce.com: 1
fetches to lightning.force.com:    0
```

The fault reached the user verbatim, the guidance is on its own line, and **the request was not re-sent to the second host** — the reported CSV-path bug, fixed, driven rather than reasoned. Note the SOAP branch correctly appends *only* the guidance, no restatement of the code, because the faultstring already leads with it.

Per-row errors unchanged and correct: `REQUIRED_FIELD_MISSING: Required fields are missing: [Name] [Name]` — statusCode, message and `fields` all present.

Failed-rows CSV export re-verified **with** a multiline message (round 1 only reasoned about `csvSerialize`):

```
"Acme One","1","INVALID_FIELD: No such column 'Emial' on entity 'Account'.
INVALID_FIELD — That field is not on the object, or your user cannot see it."
```

Every field unconditionally quoted, so the embedded newline is legal RFC-4180 and the re-import target is intact. No truncation or reformatting anywhere on the CSV path.

---

### 4. The newline sweep — 15 of 16, and the guard is why the 16th is invisible

🔧 **`ui/health-modal.ts:149` is a live 16th surface.**

`features/flow-health-check.ts:181` catches around `api.getFlowMetadata(flowId)` — a Tooling call, so `err.message` is now the annotated multiline string — and hands it to `getModal().showError(...)`, which does:

```ts
const msg = doc.createElement('div');
msg.className = 'sfdt-health-error-message';
msg.style.marginTop = '8px';          // ← no white-space rule
msg.textContent = message || 'Unknown error';
```

The class appears nowhere but as a test selector, so there is no stylesheet rule either. *Failure scenario:* user on a Flow Builder canvas runs Flow Health Check; their profile lacks Tooling access to `Flow`; org returns 403 `INSUFFICIENT_ACCESS`; the modal renders the org's sentence and *"Ask an admin to check your profile and permission sets…"* collapsed into one run-on line — the precise defect just fixed on 15 other surfaces. One line: add `white-space: pre-line` to that element.

**The guard, judged hard.** It is *not* decoration — I replicated its logic over `features/` + `ui/`: 24 matches, 16 elements genuinely held to the rule, 6 literal-only, 1 count, **0 offenders**. The value/count distinction is real and correctly implemented (`${errors.length} flows could not be loaded` skipped, `${global.error}` caught), and it did catch a regression introduced inside this PR. Narrowing away from error-panel *styling* was right — that keyed off an incidental property and would have forced edits to unrelated code.

But the narrowing went one step too far, in **two** ways, and they compound into exactly the miss above:

- **`css === null` → silently skip.** The comment calls this a judgement call not being made. It is the hole that hides `health-modal.ts:149`: that element styles itself via `msg.style.marginTop`, not `cssText`, so the guard evaluates nothing and reports nothing. My replication shows `skipNoCss: 1` — and that one is the live offender.
- **The noun alternation is too narrow.** The comment says `message`/`msg` "only produced noise: a thrown error is always reached through `err`/`error`". That is demonstrably not so — the org-error funnels for three features are `function showError(message: string)`:

  ```
  features/soql-runner.ts:1144   errorPanel.textContent = message;
  features/rest-explore.ts:223   errorPanel.textContent = message;
  features/soap-explore.ts:285   errorPanel.textContent = message;
  + apex-test-runner:109/143, ai-assistant:212, bridge-tools:108, debug-log-viewer:114
  ```

  All of these **do** carry `pre-wrap` today (I verified each), so there is no live defect there — but the guard does not hold them. Drop `pre-wrap` from the SOQL runner's error panel tomorrow, on the very surface the reported bug appeared, and nothing fails.

Net: the guard enforces the contract on 16 of ~26 error-rendering sites. Adding `message|msg` to the alternation would **not** reintroduce the destructive-changes false positive — that came from matching on panel styling, and the banner renders no message variable, so the existing `LITERAL_ONLY` filter already excludes it. Cheap widening, real coverage gain.

---

### 5. Guidance strings ✅

Both corrections are right, and `REQUEST_LIMIT_EXCEEDED` is better than what I suggested — instead of merely hedging, it turns the Setup page into a *disambiguator*: *"…if that shows headroom, it is the concurrent limit, so let running queries finish and retry."* That converts the dead end into a diagnostic step. `FIELD_CUSTOM_VALIDATION_EXCEPTION` now reads "a validation rule **or an Apex trigger**", consistent with `CANNOT_INSERT_UPDATE_ACTIVATE_ENTITY` below it. Both pinned by tests asserting the specific words (`/concurrent/i` + `/24-hour/i`; `/trigger/i` + `/validation rule/i`), so they cannot quietly re-narrow. Nits 5 and 6 also addressed — `400` added to `STATUS_GUIDANCE`, and an unrecognised `<faultcode>` now falls back to the status advice instead of yielding nothing.

---

### 6. Gates — totals **and** deltas verified

| Gate | Claimed | Measured | Δ vs round 1 |
|---|---|---|---|
| `extension` vitest | 1503 / 89 | **1503 / 89** ✅ | +13 / +1 |
| root `npm test` | 4787 / 290 | **4787 / 290** ✅ | +13 / +1 |
| `tsc` / `eslint` / `wxt` / root lint / contracts | 0 | **0** (wxt Σ 1.29 MB) ✅ | — |
| `generate:catalogs` | zero drift | **zero drift**, `git status` clean ✅ | — |

The identical +13/+1 in both totals does confirm the extension suite runs inside the root suite. The +13 accounts exactly: `error-render-newlines` ×5 (the +1 file), `sf-session-resolution` ×5, `sf-error-guidance` ×3.

**On "is the guard's own test counted as coverage of the thing it guards"** — partly, and the PR is honest about it. Of the 5 tests in `error-render-newlines.test.ts`, **2** are the guard (the sweep, and the toast rule) and **3** test the detector itself (`the guard actually detects a missing rule`, `ignores fixed single-line copy`, `distinguishes an error VALUE from an error COUNT`). That ratio is defensible — a regex guard with no self-test is how you get a check that silently matches nothing — but it should be read as 2 tests of coverage, not 5. And per §4 those 2 cover 16 of ~26 sites.

---

### 7. `sfdt-site` — reasoning checks out, verified independently

I attached and grepped the site myself rather than take the report. No hits for `Session expired`, `INVALID_SESSION_ID`, `MALFORMED_QUERY`, `HTTP 401` anywhere in `content/` or `data/`. `content/chrome-extension/troubleshooting.mdx` exists but is entirely bridge / native-host / kill-switch / telemetry / console-logging; its "Common issues" table has no Salesforce-API-error row. Nothing stale, so nothing owed. The inverse argument is also sound: documenting a symptom 0.11.0 removes would be adding staleness, not clearing it.

Unrelated pre-existing bug spotted while in there, since a staleness pass is exactly what should catch it: that page tells users the extension tags console messages with **`[SFUT]`** (three times, incl. `[SFUT] Shell mounted.`). The code emits **`[SFDT]`** — zero occurrences of `SFUT` in the extension source. Worth a separate one-line PR on `sfdt-site`.

---

### Verdict

**Blocking:** none.

**Should-fix (one line, before 0.11.0):**
1. `ui/health-modal.ts:148` — add `white-space: pre-line`. It is the 16th surface, reached from `flow-health-check.ts:181` with an annotated Tooling error, and it is invisible to the guard.

**Nits:**
2. Widen the guard's noun alternation to include `message`/`msg` (the `LITERAL_ONLY` filter already excludes the banner that caused the original false positive), and reconsider `css === null → skip`, which is what hid finding 1.
3. Optionally require a non-empty `errorCode` for *classification* in `isSalesforceErrorBody`, so a JSON-emitting gateway (`{"message":"Forbidden"}`) is not read as the org. Rendering stays permissive.
4. Follow-up, not this PR: `sf-stream-worker.ts:199` discards the response body, so event-monitor stream failures show a bare `HTTP Error 400: Bad Request`. It never had this bug; it also never gets the guidance.
5. Separate `sfdt-site` PR: `[SFUT]` → `[SFDT]` on the Chrome-extension troubleshooting page.

The scope growth was earned. The module split is the right shape and provably free of the cycle it was avoiding, the CSV path now has evidence instead of inference, and the guard — while under-covering — caught a regression this PR itself introduced, which is more than most guards do. Ship it.

---

scoobydrew83 (commented):
## Head is clean to merge. ✅

Scoped to `c24b8d9`, `2d2fef9`, `bc2bd05`.

---

### First: the correction is right and I was wrong

`features/debug-log-viewer.ts:114` — the confirm dialog's `msg` — carries **no white-space rule**:

```ts
const msg = doc.createElement('p');
msg.textContent = opts.message;
msg.style.cssText = 'margin: 0 0 12px; font-size: 13px; color: var(--sfdt-color-text-weak);';
```

My round-2 line "All of these **do** carry `pre-wrap` today (I verified each)" was not true of that one. The failure was methodological: my round-1 grep hit `debug-log-viewer.ts:332` — the log `<pre>` pane, `white-space: pre-wrap` — and I let a **file**-level hit stand in for an **element**-level claim, then wrote it up as if I'd checked each element. Different element, different rule, wrong claim.

I re-audited all nine sites element-by-element rather than assume the rest were fine:

| site | has a white-space rule? |
|---|---|
| `soql-runner.ts` (`errorPanel`) | yes ✅ |
| `rest-explore.ts` (`errorPanel`) | yes ✅ |
| `soap-explore.ts` (`errorPanel`) | yes ✅ |
| `apex-test-runner.ts` (`panel`, `msg`) | yes ✅ |
| `ai-assistant.ts` (`panel`) | yes ✅ |
| `bridge-tools.ts` (`panel`) | yes ✅ |
| **`debug-log-viewer.ts` (`msg`)** | **no** ← my error |
| `field-impact.ts` (`tdDetail`) | no — but renders `row.detail`, an analysis string, never a thrown error; correctly outside the alternation |

So: 7 of 8 claims held, one didn't. The *conclusion* survived — `confirmDialog` is defined at :88 and has exactly **one** caller, at :462, passing the literal `` `Delete ${ids.length} ${noun}?` `` — so it genuinely cannot receive a thrown error, which is why exemption rather than a `pre-line` was the right disposition. But the verification behind that conclusion wasn't what I said it was, and the author was right to check it.

---

### The `EXEMPT` mechanism — judged by mutation, not by reading

I'd have made the same call. Adding `pre-line` to an unrelated confirm dialog to silence a check is how checks stop meaning anything, and re-narrowing to dodge one false positive would have re-opened the exact hole that hid the 16th surface. A named, reasoned exemption is the correct third option.

**Is the staleness test load-bearing?** I renamed the exempted identifier `msg` → `promptEl` and re-ran:

```
× every exemption still matches real source
  AssertionError: stale exemption: features/debug-log-viewer.ts no longer assigns msg.textContent
  + "features/debug-log-viewer.ts:114 (promptEl)"
```

Both fire. That pairing is what makes it work: the exemption fails loudly **and** the site simultaneously loses its cover and reappears as an offender under the new name. It cannot rot into a silent hole by relocation, rename, or deletion.

**Its honest limit:** it pins *location*, not *safety*. If that dialog kept the name `msg` and the same assignment shape but were repurposed to receive `err.message`, nothing fires. That's the standard ceiling of an exemption list, not a defect here — but note the `because` text already names a mechanically checkable fact ("its one caller passes the literal"), so asserting *that* (one call site, literal argument) would close the gap. Nit, not a blocker.

**Can a future entry be added without this scrutiny?** Yes — append and the suite is green; the only friction is `because.length > 40`, a prose-length check, not a review gate. Cheap hardening: `expect(EXEMPT).toHaveLength(1)`, so growing the list forces an explicit, reviewed change to the bound rather than a quiet append. Also a nit.

**Is one exemption the right price for the three funnels?** Plainly yes, and I mutation-tested the payoff. Dropping `pre-wrap` from the SOQL runner's error panel — the surface the reported bug appeared on — now fails **twice**:

```
× holds the showError() funnels the org error actually flows through
```

The second test hard-codes `soql-runner` / `rest-explore` / `soap-explore` by filename, so it survives any future tuning of the regex. That is what makes this a genuine widening rather than a re-narrowing dressed up as one.

**Coverage is better than 19.** Recomputing the guard's own logic independently: **38 matched → 11 literal-only, 1 count, 1 exempt, 25 held, 0 offenders.** Up from 16 held in round 2 — +9 sites, including all three `showError` funnels and `ui/health-modal.ts:153`. One documented exemption against 25 enforced.

---

### The other two commits

`ui/health-modal.ts` — the 16th surface now sets `margin-top` and `white-space: pre-line` through `cssText`, so it is both fixed and *visible* to the guard (there's a dedicated test asserting exactly that, since the old `css === null → skip` is what hid it).

`bc2bd05` — classification now requires a non-empty `errorCode`; rendering stays permissive. Both halves pinned: three `it.each` cases prove AWS API Gateway / Azure APIM bodies fall through and try the second host, and `renders a body with no errorCode in full, even though it is not the org` pins `parseRestErrorDetails` staying loose. That second test is the important one — it's what stops a later "make these consistent" tidy-up from collapsing the asymmetry and silently re-introducing the dead-end.

---

### Gates — deltas and bundle re-verified

| Gate | Claimed | Measured |
|---|---|---|
| `extension` vitest | 1510 / 89 | **1510 / 89** ✅ (+7 / +0) |
| root `npm test` | 4794 / 290 | **4794 / 290** ✅ (+7 / +0) |
| tsc / eslint / wxt / root lint / contracts | 0 | **0** (wxt Σ 1.29 MB) ✅ |
| catalogs | zero drift | **zero drift**, tree clean ✅ |

+7 accounts exactly: **+3** guard (`showError` funnels, health-check modal, exemption staleness), **+3** gateway `it.each` cases, **+1** render-permissive.

Bundle claim re-verified against the freshly built artifact after the tightening — `SalesforceApiClient`/`SalesforceRestError` **0**, `DOMParser` **0**, guidance strings **0**, SOAP probe regex **1**. Unchanged.

*One process note, since it would otherwise look like a finding:* my first root run reported `2 failed / 4792 passed`. Those two were my own mutation experiments overlapping the run, not a real failure — a clean re-run on a verified-clean tree gives **4794 / 294 → 4794 passed, 290 files**. Flagging it so nobody re-derives a phantom.

---

### Verdict

**Clean to merge.** No blocking findings, and nothing outstanding I'd hold a release for. Two optional hardening nits on the exemption mechanism if you ever add a second entry: assert the single-literal-caller property the `because` already claims, and cap `EXEMPT.length` so the list can't grow quietly.

---

PR #309, opened 2026-07-31
release(ext): @sfdt/extension 0.11.0

scoobydrew83:
Extension-only release. **`extension/package.json` 0.10.0 → 0.11.0.** The CLI stays at **0.21.0** and `@sfdt/flow-core` stays at **0.10.0** — flow-core has no standalone trigger and rides the CLI's release commit (`RELEASING.md` §1, and the §2 gate that now forces the check). The root `CHANGELOG.md` `[Unreleased]` entry for flow-core's `field-writes` module is untouched and stays unreleased.

---

## What ships in 0.11.0

Verified against the merge history since the 0.10.0 release commit (`726a4a0`), not taken on trust from the changelog.

### Added
| | Merged as |
|---|---|
| **P4-4 Field Impact Analysis** — "what writes this field?" across Flows (parsed via flow-core's `extractFieldWrites`), workflow field updates and Apex, each row badged confirmed/inferred | #306 |
| **P4-3 SOSL mode in the SOQL Runner** — SOQL\|SOSL radiogroup with auto-detection, per-object result grouping and export, language recorded in history/bookmarks | #303 |
| **P4-1 PR-1 groundwork** — shared editability model + typed API errors. **No user-visible change**; ships tested so the edit path has one contract to obey | #307 |

### Fixed
| | Merged as |
|---|---|
| **#308 error surfacing** — a fallback host's 401 was replacing the org's real error, plus the guidance layer | #308 |
| **5-second write timeout misreporting as a lost session** — live risk on `data-import`, `field-creator`, `apex-anonymous` | #305 |
| **`sfdtKind` discriminant** on every API error | #305 |
| **`enabledByDefault` ignored at runtime** | #304 |
| **SOAP service paths** — see "found during the sweep" below | `b98f8ad` |

### Changed
| | Merged as |
|---|---|
| **Renamed "SFDT SF Helper" → "SFDT for Salesforce"** in the manifest `name` and toolbar title | `af30abf` |

**On #308 specifically:** its entry leads the Fixed section and is written so a user recognises their own bug report — it names the symptom they saw ("Session expired or invalid" on a healthy session), the surfaces it hit, and the user-facing value: *the extension now tells you what the org actually said*, and adds a "what to do" line for the error codes people actually hit.

---

## Found during the sweep

**Two user-facing changes had landed on `develop` since the 0.10.0 release commit with no changelog entry.** Both are already fixed in code on `develop` — what was missing was only the release note — so I wrote them up rather than stopping. Flagging them because release notes would otherwise have shipped without them.

1. **`b98f8ad` — every SOAP call was going to a malformed URL.** `SF_API_VERSION` carries its `v` prefix because that is what REST paths need, and it was being concatenated raw onto the SOAP service prefixes, producing `/services/Soap/m/v62.0`. Salesforce refuses that with `Invalid Api version specified on URL`. That is every SOAP surface in the extension: Metadata Retrieve & Deploy's **offline SOAP fallback**, the SOAP API Explorer, and Data Import's Partner `create`/`update`/`upsert` batches. It was long-standing rather than new — the deleted `final_14` store screenshot captured the error live — so 0.10.0's headline Metadata Retrieve & Deploy worked on the bridge path while the offline fallback it advertises could not have. Now stripped for the SOAP path only, REST untouched, path shape pinned per service by tests.

2. **`af30abf` — the manifest `name` and toolbar title still read "SFDT SF Helper"** while the listing, privacy policy, docs and UI had all moved on. Display text only; nothing to migrate. **But the Chrome Web Store item title is a dashboard field, not a manifest field** — see the human list below.

Also dropped three hand-maintained feature counts that the 45th feature had already falsified (the README's "40", a "44" in the changelog's `enabledByDefault` entry, and a "44" in a test comment).

---

## What I verified

| Gate | Result |
|---|---|
| `extension/` vitest | **1510 passed / 89 files** |
| `extension/` `tsc --noEmit` | clean |
| `extension/` `eslint .` | clean |
| `wxt build` | OK, 1.29 MB |
| version-match guard | `package.json` and built manifest both **0.11.0** |
| root `npm test` | **4794 passed / 290 files** |
| root `npm run lint` | clean |
| root `npm run check:all-contracts` | clean — catalogs (11 files), licenses (7 manifests / 2 license files / 19 prose), Node policy, auth docs (54 files), FEATURES.json, package-internal paths (117 files) |
| root `npm run generate:catalogs` | **zero drift** — tree clean after regeneration |

Environment note: root `npm ci` and `build:flow-core` were run in the worktree first, so the catalog comparison is real rather than resolving into the main checkout.

### Version scope
`generated/catalog-version.json` moved **only** `extensionVersion`. `cliVersion` (0.21.0), `flowCoreVersion` (0.10.0) and `bridgeProtocol` (1.3) are untouched, and `generated/packages.json` moved only the extension row. `generated/` was regenerated, never hand-edited (CLAUDE.md invariant #2).

### PRIVACY.md vs the built manifest
Diffed `extension/PRIVACY.md`'s permission table against `.output/chrome-mv3/manifest.json` from an actual build. **Permissions are unchanged, and the match is exact in both directions** — nothing claimed that isn't requested, nothing requested that isn't documented:

- `permissions`: `storage`, `clipboardWrite`, `cookies`, `contextMenus`, `sidePanel` — all five documented.
- `optional_permissions`: **none**.
- `host_permissions` (10): the four Salesforce hosts, the two GovCloud `.mil` hosts, `.sfcrmapps.cn`, `.mcas.ms`, `localhost`, `127.0.0.1` — all documented, grouped exactly as the table groups them.

**PRIVACY.md therefore needs no edit for this release.** Its "Last updated: 2026-06-25" is deliberately left alone: the only change to the file since then was the cosmetic rename, and bumping the date would signal a policy change to users when the policy did not change.

### Minimum `@sfdt/cli`
Checked rather than assumed. `PROTOCOL_VERSION` is **unchanged at 1.3** since 0.10.0, `bridge-contract.ts` has no diff, no request kind was added/removed/renamed, and every item in this release runs over the worker-proxied Salesforce API rather than the bridge. So **0.11.0 introduces no new requirement.** The changelog note records the standing one — 0.10.0's Metadata Retrieve & Deploy bridge path needs `@sfdt/cli >= 0.21.0`, which published to npm with that release and is already available.

### README
`RELEASING.md` §5 asks for README feature claims "sourced from the catalogs". It said **`## Features (40)`** over a **37-row** table against a **45-entry** catalog, missing `schema-browser`, `command-palette`, `trace-flags`, `field-impact` and others, with some listed ids no longer matching the catalog at all.

Per Drew's instruction this now **references rather than duplicates**, and the count is gone rather than corrected — a number maintained by hand is wrong again by the next release, which is the actual defect. It now points at `generated/chrome-features.json` (the machine-generated source of truth the docs site and skills pack already consume) and at the options page as the human-readable view of the same list.

I did not conclude a bare pointer was sufficient. The README still describes the feature set, as **eight prose categories of capability** worded so that adding or renaming an individual feature cannot make them false — a first-time reader still learns what the extension does before deciding to install. It also now names `RELEASING.md` §5's doc-staleness sweep as the owner of this file, `PRIVACY.md`, the listing and the screenshots.

### Store listing
The bullet list was already complete — **45 bullets against the 45-entry catalog**, including P4-4's. Stale around it: the "44 features" count (→ 45, annotated to be re-derived from the catalog rather than incremented), the SOQL Runner bullet which predated P4-3 (now covers SOSL, per-object grouping and export), and a sync header still saying "v0.8.0" after 0.9.x and 0.10.0 shipped.

The error-surfacing fix is deliberately **not** in the listing: it is a feature list with no fixes section, and a store listing is not where a bugfix belongs. It is in `extension/CHANGELOG.md`.

---

## Still needs a human

**None of these are done, and none can be done from this PR.**

1. **The `develop` → `main` promotion.** Deliberately not performed — that call is Drew's. Per `RELEASING.md` §3 it must be a **merge commit, never a squash**, followed by fast-forwarding `develop` back onto `main` until `git rev-list --left-right --count main...develop` reads `0	0`. That check is a release gate, not housekeeping. Merging to `main` is what triggers `extension.yml` to tag `ext-v0.11.0` and upload to the Web Store.

2. **Screenshots — not verified, and I could not verify them.** I cannot take screenshots in this environment, so I am not claiming this passed. What I *could* detect, and did:
   - `final_14` (metadata-retrieve) and `final_15` (soap-explore) are **deleted**, pending recapture — one showed a live API-version error, the other leaked a real org id, company name and email. The CWS source pool is smaller than the docs describe.
   - **The whole set is 0.3.x-era.** Nothing shipped since has a shot: no Schema Browser, Command Palette, Trace Flags, Field Impact Analysis, side panel, or SOSL mode.
   - `final_06`, the **hero** frame, shows a Workspace nav from when there were 13 tools.

   Whether the remaining frames still match current UI needs a human with a browser.

3. **Chrome Web Store item name.** The packaged manifest now reads "SFDT for Salesforce", but the store's **item title is a dashboard field** and will keep showing "SFDT SF Helper" until renamed by hand during the upload.

4. **The companion docs-site PR `scoobydrew83/sfdt-site#43`**, which per CLAUDE.md's cross-repo duty ships with this release. `RELEASING.md` §4: Cloudflare Workers Builds auto-deploys `master`, so **merging there is deploying**. Includes the Step 2b catalog sync — regenerate `generated/` **from the released ref, never `develop`** — plus `/sync-docs-site` and merging any HELD draft docs PRs for features shipping here.

5. **The user-invoked pre-release skills**, which are not runnable from CI: `/pre-release-cli-test`, `/pre-release-security` (code outside docs/tests changed), `/validate-npm-paths`. `/pre-release-ui-test` is not needed — nothing in `gui/` changed.

---


scoobydrew83 (commented):
## Verdict: one blocking accuracy fix in the SOAP entry, one stale README claim to fix — otherwise this is sound, and the changelog **is** complete.

Independent review; I did not write this. Environment set up as instructed: **root** `npm ci` + `npm run build:flow-core` in a fresh worktree before touching any catalog or test.

---

## The changelog is complete — confirmed independently

I re-derived the baseline rather than trusting it. `ext-v0.10.0` points at `e683431`; nothing between the 0.10.0 release commit `726a4a0` and that tag touches `extension/`, so the tag is the correct diff base. Since it: **48 non-merge commits, 44 of which touch `extension/`**, plus 4 flow-core-only.

I walked all 48 and mapped each to an entry. **Every one is accounted for. There is no third missing change** — `b98f8ad` and `af30abf` are the complete set of unlogged user-facing work, and the author found both.

The judgement call I checked hardest was the long tail of follow-up fixes (`e841ba3`, `843543c`, `7ce2151`, `f8383fd`, `3e99365`, `bd6929b`, `63eee7c`, `31d4ff2`, `990168b`, `9c92d87`, `e52eaa0`, `c24b8d9`, `4b5d7ff`, `3d643fd`, `bc2bd05`, `39ee517`). Folding those into their parent entries is right: every one fixes code added inside this same unreleased window, so none is a regression from 0.10.0 and none needs its own line. The four that *do* touch shipped-in-0.10.0 behaviour — `32d30ab` (describe-cache dropped the failure reason), `0c01355` (field-creator + inspect-record swallowed the org's error), `1dca831` (fallback host's 401 replacing the real error), `f8b7e3d` (`enabledByDefault`) — are each named explicitly in the entries. Correct on both sides of that line.

---

## Blocking

### 1. The SOAP entry is wrong about what still worked in 0.10.0 — and it errs toward *understating* the bug

The mechanism is exactly as described (`b98f8ad`: `SF_API_VERSION = 'v62.0'` concatenated raw onto all five service prefixes; `v` now stripped for SOAP only; REST paths byte-identical). No dispute there.

The claim I can't sign off on is this one:

> 0.10.0's headline Metadata Retrieve & Deploy worked on the bridge path while the offline SOAP fallback it advertises could not have.

The bridge path never carried retrieve or deploy. In `extension/features/metadata-retrieve.ts` at `ext-v0.10.0`, both calls are **unconditional** — not gated on `!bridge`:

```
816:  const result = await api.apiSoap<any>('Metadata', 'retrieve', { retrieveRequest });
902:  const result = await api.apiSoap<any>('Metadata', 'deploy', { zipFile, deployOptions: reqOpts });
```

The file's own header comment says so outright: *"Retrieve/deploy always run over SOAP — the bridge has no retrieve kind."* And `SfdtRequestKind` in `packages/flow-core/src/bridge-contract.ts` confirms it — protocol 1.3 added only `manifest.discover` and `manifest.render`; the `deploy` kind listed there is `flow-deploy`'s, not this feature's.

So the bridge path covered **type/member discovery and `package.xml` rendering, and nothing else**. In 0.10.0 the malformed path broke *retrieve and deploy themselves*, on both paths. The feature could browse metadata and build a manifest; it could not do the two things in its name.

This matters because it's the direction that misleads a real person: someone who tried to deploy from 0.10.0, hit `Invalid Api version specified on URL`, and reads "worked on the bridge path" will conclude their failure was something else. Please reword to something like — *the bridge path could still browse metadata types and render `package.xml`; retrieving or deploying could not work on either path.*

The rest of the entry holds up: five service paths, the three affected surfaces (offline metadata path, SOAP Explorer, Data Import's Partner batches), long-standing rather than new, and `final_14` genuinely captured it (`store-assets/README.md` records `Invalid Api version specified on URL : /m/v62.0`). Nothing overstated anywhere — understated in this one place.

### 2. `extension/README.md:96` still says the extension isn't on the store

```
### From the Chrome Web Store

(Pending submission.)
```

`extension/listing.md` — edited in this same PR — says the opposite twice: *"the last manual store upload (0.3.x era)"* and *"the live CWS listing still reflects the older 29-feature copy."* Two docs in one PR disagreeing about whether the product is published is exactly the drift this PR set out to kill, and it survived a README pass whose stated purpose was staleness. Fix the README line (or, if the item is unlisted, say that instead).

---

## Should fix

### 3. "Worded so individual features can't falsify them" is met; "won't go stale" is only half true

I checked all eight categories against the 45-entry catalog. **No false claims** — every specific capability named is real: `setup-tabs` does group (`groupingEnabled` / `GROUP_LABEL = 'Automation'`), `debug-log-viewer` does open a profiler view (timings + limits + inventories), and `command-palette`, `field-impact`, `trace-flags`, `org-release-badge`, `context-menu-inspect`, SOSL mode and the side panel all exist. That part is good, and the count being gone rather than corrected is the right instinct.

But five of the eight (Navigation, Flow tooling, Apex/logs/events, Record/data, Bridge-backed) are hand-maintained enumerations of 5–9 named capabilities each — the old table with the ids stripped. A new feature won't make them *false*; it'll make them *silently incomplete*, which is the same drift minus the count that used to catch it. It has already happened, in the version being shipped:

- **`org-health`** (the bridge-backed audit/monitor snapshot panel) has no category. Cat 7 says "live org-health checks" — that's `org-health-live`, a different feature. Cat 8's bridge list is deploy / rollback / org compare / drift check / metadata scan / AI analysis, and omits it.
- **`flow-list-search`** has no home either — cat 2's "search across canvas nodes" is `canvas-search`.

Two options: fold in the strays, or soften "What those features cover" so it doesn't promise coverage. The former is better; the latter at least makes the doc honest about what it is.

---

## Nits

4. **Section order.** `[0.11.0]` runs Fixed → Added → Changed. `[0.10.0]` and the rest of the file run Added → … → Fixed. Leading with the user-reported bug is defensible, but it breaks the file's own convention and puts a very long Fixed block in front of the three new features.

5. **`af30abf` also changed CLI-side strings** this extension release can't carry: `sfdt extension install-host`'s `--help` description, and `ApiNameLibrary.exportAsJson()`'s embedded `description` in `packages/flow-core/src/api-name.ts` (which lands inside a file users export). Neither has a root `[Unreleased]` entry. Correctly out of scope here — flagging so the next CLI/flow-core release doesn't lose them.

6. **Store-listing exclusion.** Agreed on the error-surfacing fix: a feature list has no fixes section, and the changelog is the right home. Worth noting the SOAP fix sits differently — the bullet *"Metadata Retrieve & Deploy — retrieve and deploy metadata directly from the browser"* was advertising something that could not work. It becomes true in 0.11.0, so no listing edit is needed — but that's the right outcome by luck rather than by the "no fixes section" rule.

---

## Verified clean — reproduced, not taken on trust

**Version scope.** ext `0.11.0`; CLI `0.21.0`, flow-core `0.10.0`, vscode `0.6.0`, plugin `0.21.0` all untouched. Root `CHANGELOG.md` is not in the diff — the `[Unreleased]` flow-core `field-writes` entry is intact and still unreleased. `generated/catalog-version.json` moved `extensionVersion` only (`cliVersion`, `flowCoreVersion`, `bridgeProtocol: 1.3` unchanged); `generated/packages.json` moved the extension row only. I grepped `0.10.0` across the tree independently: every remaining hit is flow-core's own version or a `^0.10.0` consumer range. `tools/generate-catalogs.mjs:218,225` is the only reader of `extension/package.json#version` besides `extension.yml` at release time — so those two generated files really are the whole surface, and both regenerate rather than being hand-edited (invariant #2).

**Minimum CLI.** The reasoning holds. `git diff ext-v0.10.0..HEAD -- packages/flow-core/src/bridge-contract.ts` is **empty**; `PROTOCOL_VERSION` is `'1.3'`; `SfdtRequestKind` unchanged. I also checked the one thing that could have quietly introduced a requirement — flow-core's new `extractFieldWrites` is **bundled** into the build, not fetched over the bridge (its Flow-parsing internals are present in `.output/chrome-mv3/content-scripts/content.js`). No new requirement; the standing `>= 0.21.0` is correctly recorded, as `RELEASING.md` §5 requires.

**PRIVACY.md vs the built manifest.** Built it and diffed both directions myself. `permissions` is exactly `storage, clipboardWrite, cookies, contextMenus, sidePanel`; `optional_permissions` is **undefined**; 10 host permissions, every one documented in the table (lines 73–82) with the same grouping. Nothing claimed that isn't requested, nothing requested that isn't claimed. `wxt.config.ts` since 0.10.0 is the two name strings and nothing else, so "no new permissions" checks out at the source too. Leaving **Last updated: 2026-06-25** alone is the right call — the only diff to `PRIVACY.md` since 0.10.0 is the title line and one `chrome://extensions` reference.

**Store listing.** 45 bullets against a 45-entry `generated/chrome-features.json`. Counted both. The 44 → 45 fix, the SOSL / bookmarks / plans additions to the SOQL Runner bullet, and the v0.8.0 → v0.11.0 sync header all check out.

**Gates** — all re-run here after root `npm ci` + `build:flow-core`:

| Gate | Claimed | Observed |
|---|---|---|
| `extension` vitest | 1510 / 89 files | **1510 passed / 89 files** ✅ |
| root `npm test` | 4794 / 290 files | **4794 passed / 290 files** ✅ |
| extension `tsc --noEmit` | clean | clean ✅ |
| `npm run lint -w @sfdt/extension` | clean | clean ✅ |
| root `npm run lint` | clean | clean ✅ |
| `npm run build:ext` | OK, 1.29 MB | OK, **1.29 MB** ✅ |
| `npm run check:all-contracts` | clean | clean — 54 auth-doc files, 117 path files ✅ |
| `npm run generate:catalogs` | zero drift | `git status` empty after regen ✅ |
| `check-version-match.ts` | both 0.11.0 | both **0.11.0** ✅ |

**Nothing weakened.** The only test change in the entire PR diff is a comment in `feature-defaults.test.ts` ("all 44 ids" → "every shipped id"). No assertion removed, no threshold lowered, no file excluded.

---

## Scope

Items 1 and 2 are release-artefact accuracy — they belong in this PR, because a wrong claim here ships to the Web Store and to users. No code defect was found that needs fixing here; the SOAP and rename fixes are already correct on `develop`, and only the notes were missing. The "still needs a human" list (promote merge, screenshot recapture, CWS dashboard rename, `sfdt-site#43`, the user-invoked pre-release skills) reads accurate against `RELEASING.md` §2–§5, and I agree screenshots cannot be signed off from CI.


---

PR #325, opened 2026-08-05
fix(extension): renderSfError() — one path for a Salesforce error, plus the sweep that keeps it one (C-FIX-4)

scoobydrew83:
Backlog item **C-FIX-4** (Chrome ext, size S; depends on C-FIX-3 / #308).

PR #308 fixed sixteen surfaces **one at a time**. This removes what made sixteen possible.

## The residual scope, verified against the code

The row was written before extension v0.12.0 and part of its wording was already done. I checked each claim; the spec was accurate on all three points, with one correction and one addition.

| Spec said | Code says |
|---|---|
| "hoist the duplicated error-panel cssText" — **already done** | Confirmed. `lib/ui-styles.ts:845` defines `.sfdt-console` with `white-space: pre-wrap` at :853 and `.sfdt-console.sfdt-error` at :857. Not redone. |
| `ui/panels.ts:28` already exports `errorPanel(message, doc)` with `role="alert"` | Confirmed. |
| "~7 features import `../ui/panels.js`, ~15 still hand-roll the block, silently omitting `role="alert"`" | Confirmed, and it is **16**, not 15 — the spec's list missed `features/rest-explore.ts`. It was not a hand-roll (it already took its panel from `errorPanel()`), but it is the third `showError()` funnel and was left with the single-string affordance, so it is migrated with the other two. |
| "no helper renders the SALESFORCE-ERROR shape … `errorPanel()` takes a single string, which is exactly the affordance that let the collapse happen" | Confirmed, and it is the crux. See below. |
| "the existing guards pin the BEHAVIOUR but do not force the single code path" | Confirmed. A fresh hand-roll that happens to satisfy both passes both. |

**Correction to the spec's file list.** `features/metadata-retrieve.ts:1296` is *not* an org-error site — it is the static destructive-manifest banner, which carries **our own prose**. Per the item's own instruction (and the comment at `.sfdt-callout` in `lib/ui-styles.ts:599`) it moved to `.sfdt-callout.sfdt-bad`, not to the helper. So: **15 sites routed through the helper, 1 to the callout, plus `rest-explore` = 16 call sites touched.**

**One thing I did that the spec did not ask for, and why.** I deleted `errorPanel()` rather than keeping it beside `renderSfError()`. Two error builders with different DOM shapes is the duplication this item exists to remove, and the single-string signature *is* the named root cause — leaving it in place leaves the next author the shorter wrong path. Its two remaining callers (`schema-browser`, `debug-log-viewer`) are migrated; `UI-SYSTEM.md`'s layer-4 table is updated to match.

## What landed

**`ui/panels.ts` — `renderSfError()` / `setSfError()` / `clearSfError()`**

The org's own text and every line we appended render as **separate element nodes**, laid out as blocks. A block box cannot run into the one above it whatever CSS the host page brings, and a caller can no longer keep the wrong half because it no longer does the joining. `setSfError(el, …)` fills a pane the caller already owns — the SOQL/SOAP/REST explorers hide-and-show a pre-built panel, and the debug-log and Execute Anonymous panes re-purpose the console they render output into; rebuilding those nodes would break their layout, and leaving them to re-apply the classes is exactly the hand-roll being removed. `clearSfError()` takes `role="alert"` back off, because a reused pane that keeps it announces the *next* successful render as a failure.

**`lib/sf-error-guidance.ts` — `splitUserFacingMessage()` beside `buildUserFacingMessage()`**

The newline between the org's records is a **separator, not formatting**. Producer and consumer of that contract now sit in one file rather than agreeing by coincidence. `errorText()` also replaces the fourteen copies of `err instanceof Error ? err.message : String(err)`.

**A11y — the load-bearing part**

All 16 hand-rolls set the classes and omitted `role="alert"`. A failed query, retrieve or test run was a red box to a sighted user and silence to a screen reader. The builder cannot forget. Covered by `test/panels.test.ts` at the unit level and by a new end-to-end assertion in `test/org-limits.test.ts` that drives a real feature to failure and reads the role off the rendered panel.

**Both themes.** The two new classes use `var(--sfdt-color-text)` and inherit the rest; no literal, no theme special-case. The note deliberately stops rendering in the org's alarm colour — advice in the same red as the failure reads as more of the failure rather than as the way out of it. A test asserts the rule contains a token and no `#`.

**`test/sf-error-panel-contract.test.ts` — the missing piece**

No file outside `ui/panels.ts` may apply `.sfdt-console` + `.sfdt-error` to one element. Deliberately the **pair**, not `.sfdt-error` alone — that is also the red `.sfdt-pill`, which several features legitimately put on a status chip; a test asserts the guard does *not* flag a file that uses both classes on two different elements.

Per **golden principle #12** the sweep excludes the artifacts that define it — `ui/panels.ts` (the implementation) and `lib/ui-styles.ts` (the stylesheet declaring the classes), each listed by name **with the reason it cannot be a violation**, plus a test that fails a stale exclusion. `test/` is not scanned at all: a test asserting on the rendered pair describes the contract rather than violating it, and this file is the proof.

I verified the sweep actually fails rather than merely passing: re-adding a `classList.add('sfdt-console', 'sfdt-error')` to `features/org-limits.ts` produced
```
× no feature, entrypoint or lib module hand-rolls the error block
  expected [ 'features/org-limits.ts' ] to deeply equal []
```
and the probe was reverted.

**Existing guards updated, not weakened.** `error-render-newlines.test.ts` learns `renderSfError` as a panel builder, and its "holds the `showError()` funnels" case now asserts the **union** (`errorPanel.textContent = message` OR `setSfError(errorPanel, message`) — which is the lesson that file's own comments record twice: assert the union, not the mechanism. Pinned to the old shape it would have read the *strongest* of the three states as a regression. One `EXEMPT` entry was added for `loadingPanel()`'s `el.textContent = message`: it had been passing only because the error builder in the same file reused the variable name `el`, which is the same-name blind spot that file warns about — now it is a reviewed exemption instead of an accident.

## Call sites migrated

Through the helper (15): `trace-flags`, `debug-log-viewer` (×3 incl. the reused `logPane`), `dependency-explorer`, `ai-assistant`, `field-impact`, `apex-anonymous` (reused `resultPane`), `flow-quality`, `bridge-tools`, `code-coverage`, `soql-runner`, `flow-trigger-explorer-enhancer`, `soap-explore`, `apex-test-runner`, `org-limits`, `schema-browser` (×2), `rest-explore`.
To `.sfdt-callout.sfdt-bad` (1): `metadata-retrieve`'s destructive banner.

`grep -rn "sfdt-error" features/ ui/ entrypoints/ lib/` now returns only pill usages and the two defining artifacts.

## Global DoD

- **Zero `innerHTML`** — `createElement` + `textContent` throughout; a test asserts an org message is never interpreted as markup.
- **No new runtime dependencies, no new manifest permissions.** Neither was needed.
- **Vitest, happy-dom, per-feature pattern** — `panels.test.ts` rewritten for the new API (13 cases), new `sf-error-panel-contract.test.ts` (8 cases), one new end-to-end a11y case in `org-limits.test.ts`.
- **A11y checklist** — item 10 (labelling) / 11 (state exposed, not styled) are the load-bearing ones here and are MET by construction. Items 1–9 and 13 are **N/A**: this adds no overlay, no keyboard path and no new mount point; it changes the contents of panels that existing overlays already present.
- **Both themes** — MET, tokens only.
- **CHANGELOG** — entry under `[Unreleased]` in `extension/CHANGELOG.md`.
- **Docs site** — **no change needed, stated explicitly rather than skipped.** This is an internal refactor plus an a11y fix: no new command, subcommand, flag, config key, feature gate, MCP tool, or page. I grepped `sfdt-site/content/chrome-extension/` for "error panel", `role="alert"`, "screen reader" and "accessib" — nothing documents error-panel rendering, so there is nothing to bring in sync.
- **`generated/`** — untouched; `check:catalog-drift` reports "Catalogs up to date (11 files)".
- **`FEATURES.json`** — not touched.

## Gates — actual output

From `extension/`:

```
$ npx vitest run
 Test Files  109 passed (109)
      Tests  1896 passed (1896)
   Duration  28.80s

$ npx tsc --noEmit
tsc OK          # no output

$ npx eslint .
eslint OK       # no output

$ npx wxt build
√ Built extension in 2.699 s
Σ Total size: 1.36 MB
√ Finished in 3.248 s
```

From the repo root:

```
$ npm test
 Test Files  313 passed (313)
      Tests  5252 passed (5252)
   Duration  80.72s

$ npm run check:all-contracts
EXIT=0
```

Baseline on `origin/develop` before any change was 108 files / 1878 tests (after `npm install` at the repo root and `npm run build:flow-core` — the extension suite cannot resolve `@sfdt/flow-core` without it).

## Not verified

- **No manual browser run.** Everything here is happy-dom; I did not load `.output/chrome-mv3` against a real org, so the *visual* result of the org-text/guidance split (block spacing, the note's colour against the console surface) is asserted by CSS-rule tests rather than seen. Worth a look in both themes before release.
- **Clipboard behaviour across the new block boundary.** Chrome inserts a newline when copying across `display: block` elements, so selecting a panel should still yield the org text and the guidance on separate lines — but `element.textContent` no longer contains the newline, which is a real semantic change for any future code reading a panel's text back. Nothing does today.

## Defects spotted and deliberately left out of this PR

1. `ui/health-modal.ts` and `features/debug-log-viewer.ts:417` use `sfdt-error` as a *pill* variant while `severityPillClass()` returns bare class strings — correct, but the naming makes the pill and the panel look related when they are not. A rename to `.sfdt-pill-danger` would make the sweep's rule self-evident.
2. `features/ai-assistant.ts:204` still sets `resultArea.style.cssText = 'margin-top: 12px;'` — a magic pixel that should be a spacing token.
3. `features/metadata-retrieve.ts:1281-1288` builds two bare `<button>` elements with `createElement('button')` rather than `button()` from `lib/ui-controls.ts` — the hand-rolled-button pattern `UI-SYSTEM.md` calls out, in the file I touched.

---

### DIFF
--- a/extension/CHANGELOG.md
+++ b/extension/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to `@sfdt/extension` are documented here. Format follows [Ke
 
 ## [Unreleased]
 
+### Fixed
+
+- **A Salesforce error is announced to screen readers, on every surface.** Sixteen features built the org-error panel by hand — `createElement('div')`, `classList.add('sfdt-console', 'sfdt-error')`, `textContent = err.message` — and every one of them set the classes and left off `role="alert"`. A failed query, retrieve or test run was a red box to a sighted user and complete silence to anyone using a screen reader. All sixteen now come from one builder, which cannot forget the role. The panes that are *reused* (the debug-log console, the Execute Anonymous result pane) also give the role back when they stop carrying an error, or the next successful run would be announced as a failure.
+
+### Changed
+
+- **`renderSfError()` — one helper renders a Salesforce error, and nothing else may.** The previous shared builder took a single string, which was the affordance behind the sixteen one-by-one fixes in the last release: with the org's own text and our "what to do" line joined into one string, keeping them apart was a `white-space` rule each surface had to remember, and several instead resolved the ambiguity by throwing the org's real error away. `renderSfError()` / `setSfError()` emit the two as separate element nodes, so a block box cannot run into the one above it whatever CSS the host page brings, and a caller can no longer keep the wrong half because it no longer does the joining. The split (`splitUserFacingMessage`) now lives beside the join (`buildUserFacingMessage`) in `lib/sf-error-guidance.ts`: the newline between the org's records is a separator, not formatting, and a producer and consumer that agree about that by coincidence eventually stop agreeing. Our guidance also stops rendering in the org's alarm colour — advice in the same red as the failure reads as more of the failure rather than as the way out of it.
+- **The destructive-manifest banner is a callout, not an error console.** It carries our own warning prose, not an org error, which `lib/ui-styles.ts` has always distinguished at `.sfdt-callout`; wearing the error console's classes made the one *static* banner in the extension look like a live Salesforce failure and dragged it into every sweep aimed at the org-error path.
+
+### Added
+
+- **A contract sweep (`test/sf-error-panel-contract.test.ts`) that fails the seventeenth hand-roll.** The two guards that came out of the last release pin what a correct error panel *looks* like; a brand-new hand-rolled panel that happens to satisfy both still passes them, which is exactly how sixteen of them accumulated. This one pins the code *path*: `.sfdt-console` + `.sfdt-error` on one element is the Salesforce error block, and only `ui/panels.ts` may build it. The rule is the class **pair**, not `.sfdt-error` alone — that is also the red `.sfdt-pill`, which several features legitimately apply to a status chip. Per golden principle #12 the sweep excludes the artifacts that define it: the helper, the stylesheet declaring the classes, and this file, each listed by name with the reason it cannot be a violation, plus a test that fails a stale exclusion.
+
 ## [0.12.0] - 2026-08-03
 
 ### Added
--- a/extension/UI-SYSTEM.md
+++ b/extension/UI-SYSTEM.md
@@ -22,7 +22,7 @@ to.
 | 1. Tokens | `lib/tokens.ts` | Colour, spacing, radius, type, shadow — as CSS custom properties | Everything |
 | 2. Components | `lib/ui-styles.ts` | 107 reusable classes | every UI file |
 | 3. Icons | `lib/icons.ts` | 53 inline SVGs + feature-id → glyph map | 8 files |
-| 4. Controls | `lib/ui-controls.ts`, `lib/code-editor.ts`, `ui/panels.ts`, `ui/confirm-dialog.ts`, `ui/meter-card.ts`, `ui/node-graph.ts`, `ui/apex-limit-tiles.ts`, `ui/apex-log-console.ts` | `button()`, `field()`, `glyph()`, `setLabel()`, `setTone()`, `toolbar()`, `createCodeEditor()`, `errorPanel()`, `loadingPanel()`, `emptyPanel()`, `busyOverlay()`, `confirmDialog()`, `meterCard()`, `buildNodeGraphSvg()`, `createLimitTiles()`, `renderApexLogBody()` | every UI file |
+| 4. Controls | `lib/ui-controls.ts`, `lib/code-editor.ts`, `ui/panels.ts`, `ui/confirm-dialog.ts`, `ui/meter-card.ts`, `ui/node-graph.ts`, `ui/apex-limit-tiles.ts`, `ui/apex-log-console.ts` | `button()`, `field()`, `glyph()`, `setLabel()`, `setTone()`, `toolbar()`, `createCodeEditor()`, `renderSfError()`, `setSfError()`, `loadingPanel()`, `emptyPanel()`, `busyOverlay()`, `confirmDialog()`, `meterCard()`, `buildNodeGraphSvg()`, `createLimitTiles()`, `renderApexLogBody()` | every UI file |
 | 5. Behaviour | `ui/menu.ts`, `ui/shadow-host.ts`, `ui/present-view.ts` | Dismissal, **focus trap**, focus restore, mounting | Menus, injected UI, every feature view |
 
 **Layers 2 and 4 are not the same thing, and layer 2 alone did not work.** The
@@ -157,6 +157,47 @@ Guarded by `test/ui-controls.test.ts` (12 tests) and a drift guard in
 `test/ui-styles.test.ts`: a **migrated** file may not contain a cssText string
 with `padding` + `border-radius` + `cursor: pointer` — the hand-rolled-button
 signature. Files join `BUTTON_MIGRATED` in the same commit that converts them.
+
+#### Layer 4, continued — the Salesforce error panel (`ui/panels.ts`)
+
+```
+renderSfError(error, { doc?, guidance? })   →  a new panel
+setSfError(el, error, { doc?, guidance? })  →  fill a panel the caller owns
+clearSfError(el)                            →  empty it, and drop role="alert"
+```
+
+The same lesson as `button()`, learned the same way. `.sfdt-console.sfdt-error`
+had existed since the design-system pass and **sixteen** features still built
+the block by hand, because `createElement('div')` + `classList.add(…)` +
+`.textContent = err.message` was the shorter path. All sixteen omitted
+`role="alert"`, so a failure was a red box to a sighted user and silence to a
+screen reader; PR #308 then had to fix each of them individually for a
+*separate* defect — collapsing our guidance line into the org's own text, and in
+several cases discarding the org's error entirely.
+
+Two things these enforce that a class cannot:
+
+- **`role="alert"` is not optional.** It comes with the block, including on
+  `setSfError`, which is what a long-lived pane (the debug-log console, the
+  Execute Anonymous result pane) uses when it turns into an error surface.
+  `clearSfError` takes the role back off, or the pane announces the *next*
+  thing rendered into it as a failure.
+- **The org's text and our guidance are separate NODES.** The joining happens in
+  `lib/sf-error-guidance.ts` (`buildUserFacingMessage`) and the splitting in the
+  same file (`splitUserFacingMessage`), so producer and consumer of the newline
+  contract sit next to each other. A caller no longer composes the string, so it
+  can no longer keep the wrong half — and a block box cannot run into the one
+  above it whatever `white-space` the surface carries.
+
+`.sfdt-console.sfdt-error` is for the **org's** text. Our own prose — a
+destructive-mode caution, a truncation warning — is `.sfdt-callout`; the comment
+at that rule in `lib/ui-styles.ts` draws the line. The destructive-manifest
+banner in `metadata-retrieve` was on the wrong side of it and moved.
+
+Guarded by `test/sf-error-panel-contract.test.ts`: **no file outside
+`ui/panels.ts` may apply the class pair.** That is the piece the two behavioural
+guards (`error-render-newlines`, `sf-error-guidance`) could not supply — they
+pin what a correct panel looks like, and a fresh hand-roll can satisfy both.
 
 ### Layer 5 — behaviour (`ui/menu.ts`)
 
--- a/extension/features/ai-assistant.ts
+++ b/extension/features/ai-assistant.ts
@@ -5,6 +5,7 @@ import { getSalesforceApi, type SalesforceApiClient } from '../lib/salesforce-ap
 import { loadSettings } from '../lib/settings.js';
 import { createBridgeClient, LONG_RUNNING_TIMEOUT_MS } from '../lib/sfdt-bridge.js';
 import { presentView, type ViewHandle } from '../ui/present-view.js';
+import { renderSfError } from '../ui/panels.js';
 import type { SfdtRequest, SfdtResponse } from '@sfdt/flow-core/bridge-contract';
 import { button } from '../lib/ui-controls.js';
 import { copyToClipboard } from '../ui/clipboard.js';
@@ -204,11 +205,7 @@ export function createAiAssistantFeature(options: AiAssistantOptions = {}): Feat
       resultArea.style.cssText = 'margin-top: 12px;';
 
       const renderResultError = (message: string): void => {
-        while (resultArea.firstChild) resultArea.removeChild(resultArea.firstChild);
-        const panel = doc.createElement('div');
-        panel.classList.add('sfdt-console', 'sfdt-error');
-        panel.textContent = message;
-        resultArea.appendChild(panel);
+        resultArea.replaceChildren(renderSfError(message, { doc }));
       };
 
       const renderResult = (responseText: string, provider: string): void => {
--- a/extension/features/apex-anonymous.ts
+++ b/extension/features/apex-anonymous.ts
@@ -18,6 +18,7 @@ import { loadSettings, registerSettingsShape } from '../lib/settings.js';
 import { showToast } from '../ui/toast.js';
 import { recordActivity } from '../lib/activity-log.js';
 import { presentView, type ViewHandle } from '../ui/present-view.js';
+import { clearSfError, setSfError } from '../ui/panels.js';
 import { parseApexLog } from '../lib/apex-log/index.js';
 import type { ParsedLog } from '../lib/apex-log/types.js';
 import { presentApexLogAnalyzer } from '../ui/apex-log-analyzer.js';
@@ -498,6 +499,9 @@ export function createApexAnonymousFeature(options: ApexAnonymousOptions = {}): 
       limits.render(null);
       resultPane.style.display = 'none';
       resultPane.className = 'sfdt-console';
+      // Reused pane: drop the role="alert" a previous failure left on it, or
+      // this run's output is announced as an error.
+      clearSfError(resultPane);
       setStatus('', 'Running');
 
       // Trace-flag setup is best-effort: failing to arm log capture must never
@@ -618,9 +622,8 @@ export function createApexAnonymousFeature(options: ApexAnonymousOptions = {}): 
         }
       } catch (err) {
         setStatus('error', 'Failed');
-        resultPane.textContent = err instanceof Error ? err.message : String(err);
+        setSfError(resultPane, err, { doc });
         resultPane.style.display = 'block';
-        resultPane.className = 'sfdt-console sfdt-error';
       } finally {
         runBtn.disabled = false;
         analyzeBtn.disabled = false;
--- a/extension/features/apex-test-runner.ts
+++ b/extension/features/apex-test-runner.ts
@@ -11,6 +11,7 @@ import {
 } from '../lib/salesforce-api.js';
 import { showToast } from '../ui/toast.js';
 import { presentView, type ViewHandle } from '../ui/present-view.js';
+import { renderSfError } from '../ui/panels.js';
 import { button, toolbar } from '../lib/ui-controls.js';
 
 // ---------------------------------------------------------------------------
@@ -106,10 +107,7 @@ export function createApexTestRunnerFeature(options: ApexTestRunnerOptions = {})
   }
 
   function renderError(results: HTMLElement, status: HTMLSpanElement, message: string): void {
-    const panel = doc.createElement('div');
-    panel.classList.add('sfdt-console', 'sfdt-error');
-    panel.textContent = message;
-    results.appendChild(panel);
+    results.appendChild(renderSfError(message, { doc }));
     status.textContent = 'Failed';
   }
 
--- a/extension/features/bridge-tools.ts
+++ b/extension/features/bridge-tools.ts
@@ -23,6 +23,7 @@ import { loadSettings } from '../lib/settings.js';
 import { showToast } from '../ui/toast.js';
 import { recordActivity } from '../lib/activity-log.js';
 import { presentView, type ViewHandle } from '../ui/present-view.js';
+import { renderSfError } from '../ui/panels.js';
 import type { SfdtRequest, SfdtResponse } from '@sfdt/flow-core/bridge-contract';
 import { button, toolbar } from '../lib/ui-controls.js';
 
@@ -106,10 +107,7 @@ function createBridgeToolFeature(spec: ToolSpec, options: BridgeToolOptions): Fe
   }
 
   function renderError(results: HTMLElement, status: HTMLSpanElement, message: string): void {
-    const panel = doc.createElement('div');
-    panel.classList.add('sfdt-console', 'sfdt-error');
-    panel.textContent = message;
-    results.appendChild(panel);
+    results.appendChild(renderSfError(message, { doc }));
     status.textContent = 'Failed';
     // The single failure sink for runOnce — every early return routes through
     // here, so one call covers a bad request, a !ok bridge reply, and a throw.
--- a/extension/features/code-coverage.ts
+++ b/extension/features/code-coverage.ts
@@ -14,7 +14,7 @@ import {
 } from '@sfdt/flow-core';
 import { button, toolbar } from '../lib/ui-controls.js';
 import { meterCard, meterGrid, type MeterTone } from '../ui/meter-card.js';
-import { emptyPanel } from '../ui/panels.js';
+import { emptyPanel, renderSfError } from '../ui/panels.js';
 
 // Per-class coverage shaping/banding now lives in @sfdt/flow-core so the Chrome
 // viewer, the GUI Coverage page, and `sfdt coverage` band identically. These
@@ -118,10 +118,7 @@ export function createCodeCoverageFeature(options: CodeCoverageOptions = {}): Fe
       }
       results.appendChild(grid);
     } catch (err) {
-      const errorPanel = doc.createElement('div');
-      errorPanel.classList.add('sfdt-console', 'sfdt-error');
-      errorPanel.textContent = err instanceof Error ? err.message : String(err);
-      results.appendChild(errorPanel);
+      results.appendChild(renderSfError(err, { doc }));
       status.textContent = 'Failed';
     }
   }
--- a/extension/features/debug-log-viewer.ts
+++ b/extension/features/debug-log-viewer.ts
@@ -12,7 +12,7 @@ import { button, field, glyph, toolbar } from '../lib/ui-controls.js';
 import { createLimitTiles, pickLimitSnapshot } from '../ui/apex-limit-tiles.js';
 import { renderApexLogBody } from '../ui/apex-log-console.js';
 import { confirmDialog } from '../ui/confirm-dialog.js';
-import { errorPanel } from '../ui/panels.js';
+import { clearSfError, renderSfError, setSfError } from '../ui/panels.js';
 
 const DEBUG_LOG_SETTINGS_SCHEMA = z.object({
   pageSize: z.number().int().min(1).max(200).default(50),
@@ -240,7 +240,7 @@ export function createDebugLogViewerFeature(options: DebugLogViewerOptions = {})
     // Org-side failure (session, permissions, a bad query). A div rather than a
     // <pre> on purpose: the log pane below is the only <pre> in this view, and
     // tests — like a user's eye — reach for it by that.
-    const errPanel = errorPanel('', doc);
+    const errPanel = renderSfError(null, { doc });
     errPanel.style.display = 'none';
     main.appendChild(errPanel);
 
@@ -316,6 +316,9 @@ export function createDebugLogViewerFeature(options: DebugLogViewerOptions = {})
       tr.setAttribute('aria-current', 'true');
       logPane.style.display = 'block';
       logPane.className = 'sfdt-console';
+      // The pane is reused, so a previous failure's role="alert" would still be
+      // on it — announcing the next log body as an error.
+      clearSfError(logPane);
       logPane.textContent = 'Loading log…';
       limits.render(null);
       try {
@@ -326,8 +329,7 @@ export function createDebugLogViewerFeature(options: DebugLogViewerOptions = {})
         // costing a second fetch behind a second button.
         limits.render(pickLimitSnapshot(parseApexLog(raw).limits));
       } catch (err) {
-        logPane.className = 'sfdt-console sfdt-error';
-        logPane.textContent = err instanceof Error ? err.message : String(err);
+        setSfError(logPane, err, { doc });
       }
     }
 
@@ -447,6 +449,7 @@ export function createDebugLogViewerFeature(options: DebugLogViewerOptions = {})
       try {
         const result = await api.toolingQuery<ApexLogRow>(buildApexLogQuery(config.pageSize));
         loaded = result.records;
+        clearSfError(errPanel);
         errPanel.style.display = 'none';
         renderRows();
       } catch (err) {
@@ -454,7 +457,7 @@ export function createDebugLogViewerFeature(options: DebugLogViewerOptions = {})
         while (tbody.firstChild) tbody.removeChild(tbody.firstChild);
         selectedRow = null;
         status.textContent = '';
-        errPanel.textContent = err instanceof Error ? err.message : String(err);
+        setSfError(errPanel, err, { doc });
         errPanel.style.display = 'block';
       }
     }
--- a/extension/features/dependency-explorer.ts
+++ b/extension/features/dependency-explorer.ts
@@ -6,6 +6,7 @@ import {
 } from '../lib/salesforce-api.js';
 import { showToast } from '../ui/toast.js';
 import { presentView, type ViewHandle } from '../ui/present-view.js';
+import { renderSfError } from '../ui/panels.js';
 import {
   resolveQueryFor,
   referencesQuery,
@@ -144,10 +145,7 @@ export function createDependencyExplorerFeature(
       results.appendChild(renderSection('References (this → others)', refGroups));
       results.appendChild(renderSection('Referenced by (others → this)', refByGroups));
     } catch (err) {
-      const errorPanel = doc.createElement('div');
-      errorPanel.classList.add('sfdt-console', 'sfdt-error');
-      errorPanel.textContent = err instanceof Error ? err.message : String(err);
-      results.appendChild(errorPanel);
+      results.appendChild(renderSfError(err, { doc }));
       status.textContent = 'Failed';
     }
   }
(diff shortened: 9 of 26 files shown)


scoobydrew83 (commented):
## Independent review — C-FIX-4 (PR #325)

> Filed as a COMMENT review only because GitHub refuses `REQUEST_CHANGES` on a PR opened by the same
> account. **Treat this as REQUEST_CHANGES** — there are two blocking findings below.

Reviewed in my own worktree off `origin/ext/c-fix-4-render-sf-error` (102f5dd), with a baseline established
on `origin/develop` (7fa2a5e) first so I could tell a new failure from a pre-existing one. Every gate below
is output I produced, not the PR body's. I also wrote throwaway probes that drive the migrated features to
failure in happy-dom and read the resulting DOM, and I attacked the new sweep rather than trusting it.

**Verdict: REQUEST_CHANGES.** The helper itself is good and the a11y fix is real — I confirmed
`role="alert"` reaching the DOM at three surfaces and the reused-pane role being taken back off. But
acceptance criterion 3 — the sweep, which is the entire root-cause deliverable of this item — does not
hold: I reintroduced the exact PR #308 defect in a new feature file and it passed the whole suite. And
there is one live, unmigrated org-error render left in a file this PR edited.

---

## Gates — my output, against my baseline

| Gate | `origin/develop` (baseline) | PR head 102f5dd |
|---|---|---|
| `extension/ $ npx vitest run` | **108 files / 1878 tests passed** | **109 files / 1896 tests passed** |
| `extension/ $ npx tsc --noEmit` | exit 0, no output | exit 0, no output |
| `extension/ $ npx eslint .` | exit 0, no output | exit 0, no output |
| `extension/ $ npx wxt build` | ok, Σ 1.36 MB | ok, Σ 1.36 MB |
| root `$ npm test` | **312 files / 5234 tests passed** | **313 files / 5252 tests passed** |
| root `$ npm run check:all-contracts` | EXIT=0 | EXIT=0 |

`check:all-contracts` on the PR head:
```
```

The builder's setup note is correct and I verified it: `npm install` at the repo root plus
`npm run build:flow-core` are both required before the extension suite can resolve `@sfdt/flow-core`.

**No new dependencies, no new permissions — checked against the BUILT manifest, not the diff.**
`git diff origin/develop...HEAD -- '**/package.json' package-lock.json` is empty, and I diffed
`.output/chrome-mv3/manifest.json` from a `wxt build` on each side: **byte-identical**.
`generated/` and `FEATURES.json`: untouched (empty diffs).

---

## What I verified as correct

**AC1 — MET.** Driven through real features, not read off the diff. `org-limits` failing with
`"REQUEST_LIMIT_EXCEEDED\nCheck Setup › Company Information."` renders:
```
<div class="sfdt-console sfdt-error" role="alert">
  <span class="sfdt-sf-error-text">REQUEST_LIMIT_EXCEEDED</span>
  <span class="sfdt-sf-error-note">Check Setup › Company Information.</span>
```
Two `display: block` element nodes; neither can collapse into the other. Markup in an org message is
escaped (`&lt;img src=x onerror=alert(1)&gt;`, `querySelector('img')` is null).

**AC2 — MET, and I checked the classification myself rather than taking the count.** Running the pair
detector over `origin/develop` finds exactly 15 hand-rolled sites plus `ui/panels.ts:31`. All 15 are
handled: 14 to the helper, 1 (`metadata-retrieve`) to `.sfdt-callout.sfdt-bad`. Plus the four
`errorPanel()` call sites (`rest-explore`, `schema-browser` ×2, `debug-log-viewer`). The
`metadata-retrieve` call is **right**: `features/metadata-retrieve.ts:1301-1318` builds an entirely
static banner — a hardcoded `<strong>`, two literal text nodes and a docs link — with no org value
anywhere near it. `.sfdt-callout.sfdt-bad` is exactly what the comment at `lib/ui-styles.ts:598-601`
prescribes, and `role="alert"` survives the move. The `rest-explore` addition to the spec's list is
also correct.

**AC4 — MET, and the exclusions are genuinely narrow.** Two named files with per-entry reasons, and
`sf-error-panel-contract.test.ts:181-193` fails a stale one. `test/` being unscanned is justified and I
confirmed it isn't hiding anything — I found no shipped DOM-building code outside the four scanned dirs
(`mockups/` is not shipped and doesn't name the pair).

**Attack 1 — deleting `errorPanel()` is safe.** `grep -rn "\berrorPanel\b"` across the whole extension
returns only local variable names in the three explorers and comments/regex literals in tests. No
dynamic reference. It was a named ESM export, so a stale caller would be a **compile error**, not a
silent runtime `undefined` — and `tsc --noEmit` is clean. Not a risk.

**Attack 2 — the regression guard was GENERALISED, not weakened. I reconstructed the defect.** I reverted
`features/soap-explore.ts` to the pre-#308 shape (a plain `div` with the old cssText minus `white-space`,
and `errorPanel.textContent = message`) and ran `test/error-render-newlines.test.ts`:
```
    expected [ 'features/soap-explore.ts:289 (errorPanel)' ] to deeply equal []
× holds the showError() funnels the org error actually flows through
    AssertionError: features/soap-explore.ts: expected false to be true
```
Both the general sweep **and** the funnel-specific case fire. The `fills ||` short-circuit only engages
when `setSfError(errorPanel, message` is actually present; remove it and the white-space chain is enforced
exactly as before. This is not a blocking finding — the union assertion is sound.

**Attack 3 — the `loadingPanel()` EXEMPT claim is TRUE.** On `develop`, `ui/panels.ts:31`
(`el.className = 'sfdt-console sfdt-error'` inside `errorPanel`) made `setsWhiteSpaceByClass(source, 'el')`
return true for the whole file, silently vouching for `loadingPanel`'s `el.textContent = message` at :43.
I emptied the `EXEMPT` array on the PR head and the guard correctly fires:
```
    expected [ 'ui/panels.ts:128 (el)' ] to deeply equal []
```
So the exemption is real and necessary, not cover. (One narrowness nit below.)

**Attack 4 — the `textContent` semantic change is real but currently harmless.** Confirmed:
`panel.textContent` is now `"SESSION EXPIREDLog in again."` — concatenated with no separator. I searched
every `copyToClipboard(` call site (21) and every `*.textContent` read-back in `features/ ui/ entrypoints/
lib/ test/`: **nothing serialises an error panel.** All 21 copy sites pass a model object or a string, never
a rendered node, and the three explorers hide the copy button on the error path. The builder flagged this
correctly.

**A11y, driven end-to-end.** `apex-anonymous` failing → `<pre class="sfdt-console sfdt-error"
role="alert">`; on the next successful run `document.querySelectorAll('[role="alert"]')` is `[]`, so the
`clearSfError` reuse contract actually holds. Same for `soql-runner` and `org-limits`. Zero `innerHTML`
in every changed file and none anywhere in shipped extension source.

**Both themes.** `--sfdt-color-text` is defined in both `SFDT_TOKENS` (`#3e3e3c`) and `SFDT_TOKENS_DARK`
(`#e6e6e8`); `--sfdt-space-2` is theme-invariant. Contrast against `.sfdt-console`'s
`--sfdt-color-surface-alt`: light org-text 5.12:1 / note 10.26:1; dark 7.61:1 / 13.93:1. All pass AA.

**Docs site.** Verified independently — `grep -rniE "error panel|role=\"alert\"|screen reader|accessib"`
over `sfdt-site/content/` returns nothing about error-panel rendering. No MDX change needed.

---

## BLOCKING findings

### B1 — The sweep does not bite for a hand-roll written as two `classList.add` calls. `test/sf-error-panel-contract.test.ts:73`

`CLASS_APPLICATION` reads each class *application* in isolation and requires both names inside one of
them. A hand-roll that adds the classes in two statements — which is the natural shape when a pane is
created as a console and becomes an error later, i.e. exactly the reused-pane pattern this PR itself
formalises in `setSfError` — is invisible.

I did not just unit-test the matcher. I wrote a brand-new feature file, `features/zz-seventeenth.ts`,
containing the **full PR #308 defect** — no `role="alert"`, org guidance discarded — with the classes
applied in two adds:

```ts
const panel = doc.createElement('div');
panel.classList.add('sfdt-console');
panel.classList.add('sfdt-error');
panel.style.whiteSpace = 'pre-wrap';
panel.textContent = err instanceof Error ? err.message : String(err);
```

Result: `test/sf-error-panel-contract.test.ts` and `test/error-render-newlines.test.ts` — **18 passed, 0
failed**. I then did the same inside a real feature (`features/trace-flags.ts`, deliberately not the
`org-limits.ts` the builder probed with) and ran the whole toolchain: **111 files / 1912 tests passed,
`tsc --noEmit` clean, `eslint` exit 0**. The only signal at all was an unused-import *warning*, and that
only because I left the now-dead `renderSfError` import in place; a genuinely new file would not have it.

For contrast, the canonical shape *is* caught — I confirmed the sweep fires properly:
```
× no feature, entrypoint or lib module hand-rolls the error block
    expected [ 'features/trace-flags.ts' ] to deeply equal []
```

Full bypass table from driving the exported `appliesSfErrorPair` directly:

| Shape | Result |
|---|---|
| `classList.add('sfdt-console', 'sfdt-error')` | FLAGGED |
| `className = 'sfdt-console sfdt-error'` | FLAGGED |
| multi-line `classList.add(` … `'sfdt-console',` … `'sfdt-error',` … `)` | FLAGGED |
| `classList.add(...['sfdt-console','sfdt-error'])` | FLAGGED |
| **two separate `classList.add` calls** | **MISSED** |
| **`setAttribute('class', 'sfdt-console sfdt-error')`** | **MISSED** |
| **`className = 'sfdt-console'; className += ' sfdt-error'`** | **MISSED** |
| **`className = \`sfdt-console ${ERR}\``** | **MISSED** |
| **`const CLS = ['sfdt-console','sfdt-error']; classList.add(...CLS)`** | **MISSED** |
| **`classList.add('sfdt-console'); classList.toggle('sfdt-error', true)`** | **MISSED** |
| **`const A='sfdt-console', B='sfdt-error'; classList.add(A, B)`** | **MISSED** |
| **`const CLS = 'sfdt-console sfdt-error'; className = CLS`** | **MISSED** |
| **`Object.assign(p, { className: 'sfdt-console sfdt-error' })`** | **MISSED** |

I would not block on the exotic tail (`Object.assign`, `classList.toggle`, const-array spread). I am
blocking on the top three, and above all on the split-`classList.add` form, because it is the shape a
real author writing a reused pane produces on the first try — this PR's own `setSfError` is conceptually
that pattern, and both `debug-log-viewer` and `apex-anonymous` already carry panes that acquire
`sfdt-console` at construction and their error state later.

The item's stated deliverable is "a contract sweep in `extension/test/` that makes the next regression
impossible." As written it makes one *spelling* of the next regression impossible.

**Suggested fix:** accumulate class applications per receiver identifier across the file instead of
per-statement — union the classes seen for a given `<name>.` receiver and flag when that union contains
both — and add `setAttribute('class', …)` and `className +=` to `CLASS_APPLICATION`. That closes the top
three with no new false-positive surface, because the existing "two different elements" case
(`sf-error-panel-contract.test.ts:141-150`) already keys on the receiver name. Please also add the
split-add form to the `the guard actually detects a hand-roll` case at :133 so the shape is pinned.

### B2 — A live org-error render was left hand-rolled, in a file this PR migrated. `features/apex-anonymous.ts:660`

```ts
} catch (err) {
  logPane.textContent = err instanceof Error ? err.message : String(err);
}
```

This is the `openLogBtn` handler: `fetchLogBody(capturedLogId)` throws, and the org's Tooling API error is
written straight into the `<pre class="sfdt-console">` created at :435-436. It gets **no `role="alert"`**
— the exact a11y defect this item exists to remove — and, worse than the original sixteen, **no
`.sfdt-error` class either**, so the failure renders as ordinary console text and is visually
indistinguishable from a successfully fetched log body.

It sits 35 lines below `resultPane`, which this PR *did* migrate at :625. Neither guard sees it: the
contract sweep only looks for the class pair, which this site never applies; `error-render-newlines`
passes it because `logPane` wears `sfdt-console`, a white-space class. So it is a permanent blind spot
under the current rule, which is also part of why B1 matters.

This falsifies, for this feature, the claim in `extension/CHANGELOG.md` and `UI-SYSTEM.md` that "All
sixteen now come from one builder, which cannot forget the role." Please route it through
`setSfError(logPane, err, { doc })` like its sibling at :332, and either widen the sweep to catch
"error value assigned into an element that carries `sfdt-console`" or say explicitly in the sweep's
header comment that a hand-roll which omits `.sfdt-error` is out of scope.

---

## Non-blocking nits

1. **`lib/sf-error-guidance.ts:197-202` tears a multi-line ORG message apart and re-labels it as ours.**
   `splitUserFacingMessage` assumes line 1 is the org and lines 2..n are ours. When the org's own
   `message` is multi-line that is false, and the continuation lines render as `.sfdt-sf-error-note` —
   body colour, `margin-top: var(--sfdt-space-2)` between each. Observed at `apex-anonymous`:
   ```
   <span class="sfdt-sf-error-text">System.LimitException: Apex CPU time limit exceeded</span>
   <span class="sfdt-sf-error-note">Class.Foo.bar: line 12, column 1</span>   ← the org's own stack frame
   <span class="sfdt-sf-error-note">Triggered by MyTrigger</span>
   ```
   On `develop` that was one red monospace block. Blank lines are dropped too. This contradicts the PR's
   own stated invariant ("the red belongs to the org's wording"). I am not blocking because the dominant
   paths are fine — Apex compile errors and stack traces reach `resultPane` via `apex-anonymous.ts:558`,
   not the helper — but it is worth a think about whether the split should be structural (take the
   `notes` from `buildUserFacingMessage` rather than re-derive them from a string).

2. **The `white-space: pre-wrap` on the two new rules can never see a newline.** `lib/ui-styles.ts:866`
   justifies it with "an org message can be multi-line inside a single record (Apex stack traces are)" —
   but `splitUserFacingMessage` has already split every `\n` into its own span, so no span can contain
   one. `sf-error-panel-contract.test.ts:205` correspondingly asserts `must keep newlines` on behaviour
   that cannot occur. The declaration still earns its place (it preserves indentation), but the comment
   and the assertion message both describe something else.

3. **PR body: "`errorText()` also replaces the fourteen copies of `err instanceof Error ? err.message :
   String(err)`" is not accurate.** `errorText` is called in exactly one place — `ui/panels.ts:94` — and
   is not adopted by any feature. The idiom count in shipped source went 63 → 54; the 9 that went away
   did so because the panel call sites were replaced wholesale. Fine as an implementation detail; just
   don't bank the claim.

4. **`lib/sf-error-guidance.ts:213` — the `[object Object]` claim in the doc comment is wrong.** It says
   the helper is how a feature "ends up rendering `[object Object]` and nobody notices", but `errorText`
   falls through to `String(error)` and does exactly that. `renderSfError({ weird: 1 }).textContent` is
   `"[object Object]"` — and `panels.test.ts` documents the `String()` behaviour. Either `JSON.stringify`
   the plain-object case or drop the claim.

5. **`renderSfError(null)` mounts an empty `role="alert"` region.** At `soql-runner.ts:960`,
   `soap-explore.ts:208` and `rest-explore.ts:139` the hidden placeholder is created with
   `class="sfdt-console sfdt-error" role="alert"` and zero children — precisely the "empty `role="alert"`
   region left in the tree" that `clearSfError`'s own comment at `ui/panels.ts:104-105` exists to prevent.
   In practice having the live region present *before* content lands is the more reliable announcement
   pattern, so I think the behaviour is right and only the reasoning is inconsistent — but the two
   comments should agree.

6. **The `EXEMPT` entry at `error-render-newlines.test.ts:67-78` is keyed slightly wider than the case it
   names.** It exempts `ui/panels.ts` + identifier `el`, and `el` is also the parameter name of
   `setSfError`/`clearSfError`. It covers only `loadingPanel` today, but it would silently swallow a future
   `el.textContent = <error>` added inside the error builder — the same-name blind spot the entry's own
   prose warns about. Consider keying on the assigned expression (`message`) as well.

7. **`UI-SYSTEM.md:25`** — the layer-4 table lists `renderSfError()` and `setSfError()` but not
   `clearSfError()`, which the code block below it does document. Cosmetic.

8. `features/ai-assistant.ts:258` (`loading.textContent = "Error: " + …`) is another unmigrated error
   render in a touched file. Pre-existing, a view-construction failure rather than an org error, and out
   of the original sixteen — mentioning it only so it is a decision rather than an oversight.

---

## What I could NOT verify

- **Any real browser.** Everything above is happy-dom. I did not load `.output/chrome-mv3` into Chrome
  against a live org, so the *visual* outcome of the org-text/guidance split — block spacing, the note's
  colour against the console surface, the `<span>`s-inside-`<pre>` layout — is inferred from the CSS rules
  and the token values, not seen. The builder says the same; I agree it is worth a look in both themes
  before release, and finding 1 above is the specific thing to look at.
- **Actual screen-reader announcement.** I confirmed `role="alert"` is on the element that receives the
  content and that it is removed on reuse. Whether NVDA/JAWS/VoiceOver announce a live region that is
  populated while `display: none` and then shown (the three explorers' pattern) is not something happy-dom
  can answer.
- **Clipboard behaviour across the new block boundary.** I verified no code path serialises an error
  panel, but whether a *user's* manual selection-and-copy across the spans yields a newline is a browser
  behaviour I could not exercise.
- **Real org error shapes.** Finding 1's severity depends on how often a Salesforce record's own `message`
  contains newlines on these specific surfaces. I reasoned about it from the code; I have no live-org
  sample to weigh it against.

I did not push anything, did not touch `FEATURES.json`, and left no changes behind in the branch.

---

scoobydrew83 (commented):
## Re-review — round 2 (`f0ef442`)

**Verdict: REQUEST CHANGES** — one blocking finding, narrow and cheap to fix. Both of my round-1 blockers are genuinely fixed, all eight nits are done, and the nit-1 refactor is the right structural fix rather than a patch. What blocks is new surface introduced in round 2: **rule 2's identifier alternation is narrower than its own comment claims, and the hole is named after the variable all three `showError()` funnels use.** I demonstrated the defect end-to-end, in a real feature file, undetected by all three guards, with `tsc` clean.

### ⚠️ Process note, before anything else

**PR #325 is closed and merged.** It was merged at 16:17 UTC at `102f5dd` — round 1, the commit I filed REQUEST_CHANGES against. `origin/develop` (`e493254`) has a tree byte-identical to `102f5dd`, so **both round-1 blockers (B1's sweep bypass and B2's unmigrated `apex-anonymous.ts:660`) are on `develop` right now**, and `f0ef442` — the commit that fixes them — is stranded on the branch with no open PR. This review is a comment on a closed PR; round 2 needs a new PR to land. Separately, #326's description says it reuses `errorPanel()` from `ui/panels.ts`, which round 1 deleted — worth checking before that one rebases.

---

## Gates — my output, against my baseline

Baseline is `origin/develop` @ `e493254` (= round-1 content), measured in a fresh worktree with a fresh `npm install` + `npm run build:flow-core`.

| Gate | baseline `e493254` (round 1) | `f0ef442` (round 2) |
|---|---|---|
| `extension/ npx vitest run` | 109 files / **1896** pass | 109 files / **1907** pass |
| `extension/ npx tsc --noEmit` | exit 0 | exit 0 |
| `extension/ npx eslint .` | exit 0 | exit 0 |
| `extension/ npx wxt build` | ok, Σ 1.36 MB | ok, Σ 1.36 MB |
| root `npm test` | 313 / **5252** pass | 313 / **5263** pass |
| root `npm run check:all-contracts` | EXIT=0 | EXIT=0 |

Every number matches the builder's claim exactly. Built `chrome-mv3/manifest.json` diffed on both sides: **identical**. `git diff origin/develop HEAD -- generated/ FEATURES.json '**/package.json' package-lock.json` is **empty** — nothing hand-edited, no new deps, no new permissions.

---

## BLOCKING

### B3 — Rule 2 misses the funnel variable name. `test/sf-error-panel-contract.test.ts:229-235`

```ts
// Rule 2's matcher. The identifier alternation mirrors
// `error-render-newlines.test.ts` so the two guards agree about what "renders
// an error" means; …
const RENDERS_ERROR_VALUE =
  /\b([A-Za-z_$][\w$]*)\s*\.\s*textContent\s*=\s*([^;\n]*\b(?:err|error|errorMsg)\b[^;\n]*)/g;
```

It does **not** mirror it. `error-render-newlines.test.ts:38-39` is `(?:err|error|errors|errorMsg|message|msg)` **with `/i`**. Rule 2 drops `errors`, `message`, `msg`, and the case-insensitive flag.

`message` is not an arbitrary omission. It is the parameter name of `showError(message)` in `soql-runner`, `rest-explore` and `soap-explore` — the three surfaces #308 was reported against — and `error-render-newlines.test.ts:28-37` carries a nine-line comment explaining that an earlier draft left `message`/`msg` out and that doing so "meant the guard did not hold the very surface the reported bug appeared on." Rule 2 re-made that exact mistake while claiming in a comment to have copied the fix.

Demonstrated, not asserted. Injected into `features/dependency-explorer.ts` (a third file — not your `trace-flags.ts`, not my round-1 one), Prettier-faithful single quotes:

```ts
const logPane = doc.createElement('pre');
logPane.className = 'sfdt-console';
const message = err instanceof Error ? err.message : String(err);
logPane.textContent = message;
results.appendChild(logPane);
```

That is B2 exactly — a live org error in a `.sfdt-console` with no error styling and no `role="alert"`. Result:

```
tsc=0    Tests  24 passed (24)     *** NOT CAUGHT BY ANY GUARD ***
```

Rule 1 can't see it (no `.sfdt-error`). Rule 2 can't see it (`message` not in the alternation). `error-render-newlines` can't see it (`logPane` wears `.sfdt-console`, which satisfies its white-space check). **Fix:** use the same alternation and the `/i` flag as the sibling guard, and add a case for the `message` spelling to the `rule 2 detects…` block at :382.

---

## Non-blocking, but fix alongside B3 (one line)

### N1 — Rule 1: `let`/`var` class accumulators are invisible. `test/sf-error-panel-contract.test.ts:158-165`

`constClassTokens` matches `/\bconst\s+…/` only, so:

```ts
let cls = 'sfdt-console';
if (err) cls += ' sfdt-error';
errPanel.className = cls;
```

records nothing (`classesByElement` → `[]`, confirmed directly). In `features/dependency-explorer.ts` this is *partially* backstopped by `error-render-newlines` — but only incidentally, because `setsWhiteSpaceByClass` needs a quoted literal and `className = cls` isn't one. Add one line and the backstop is gone:

```
E: let-accumulator + errPanel.style.whiteSpace = 'pre-wrap'
   tsc=0   Tests 24 passed (24)   *** NOT CAUGHT BY ANY GUARD ***
```

`const` → `(?:const|let|var)` closes it. I rate this a nit rather than a blocker because this repo is const-heavy and the form is less natural than B3's, but it is the same class of hole and the same edit session.

### N2 — `lib/ui-styles.ts`'s exclusion is not load-bearing. `test/sf-error-panel-contract.test.ts:76-81`

You added an excellent test at :454 proving `ui/panels.ts` would trip rule 1 — that exclusion earns its place. The same probe on the stylesheet:

```
ui/panels.ts    : rule1=true   rule2=[]
lib/ui-styles.ts: rule1=false  rule2=[]
```

It trips neither rule, so excluding it removes a file from the scan for no benefit — a GP#12 hole rather than a GP#12 exclusion. The stale-exclusion test at :440 doesn't catch this: it only asserts the file still *names* the classes, which a stylesheet does by definition. Either drop the entry or give it the same "would trip the rule" proof `ui/panels.ts` gets. (My round-1 review missed this; it has been decorative since then.)

### N3 — Nothing guards the `unknown` trap. Audit clean, guard absent.

I audited **every** call site of all six widened signatures. Result: **no live violation.**

| Signature | error-object calls | string calls, and what they are |
|---|---|---|
| `soql-runner.ts:1139 showError` | :1506, :1602, :1634 ✅ | :1456, :1589 — our literal prose ("Enter a SOQL query to run.") |
| `rest-explore.ts:219 showError` | :290 ✅ | :235 literal; :263 `Body is not valid JSON: …` — a local `JSON.parse` `SyntaxError`, never `.userFacing` |
| `soap-explore.ts:287 showError` | :345 ✅ | :304 literal; :313 `Payload is not valid JSON: …` — same |
| `apex-test-runner.ts:111 renderError` | :213 ✅ | :177, :185 — our prose about a malformed org response |
| `bridge-tools.ts:110 renderError` | :135, :149 ✅ | :142 — `response.error`, a string field on an `SfdtResponse`, not a throw |
| `ai-assistant.ts renderResultError` | :251 ✅ | :248 — same bridge string field |

Every string passed is either our own literal or a value that provably never carries `.userFacing`. But there is **no rule anywhere** pinning it — I grepped both guards; every `err.message` occurrence in them is their own fixture. And the failure is silent, which I measured:

```
passing the ERROR   : 1 org-text node + 3 note nodes
passing err.message : 1 org-text node + 0 note nodes   ← whole composition flattened into one blob
```

`tsc` accepts both. A rule of the form "no argument to `renderSfError`/`setSfError`/`showError`/`renderError` may match `\.message\b` or `instanceof Error \?`" is writable; it would flag `rest-explore.ts:263` and `soap-explore.ts:313`, which need either an exemption or a rewrite to `showError(err, { guidance: … })`. Not blocking — the audit is clean — but this is precisely the "invariant enforced by nothing" this item exists to remove, and it is now the item's own weakest link.

### N4 — `ai-assistant.ts:264` relies on the new `renderSfError(null)` semantics

`renderSfError` now returns a classless, role-less, empty `<div>` for `null`/`undefined` (correct, and it fixes my round-1 nit 5). But `loading.replaceWith(panel)` would then silently swap the "Fetching…" line for an empty div if something threw `undefined`. Vanishingly unlikely; noting it only because the null contract changed under a brand-new call site.

---

## What I verified as correct

**B1 is fixed, and I re-tested every form the builder claims closed.** All ten flagged, driving the exported `buildsSfErrorPanel` directly: split-across-statements, `setAttribute('class', …)`, `className +=`, template interpolation, const-array + spread, const identifiers, `classList.toggle`, `classList.replace`, `Object.assign`, wrapped multi-line assignment. `this.pane.…`, `state.pane.…`, a const object field, and two adds 200 lines apart are also caught. The three documented-open forms (non-identifier access, aliasing, runtime-computed names) behave exactly as documented — no surprises there. The wrapped-assignment find the builder reports is real and was not on my list.

**Third-file demonstration, mine not yours.** Split-add hand-roll injected into `features/dependency-explorer.ts`, `tsc --noEmit` exit 0 so it genuinely compiles:
```
× no feature, entrypoint or lib module hand-rolls the error block  → ['features/dependency-explorer.ts']
× no console pane is handed a caught error directly                → ['features/dependency-explorer.ts (errPanel)']
Tests  2 failed | 12 passed (14)
```
Both rules fire. Reverted.

**B2 is fixed** — `apex-anonymous.ts:661` now `setSfError(logPane, err, { doc })` with `clearSfError(logPane)` at :651 on both success branches. The extra `ai-assistant.ts:258` find is real and was mine to have caught as more than a nit.

**Nit 1 is fixed structurally, and I verified the whole journey through the real client.** Driving `SalesforceApiClient` with a mocked bus and a genuine two-record org body whose first record's `message` is multi-line:
```
.userFacing.orgText = "Salesforce GET request failed (HTTP 400): No such column 'Foo__c' on entity 'Account'.
                       If you are attempting to use a custom field, be sure to append __c."
.userFacing.notes   = [INVALID_FIELD · field: Foo__c — …, "Also: unexpected token: SELCT", MALFORMED_QUERY — …]

<span class="sfdt-sf-error-text">…both org lines, one node, error colour…</span>
<span class="sfdt-sf-error-note">×3</span>          role="alert"
```
The round-1 misattribution is gone: the org's continuation line stays in the org's node and the org's colour. `buildUserFacingMessage(…) === [orgText, ...notes].join('\n')` still holds, so nothing downstream of the legacy string moved.

**The structured path does not cross the worker boundary — verified by reading and by test.** `buildRequestError` is called at `lib/salesforce-api.ts:457` and `:515`, i.e. **after** `await sendProxied(…)` resolves. What crosses `chrome.runtime.sendMessage` is a plain `SfApiFetchResponse` (`{ok, status, bodyText, errors}`), which I confirmed round-trips `structuredClone` losslessly. The `SalesforceRestError` is built in the renderer's own realm. I also measured the counterfactual: `structuredClone(new SalesforceRestError(…))` yields `{isError: true}` with **`userFacing` dropped**, so if an error object ever were postMessage'd, `sfErrorParts` would silently fall back to the one-node path. Latent, not live — worth a comment on `CarriesUserFacingParts` rather than code.

**Nit 2's replacement is not vacuous.** I mutation-tested it: deleting `white-space: pre-wrap` from `.sfdt-sf-error-text` in `lib/ui-styles.ts` produces `× the org text keeps the newlines it arrived with`. And it is now genuinely load-bearing, because the structured path means `orgText` really can contain newlines in one node — which is exactly what the round-1 version could never exercise.

**Round-1 clearances all still hold after the refactor.** Pre-#308 defect reconstructed at `features/soap-explore.ts` (classless div, no white-space rule, `showError(message: string)` + `errorPanel.textContent = message`): both `every element assigned an error message declares a white-space rule` and `holds the showError() funnels…` fail. No dangling `errorPanel()` callers anywhere. `metadata-retrieve.ts:1302-1303` still `.sfdt-callout .sfdt-bad` with `role="alert"`.

**The `expression` field on `EXEMPT` (round-1 nit 6) is the right fix** — the entry can no longer swallow a future `el.textContent = err.message` in the same function.

**Your six-false-positives argument is right, and understated.** I implemented the wide rule (drop the `.sfdt-console` requirement) and ran it over the tree: it flags **11** sites, not six — `apex-test-runner.ts:145` (a per-failure list item), `event-monitor.ts:250`, `metadata-retrieve.ts:229`, `schema-browser.ts:307/521/577` (describe-cache cell annotations), `ui/confirm-dialog.ts:65`, `ui/health-modal.ts:174`, `ui/toast.ts:80`, `options/main.ts:323/677`. I judge all eleven correct code. I looked for a narrower formulation that catches a classless `<div>` used as a failure region without flagging a status line, and there isn't one — the codebase encodes element role only in the class, and `loading` had no class. **Accepting this as a documented limit is the right call.** Fixing B3 is the cheap win in the same area, and it is orthogonal.

---

## What I could not verify

- **No real browser.** Everything is happy-dom. The visual result of the structured split — the org's multi-line text wrapping in one `pre-wrap` node, note spacing, the `<span>`-inside-`<pre>` layout in both themes — is inferred from CSS rules and token values, not seen.
- **Screen-reader announcement.** I confirmed `role="alert"` lands on the element receiving the content and is removed on reuse and on empty. Whether a real AT announces a live region populated while `display:none` and then shown is not answerable here.
- **Live-org error shapes.** My multi-line-org-message fixture is synthetic, built to the documented REST error contract. I have no real org response to confirm how often the first record's `message` actually contains newlines.
- **`f0ef442` under CI.** It has no open PR, so no checks have run on it.

I pushed nothing, touched no `FEATURES.json`, and left the branch clean.

---

PR #327, opened 2026-08-05
fix(extension): the C-FIX-4 sweep does not actually bite — follow-up to #325 (B1/B2)

scoobydrew83:
**Follow-up to #325, not a new feature.** #325 was merged at `102f5dd` while a review round was still open, so two blocking findings from that review shipped. This PR carries the fixes, plus three further rounds of review on the guard those fixes added.

> **Round 4 update.** Rounds 2 and 3 each closed a hole in the sweep and each opened another, both found by *constructing the #308 defect and watching the suite stay green* — never by reading the regex. Round 3's was a regression: a `LITERAL_ONLY` suppressor that measurably suppressed **nothing** while hiding an interpolated error and a live offender. Round 4 deletes it and closes the root cause. The gate numbers below are for the current head, `1f802d2`. The description of rule 2 is round 4's; where it differs from what rounds 2 and 3 claimed, round 4 is right and the earlier claim is corrected in place.

## What was broken on `develop`

Both verified by reading the merged tree, not inferred.

**B1 — the contract sweep did not catch a hand-rolled error panel.** `extension/test/sf-error-panel-contract.test.ts:73` on `develop` read each class application **in isolation**, so a hand-roll that applies the two classes in two statements was invisible:

```ts
pane.classList.add('sfdt-console');
pane.classList.add('sfdt-error');       // ← sweep saw nothing
```

That is the shape an author writing a reused pane produces *first*, not an exotic bypass. `setAttribute('class', …)`, `className +=`, template interpolation, const-array spread, `classList.toggle`, const identifiers and `Object.assign` were all equally invisible. The sweep is the entire deliverable of C-FIX-4, and **#325's own description overstated it.**

**B2 — a seventeenth site, still carrying the defect.** `extension/features/apex-anonymous.ts:660` rendered a live Tooling failure into a plain `.sfdt-console` pane: no `.sfdt-error`, no `role="alert"`. Neither the new sweep nor either #308 guard could see it — the pane's `.sfdt-console` already satisfied the white-space rule they check.

## B1 — the fix

Rule 1 no longer asks a question of each class application. A string- and nesting-aware `readExpression()` plus a `const`-name map feed **per-element class accumulation**: every class each element identifier receives anywhere in the file, asked once per element.

**Closed**, each with its own case in `survives every class-application form the codebase uses`: split across statements · `setAttribute('class', …)` · `className +=` · template interpolation · const array + spread · const identifiers · `classList.toggle`/`.replace` · `Object.assign` · wrapped multi-line assignment · `let`/`var` accumulators.

The wrapped assignment was not on the reviewer's list; it turned up while fixing the others. Reading originally stopped at the first depth-0 newline, so `debug-log-viewer.ts`'s existing two-line `statusPill.className =` recorded **no classes at all** — a live blind spot, harmless only because those happen to be pill variants.

**Consciously left open**, with the reason in a comment on `classesByElement`: non-identifier element access (`panels[i].className`), aliasing, and runtime-computed class names (`'sfdt-' + kind`). Each is a deliberate evasion rather than ordinary authoring.

## B2 — the fix, and the rule that finds its shape mechanically

`apex-anonymous.ts:660` now calls `setSfError(logPane, err, { doc })`, with `clearSfError(logPane)` on entry so a retry that succeeds is not still announced as the previous failure.

Rule 1 structurally cannot see a hand-roll that never applies the classes, so the sweep gains **rule 2**: *no `.sfdt-console` element may be handed a caught error's text directly.* Scoped to consoles, not all `textContent` assignments — a status line and a table cell legitimately show an error string inline.

**One more found by reading, not tooling:** `features/ai-assistant.ts:258` wrote a live Tooling failure as unstyled body text. Rule 2 misses it — `loading` is a bare `<div>` with no `.sfdt-console`. Fixed. A rule wide enough to catch it flags eleven correct sites, so that limit is accepted and documented rather than papered over.

## The `unknown`-signature audit, and rule 3

The nit-1 refactor widened six signatures to `unknown`, and **`unknown` accepts a string** — a call site passing `err.message` compiles clean and silently flattens the structure. I audited all 24 call sites. **Two were doing it** (`rest-explore.ts:263`, `soap-explore.ts:313`); both now pass the error plus their own line as `guidance`, which is the option that existed for this and previously had no caller at all.

A wrong call site cannot be made a compile error, so the sweep gains **rule 3**: nothing may hand a stringified error to a shared renderer or to a funnel that forwards to one. Funnels are identified as `unknown`-parameter functions in a file importing the renderers — an over-approximation on purpose, which only over-includes, and which correctly scopes **out** `flow-health-check.ts`'s `showError(message: string)`.

## Round 4 — the suppressor, and the root cause behind three rounds of holes

**Round 3 shipped a regression.** It added, to both rule 2 and `error-render-newlines.test.ts`:

```ts
const LITERAL_ONLY = /=\s*['"`][^'"`]*['"`]\s*;?\s*$/;
```

The character class excludes quote characters but not `$` or `{`, so **a template literal that interpolates an error was classified as fixed copy**. In a `.sfdt-console` pane with no `.sfdt-error` and no `role="alert"`:

```ts
preview.textContent = `Could not save: ${err instanceof Error ? err.message : String(err)}`;
```

Round 2's rule 2 caught that; round 3 did not. Worse, on rule 2 the suppressor was measured against the whole tree with `if (LITERAL_ONLY.test(whole!)) continue;` short-circuited to `false &&`, and the suite stayed at **111 files / 2000 tests**. It suppressed **zero** false positives — including `pane.textContent = 'Loading log…'`, the case written to justify it, whose string contains no word from rule 2's alternation. An exclusion that excludes nothing is a hole, not an exclusion, which is the sentence this PR itself added one round earlier about `lib/ui-styles.ts`.

**Deleted from rule 2.** In `error-render-newlines.test.ts` it *was* earning its keep on two static sites, so there it is **replaced** by `dynamicParts()`, which blanks the text inside string literals in place — a template keeps its `${…}` holes and loses its prose. Static copy is excluded because nothing is left to look at, not because a pattern guessed where the literal ended. It is length- and newline-preserving, which also stops a scan matching `query = """` inside the Python code template embedded in `soql-runner.ts`.

### Two live offenders it was hiding

- **`features/event-monitor.ts:250`** — the Platform-Event limits pane rendered a caught failure as bare text into a hand-styled `div`: no class, no `role="alert"`, no `white-space` rule. Now `setSfError(limitsContainer, err, { doc })` with the error **object**, plus `clearSfError()` on re-entry. A behavioural test pins it and fails without the source change.
- **`features/metadata-retrieve.ts:229`** — called a false positive on review, and is not. An `addLog('error', …)` whose text interpolates `err.message` is written at ten sites, so that log line really does carry a multi-line org message. It now wears `.sfdt-msg`.

The other two sites the swap surfaces (`field-creator.ts:398` `'Error'`, `data-import.ts:745`) are genuine false positives and are declined **structurally** by `dynamicParts()`. **No exemption entry was added.** The exemption list did gain the check it was missing — `every exemption would actually be flagged without it`, the counterpart to the sweep's `every exclusion would actually fail without it`, and the check whose absence let `LITERAL_ONLY` in.

### The root cause, closed rather than patched again

Rule 2 and rule 3 carried two different name allowlists **in one file** that disagreed — rule 3 accepted `ex`/`caught`/`reason`, rule 2 did not — and `error-render-newlines.test.ts` carried a third. Every round of review found another spelling one accepted and another did not. All three now consult one module, `test/error-source-scan.ts`, and the answer is **mostly structural**:

- every `catch (X)` binding — X *is* the thrown error, whatever it is called, which closes `ex`, `caught`, `reason`, `problem`, `thrown` and a bare `e` in one stroke;
- every `const`/`let`/`var X = <expression that flattens one>`, plus `+=` accumulators and aliases, to a fixed point;
- spelling survives **only** as the fallback for a function *parameter*, bound at the call site, which is the `showError(message)` case;
- and that fallback is word-based rather than `\b`-delimited, so `errMsg`, `errMessage`, `sfError` and `errorText` are covered.

Each flattening idiom now carries its own **subject** test — `.message` requires the object to be error-ish, `String(x)` its argument, `${x}` a bare error-ish identifier — so `${response.error}`, `${root}` in `schema-browser.ts` and `summary.message` in `apex-anonymous.ts` are correctly declined. `reason`, `problem`, `thrown` and `failure` came **off** the spelling list: `ui/apex-log-analyzer.ts` names a log *truncation* reason `reason`. They are reached by the catch scan instead.

Rule 2 also reads its right-hand side with `readExpression()` instead of `[^;\n]*`, closing the two shapes Prettier produces from a long assignment.

**Still open, with its own test saying so:** the two-hop in-file funnel through a function parameter. It needs dataflow a regex guard cannot do, and it is now a documented decision rather than a surprise.

## Demonstrated failure, by construction

24 injections of the #308 defect into `features/debug-log-viewer.ts` and `features/flow-quality.ts` — two files no round has used, neither touched by this diff. Each written into the file, `prettier --write`-normalised, `tsc --noEmit` exit 0, `eslint` exit 0, both guards run, then reverted.

```
unfixed head 4ae470d:   3 / 24 caught   (the three controls, and nothing else)
head         1f802d2:  23 / 24 caught   (all but the documented two-hop funnel)
```

Every one of the twenty new spellings is confirmed to fail against the unfixed code. The full CAUGHT/NOT-CAUGHT table is in the round-4 review reply.

## Also carried from round 2 (unchanged)

**Nit 1 — multi-line org messages were being re-labelled as our notes.** The org's own message can be multi-line inside one record (an Apex compile error, a stack trace), so "line one is theirs, the rest is ours" is wrong exactly when it matters. The guess is removed rather than improved: the parts now travel **as structure** — `buildUserFacingParts()` composes them, `SalesforceRestError` carries them on `.userFacing`, `sfErrorParts()` reads them back, and an error with no structure renders as **one** `pre-wrap` node. `splitUserFacingMessage` is deleted.

Plus: the vacuous white-space assertion now tests newlines *inside* the org's message (2); `errorText()`'s reach claim corrected (3); the false `[object Object]` claim dropped (4); `renderSfError(null)` no longer mounts an empty `role="alert"` (5); the `EXEMPT` entry narrowed to the exact expression (6); `clearSfError` added to the `UI-SYSTEM.md` table (7); `ai-assistant.ts:258` fixed (8).

`CHANGELOG.md` and `UI-SYSTEM.md` no longer claim sixteen surfaces come from one builder. The changelog also no longer claims the **stylesheet** is an excluded defining artifact — round 3 removed `lib/ui-styles.ts` from `DEFINING_ARTIFACTS` (it declares the classes in CSS, which no rule reads, so it tripped nothing) and the changelog had not caught up.

## Gates — commit `1f802d2`

Baseline re-measured on `origin/develop` @ `cbffd36` in the same worktree and install.

| | `origin/develop` `cbffd36` | head `1f802d2` |
|---|---|---|
| `extension/ npx vitest run` | 111 files / **1983** passed | 111 files / **2009** passed |
| `extension/ npx tsc --noEmit` | exit 0 | exit 0 |
| `extension/ npx eslint .` | exit 0 | exit 0 |
| `extension/ npx wxt build` | Σ 1.38 MB | Σ 1.38 MB |
| root `npm test` | 315 files / **5339** passed | 315 files / **5365** passed |
| root `npm run check:all-contracts` | exit 0 | exit 0 |

Built `chrome-mv3/manifest.json` **byte-identical** both sides (`sha256 2a302f6a1b97539b0857f162e24fb498774b602f8840b515d865ecc9e8359c7d`).

## Diff contents

Branched clean off `origin/develop` and cherry-picked the round-2 work, so the diff is only the follow-up — no already-merged content. No new runtime dependencies, no new manifest permissions, `generated/` regenerated with no drift, `FEATURES.json` untouched. Still an internal refactor plus a11y fixes, so no docs-site MDX change.

## Still not verified

- **No real browser run.** happy-dom only, on all four rounds. The nit-1 change alters the common case, so both-themes remains CSS-rule assertions rather than eyes on a screen.
- **The limits pane's appearance.** `limitsContainer` carries a hand-written `cssText` with a warning-tinted background and now gains `.sfdt-console` on top of it. The cascade was reasoned about, not looked at.
- **Timeout and no-session errors** from `lib/salesforce-api.ts` are plain `Error`s and render as one node. Correct — nothing was appended to them — but not driven through a panel in Chrome.
- **Rule 3's funnel detection is an approximation.** A funnel that takes a concrete type rather than `unknown` and still forwards to the renderer would not be watched. None exists today.
- **The injections nobody thought of.** 24 constructed spellings across two files is more than any previous round and still not the space. The structural claim worth testing is that the catch scan and the binding scan close *categories*, so the next hole should be a new category rather than a new spelling.

---


scoobydrew83 (commented):
**REQUEST CHANGES** — one blocking finding, a one-word fix. Everything else in this PR is right, and it fixes both of my #325 blockers plus all eight nits. What blocks is a hole in **rule 2**, added in this PR, whose alternation is narrower than its own comment claims — and the missing word is the variable name all three `showError()` funnels use.

> Filed as a COMMENT review only because GitHub refuses `REQUEST_CHANGES` on a PR opened by the same account. The verdict above is the verdict.

Reviewed at `edb5829` in my own worktree. I re-ran every gate myself and re-tested every claim rather than carrying forward my earlier analysis, because this head is not the `f0ef442` I looked at before: it adds rule 3 and changes two call sites.

**Diff verified as follow-up-only.** Branched off `e493254`, 2 commits, 20 files, `git diff --stat origin/develop...HEAD` contains no already-merged content (`org-limits.ts`, `code-coverage.ts`, `trace-flags.ts`, `flow-quality.ts` — all untouched here, as they should be).

---

## Gates — my output, against the current `develop` tip

Baseline re-measured on `origin/develop` @ `e493254` in this same worktree and install, so both sides are apples to apples.

| Gate | baseline `e493254` | `edb5829` |
|---|---|---|
| `extension/ npx vitest run` | 109 files / **1896** pass | 109 files / **1910** pass |
| `extension/ npx tsc --noEmit` | exit 0 | exit 0 |
| `extension/ npx eslint .` | exit 0 | exit 0 |
| `extension/ npx wxt build` | ok, Σ 1.36 MB | ok, Σ 1.36 MB |
| root `npm test` | 313 / **5252** pass | 313 / **5266** pass |
| root `npm run check:all-contracts` | EXIT=0 | EXIT=0 |

Every number matches the PR description exactly. Built `chrome-mv3/manifest.json` diffed on both sides: **identical**. `generated/`, `FEATURES.json`, all `package.json`, `package-lock.json`: **empty diff**.

---

## BLOCKING

### B1 — Rule 2's alternation omits the funnel variable name. `test/sf-error-panel-contract.test.ts:239-245`

```ts
// Rule 2's matcher. The identifier alternation mirrors
// `error-render-newlines.test.ts` so the two guards agree about what "renders
// an error" means; …
const RENDERS_ERROR_VALUE =
  /\b([A-Za-z_$][\w$]*)\s*\.\s*textContent\s*=\s*([^;\n]*\b(?:err|error|errorMsg)\b[^;\n]*)/g;
```

It does not mirror it. `error-render-newlines.test.ts:38-39` is `(?:err|error|errors|errorMsg|message|msg)` **with the `/i` flag**. Rule 2 drops `errors`, `message`, `msg` and the flag.

`message` is not an arbitrary omission — it is the parameter name of `showError(message)` in `soql-runner.ts:1139`, `rest-explore.ts:221` and `soap-explore.ts:289`, the three surfaces #308 was reported against. And `error-render-newlines.test.ts:28-37` carries a nine-line comment recording that an earlier draft left `message`/`msg` out and that doing so "meant the guard did not hold the very surface the reported bug appeared on." Rule 2 reproduces that exact omission while citing that file as its source.

Demonstrated, in `features/code-coverage.ts` — a third file neither of us has used (you took `org-limits.ts` and `trace-flags.ts`; I took `trace-flags.ts` and `dependency-explorer.ts`). Prettier-faithful single quotes:

```ts
const logPane = doc.createElement('pre');
logPane.className = 'sfdt-console';
const message = err instanceof Error ? err.message : String(err);
logPane.textContent = message;
results.appendChild(logPane);
```

That is B2 from #325 verbatim — a live org error in a `.sfdt-console` with no `.sfdt-error` and no `role="alert"`:

```
C  B3: console pane, value named `message`
      tsc=0   Tests  27 passed (27)      *** NOT CAUGHT BY ANY GUARD ***
```

Rule 1 can't see it (never applies `.sfdt-error`). Rule 2 can't see it (`message` not in the alternation). `error-render-newlines` can't see it (`logPane` wears `.sfdt-console`, which satisfies its white-space check). Rule 3 can't see it (no renderer call).

**Fix:** `(?:err|error|errors|errorMsg|message|msg)` plus `/gi` — and add the `message` spelling as a case in `rule 2 detects an error rendered into a bare console pane` at :392 so the shape stays pinned. Please check the `ERROR_COUNT` strip at :246 and the re-test at :260 together, since they carry the same short alternation.

---

## Non-blocking

### N1 — Rule 1: `let`/`var` class accumulators are invisible. `test/sf-error-panel-contract.test.ts:170`

`constClassTokens` matches `/\bconst\s+…/` only:

```ts
let cls = 'sfdt-console';
if (err) cls += ' sfdt-error';
errPanel.className = cls;
errPanel.style.whiteSpace = 'pre-wrap';
errPanel.textContent = err instanceof Error ? err.message : String(err);
```
```
D  N1: let-accumulator + explicit whiteSpace
      tsc=0   Tests  27 passed (27)      *** NOT CAUGHT BY ANY GUARD ***
```

Without the explicit `whiteSpace` line, `error-render-newlines` catches it incidentally — but only because `setsWhiteSpaceByClass` needs a quoted literal and `className = cls` isn't one, so the backstop is one line away from gone. `const` → `(?:const|let|var)` closes it. Nit rather than blocker because this repo is const-heavy, but it is the same class of hole and the same edit.

### N2 — Rule 3's holes: an intermediate `const` defeats it entirely. `test/sf-error-panel-contract.test.ts:307`

`STRINGIFIED_ERROR` is `/\binstanceof\s+Error\s*\?|\b(?:err|error|e)\s*\.\s*message\b/`. Driving the exported `passesStringifiedError`:

| Form | |
|---|---|
| `renderSfError(err instanceof Error ? err.message : String(err))` | FLAGGED |
| `setSfError(pane, err.message)` | FLAGGED |
| `showError(err.message)` via an `unknown` funnel | FLAGGED |
| **`const msg = err instanceof Error ? … ; renderSfError(msg)`** | **MISSED** |
| **`renderSfError(String(err))`** | **MISSED** |
| ``renderSfError(`${err}`)`` | MISSED |
| `renderSfError(err.toString())` | MISSED |
| `renderSfError((err as Error).message)` | MISSED |
| `renderSfError(err?.message)` | MISSED |
| `renderSfError(caught.message)` / `ex.message` / `reason.message` | MISSED |

The first two also confirmed end-to-end in `features/code-coverage.ts` — both `tsc=0`, both `27 passed (27)`, neither caught. The intermediate-`const` one is the notable one: it is the ordinary refactor when the expression is used twice, and it is the shape rule 3's own fixture at :396 uses *inline*.

Not blocking. Rule 3 answers a nit I raised where the audit was already clean, it catches the three canonical shapes, and the tree has zero violations. But its comment describes it as closing the `unknown` hole, and `const msg = …` reopens it, so the comment should say what it actually covers. Adding `String\s*\(|\.toString\s*\(\)` and widening the identifier set to `\w+\s*\.\s*message` would close most of it cheaply; the intermediate const needs const-resolution like rule 1 already has.

### N3 — `lib/ui-styles.ts`'s exclusion is load-bearing for none of the three rules. `test/sf-error-panel-contract.test.ts:76-81`

Your `the helper itself would trip rule 1 — which is why it is excluded` test at :471 is exactly the right idea. Applying the same probe per rule:

```
ui/panels.ts       rule1=true   rule2=[]   rule3=[]
lib/ui-styles.ts   rule1=false  rule2=[]   rule3=[]
```

The stylesheet trips nothing, so excluding it removes a file from all three scans for no benefit — a GP#12 hole rather than a GP#12 exclusion. The stale-exclusion test at :453 doesn't catch it: it only asserts the file still *names* the classes, which a stylesheet does by definition. Drop the entry, or give it the same "would trip a rule" proof. (I missed this in round 1; it has been decorative since #325.)

### N4 — the `structuredClone` counterfactual deserves a comment on `CarriesUserFacingParts`

Measured: `structuredClone(new SalesforceRestError(…)).userFacing` is `undefined`. Not live today (see below), but if an error object were ever routed through `postMessage`, `sfErrorParts` would silently fall back to the one-node path and the guidance would vanish with no failing test. One sentence on the interface would stop that being rediscovered.

### N5 — `ai-assistant.ts:264` leans on the new `renderSfError(null)` contract

`renderSfError` now correctly returns a classless, role-less empty `<div>` for nullish input. `loading.replaceWith(panel)` would therefore silently swap "Fetching…" for an empty div if something threw `undefined`. Vanishingly unlikely; flagging only because the null contract changed under a new call site.

---

## Verified — the substance of this PR

**The sweep bites, in a third file.** Canonical split-add hand-roll injected into `features/code-coverage.ts`, `tsc --noEmit` exit 0 so it genuinely compiles:
```
A  canonical split-add hand-roll
      tsc=0   Tests  2 failed | 25 passed (27)
      CAUGHT BY: × no feature, entrypoint or lib module hand-rolls the error block
      CAUGHT BY: × no console pane is handed a caught error directly
```
Both rules fire. Reverted. **Rule 3 also bites** on its target (`renderSfError(err instanceof Error ? err.message : String(err))` → `× nothing hands the renderer an already-stringified error`).

**All ten claimed-closed rule-1 forms are genuinely closed** — I re-drove `buildsSfErrorPanel` at this head: split statements, `setAttribute('class', …)`, `className +=`, template interpolation, const array + spread, const identifiers, `classList.toggle`, `classList.replace`, `Object.assign`, wrapped multi-line assignment. **The three documented-open forms behave exactly as documented** — alias, `panels[i].className`, `'sfdt-' + kind` all missed, all named in the comment at :186-204. No surprises, and the wrapped-assignment find you report is real and was not on my list.

**The `unknown` audit: zero violations.** I enumerated every call site of all seven widened surfaces at this head:

| | error object | our literal / bridge string |
|---|---|---|
| `soql-runner.ts:1139` | :1506, :1602, :1634 | :1456, :1589 |
| `rest-explore.ts:221` | :292, **:265 `showError(err, 'The request body must be valid JSON.')`** | :237 |
| `soap-explore.ts:289` | :347, **:315 `showError(err, 'The payload must be valid JSON.')`** | :306 |
| `apex-test-runner.ts:111` | :213 | :177, :185 |
| `bridge-tools.ts:110` | :135, :149 | :142 (`response.error`) |
| `ai-assistant.ts` | :251 | :248 (`response.error`) |
| `field-impact.ts:877` | `renderSfError(err, { doc })` | — |

The two sites I flagged as precedent are fixed, and fixed the right way — the error plus the `guidance` option rather than concatenation. Rendered result confirmed in happy-dom: parser words as node one, our line as node two.
```
<span class="sfdt-sf-error-text">Unexpected token } in JSON at position 42</span>
<span class="sfdt-sf-error-note">The request body must be valid JSON.</span>
```
Noting that an option with zero callers is its own smell was the right read.

**`.userFacing` does not cross the worker boundary — verified by reading and by test.** `buildRequestError` is called at `lib/salesforce-api.ts:457` and `:515`, i.e. *after* `await sendProxied(…)` resolves. What crosses `chrome.runtime.sendMessage` is a plain `SfApiFetchResponse` (`{ok, status, bodyText, errors}`), which I confirmed round-trips `structuredClone` losslessly. The `SalesforceRestError` is constructed in the renderer's own realm. **No regression against what #325 shipped.**

**The nit-1 refactor works end to end through the real client.** Driving `SalesforceApiClient` with a mocked bus and a two-record org body whose first record's `message` is multi-line:
```
userFacing.orgText = "Salesforce GET request failed (HTTP 400): No such column 'Foo__c' on entity 'Account'.
                      If you are attempting to use a custom field, be sure to append __c."
userFacing.notes   = 3 notes

<span class="sfdt-sf-error-text">…both org lines, ONE node, error colour…</span>
<span class="sfdt-sf-error-note">×3</span>            role="alert"
```
The org's continuation line stays in the org's node and the org's colour. Removing the guess rather than improving it was the right call.

**Nit 2's replacement is not vacuous.** Mutation-tested: deleting `white-space: pre-wrap` from `.sfdt-sf-error-text` produces `× the org text keeps the newlines it arrived with`. And it is now genuinely load-bearing, because the structured path means `orgText` really can contain newlines in one node — which the original assertion could never exercise.

**All my #325 clearances still hold after the refactor.** Pre-#308 defect reconstructed at `features/soap-explore.ts` (classless div, no white-space rule, `showError(message: string)` + `errorPanel.textContent = message`): three cases fail, including rule 3's `finds the funnels this PR widened, and not the health modal`. No dangling `errorPanel()` callers anywhere. `metadata-retrieve.ts:1302-1303` still `.sfdt-callout .sfdt-bad` with `role="alert"`.

**Your six-correct-sites claim is right, and understated.** I implemented the wide rule (drop the `.sfdt-console` requirement) and ran it over the tree: **11** sites, not six — `apex-test-runner.ts:145`, `event-monitor.ts:250`, `metadata-retrieve.ts:229`, `schema-browser.ts:307/521/577`, `ui/confirm-dialog.ts:65`, `ui/health-modal.ts:174`, `ui/toast.ts:80`, `options/main.ts:323/677`. I judge all eleven correct code, and I could not construct a narrower formulation that catches a classless `<div>` used as a failure region without flagging a status line — the codebase encodes element role only in the class, and `ai-assistant`'s `loading` had none. **Accepting this as a documented limit is the right outcome.**

**Rule 3's scoping-out of `flow-health-check.ts` is correct** — I confirmed `errorSinks` returns `[]` for a file with no renderer import, and the health modal's `showError(message: string)` is a different surface with its own `.sfdt-msg` contract and its own case in `error-render-newlines.test.ts`.

---

## What I could not verify

- **No real browser.** All happy-dom. The nit-1 change alters the common case, so both-themes remains CSS-rule assertions and token arithmetic rather than eyes on a screen — your own caveat, and I agree with it.
- **Screen-reader announcement.** I confirmed `role="alert"` lands on the element receiving the content, is removed on reuse, and is never applied to an empty panel. Whether a real AT announces a live region populated while `display:none` and then shown is not answerable here.
- **Live-org error shapes.** My multi-line-org-record fixture is synthetic, built to the documented REST error contract. I have no real org response to weigh how often the first record's `message` contains newlines.
- **Timeout / no-session errors through a panel in Chrome** — same limitation you noted; they are plain `Error`s and correctly render as one node in happy-dom.

I pushed nothing, touched no `FEATURES.json`, and left the branch clean.

---

scoobydrew83 (commented):
**VERDICT: REQUEST CHANGES** — one blocking finding. Filed as a `COMMENT` for the mechanical reason that GitHub refuses `REQUEST_CHANGES` on a same-account PR; treat the first line as the verdict.

Round 3, independent review at `4ae470d`, base `cbffd36`. Own worktree, nothing pushed, no `FEATURES.json` touched, tree clean at exit.

---

## How I looked

The three previous holes in this guard were all found by construction, never by reading, so I did not read my way to a verdict. In a fresh worktree, on top of a green `4ae470d`, I injected the #308 defect into **`features/api-name-generator.ts`** (a `.sfdt-console` pane named `preview`, a file this diff does not touch and no previous round used) and into **`features/event-monitor.ts`**:

- **49 compiling injections** into those two real files. Each one was written into the file, normalised with `npx prettier --write` (so the tested text is what Prettier would actually leave — the base file is not Prettier-clean, so `--check` would have been meaningless), then `npx tsc --noEmit` (all 49 exit 0) and `npx vitest run test/sf-error-panel-contract.test.ts test/error-render-newlines.test.ts`. Every injection reverted; `git status` clean after each sweep.
- **41 further probes** through the exported matchers (`rendersErrorIntoConsole`, `buildsSfErrorPanel`, `passesStringifiedError`) for identifier spellings and rule-3 argument shapes.

**Result: 19 of the 33 rule-2 spellings, and 2 of 3 Prettier-wrap shapes, were NOT CAUGHT.** One of them is a regression this PR introduces in round 3.

---

## BLOCKING — B1. The new `LITERAL_ONLY` suppressor on rule 2 is a pure loss: it catches nothing and it un-catches the ordinary spelling

The suppressor added in `4ae470d`:

```ts
const LITERAL_ONLY = /=\s*['"`][^'"`]*['"`]\s*;?\s*$/;
…
if (LITERAL_ONLY.test(whole!)) continue;
```

The character class ``[^'"`]*`` does not exclude `$` or `{`, so **a template literal that interpolates an error is "literal-only"**. Injected into `features/api-name-generator.ts` (Prettier output shown verbatim, `tsc` exit 0):

```ts
} catch (err) {
  preview.textContent = `Could not save: ${err instanceof Error ? err.message : String(err)}`;
}
```

`preview` is a `.sfdt-console` pane. No `.sfdt-error`, no `role="alert"`, org guidance flattened. That is the #308 defect. **The suite stays green: 111 files / 2000 tests passed.**

Same result for every interpolated spelling I tried:

| injected into `features/api-name-generator.ts` (`preview` is `.sfdt-console`) | |
|---|---|
| ``preview.textContent = `Could not save: ${err instanceof Error ? err.message : String(err)}`;`` | **NOT CAUGHT** |
| ``preview.textContent = `${err}`;`` | **NOT CAUGHT** |
| ``preview.textContent = `${String(err)} — settings not saved`;`` | **NOT CAUGHT** |
| ``const message = …; preview.textContent = `${message}`;`` | **NOT CAUGHT** |

### This is a regression introduced by round 3, not a pre-existing gap

Round 2's rule 2 (`058f9af`) had no `LITERAL_ONLY`. I dropped that exact file into the tree under a temp name with the defect present:

```
 FAIL  test/zz-round2-rule2.test.ts > no console pane is handed a caught error directly
AssertionError: these assign an error message straight onto a `.sfdt-console` element,
so the failure carries no error styling and no role="alert":
features/api-name-generator.ts (preview)
```

Round 2 caught it. Round 3 does not. The fix for "the alternation is missing `message`" was paid for with a wider hole than the one it closed — interpolation is a far more common way to render an error than a bare identifier.

### And the suppressor provably buys nothing on rule 2

The justification in the comment is `pane.textContent = 'Loading log…'` — but that string contains no alternation word, so rule 2 never matched it in the first place. I measured it directly. With `if (LITERAL_ONLY.test(whole!)) continue;` short-circuited to `false &&`, on the unmodified tree:

```
Test Files  111 passed (111)
     Tests  2000 passed (2000)
```

**Zero false positives suppressed, tree-wide.** Including the test case written to justify it, which passes without it.

This is golden principle #12's discipline applied one level up: this PR added `every exclusion would actually fail without it` precisely because "an exclusion for a file that trips nothing is not an exclusion, it is a hole". `LITERAL_ONLY` is an exclusion of the same kind, added on the same round, with no such proof — and it does not have the property.

### The fix is free

Either delete `LITERAL_ONLY` from rule 2 (measured above: 2000/2000 still green), or make it interpolation-aware:

```ts
const LITERAL_ONLY = /=\s*(?:'[^']*'|"[^"]*"|`[^`$]*`)\s*;?\s*$/;
```

I ran both. Both catch the defect; both leave the full suite at **111 files / 2000 tests passed**. Note this narrower form is safe **for rule 2 only** — see N1 below for why `error-render-newlines.test.ts` needs a different treatment.

---

## Non-blocking

### N1. `LITERAL_ONLY` in `error-render-newlines.test.ts` is hiding a live offender — `features/event-monitor.ts:250`

Same regex, and there it *is* doing real work — but not only that. Disabling it surfaces four sites:

```
features/data-import.ts:745 (td)
features/event-monitor.ts:250 (limitsContainer)
features/field-creator.ts:398 (tdStatus)
features/metadata-retrieve.ts:229 (text)
```

Three are genuine false positives (`'Error'`, a row-count line mentioning "errors", a progress log line whose object is called `msg`). The fourth is not:

```ts
// features/event-monitor.ts:248
} catch (err) {
  const message = err instanceof Error ? err.message : String(err);
  limitsContainer.textContent = `Failed to load limits: ${message}`;
}
```

`limitsContainer` is a hand-styled `div` (`style.cssText` at line 381, no `white-space`, no class, no `role`). A live Platform-Event-limits failure rendered as bare text with the guidance line collapsed. It is exactly the shape the PR describes as "a bare-`<div>` error render is not mechanically separable from a status line", which is honestly disclosed — but the PR also says `ai-assistant.ts:258` was "found by reading", and this one was not. Pre-existing on `develop`, so not blocking this PR, but it should be named rather than left for round 4 to rediscover. Fixing it in-place with `setSfError` is a one-liner; fixing the guard needs `rendersAnErrorValue` applied to the interpolated segments rather than the whole-line `LITERAL_ONLY` short-circuit (my narrower regex above would turn `data-import.ts:745` and `metadata-retrieve.ts:229` red, so don't just copy it across).

### N2. Rule 2 is line-scoped where rule 1 is expression-scoped — two Prettier-produced shapes walk through

Rule 1 has a nesting- and continuation-aware `readExpression()`. Rule 2 uses `[^;\n]*`, so it reads the assignment line plus (via `\s*` after `=`) at most the first continuation line. Injected and Prettier-formatted:

```ts
preview.textContent =
  'Could not save the naming pattern for this object type: ' +
  (err instanceof Error ? err.message : String(err));       // NOT CAUGHT
```
```ts
preview.textContent = [
  'Could not save the naming pattern.',
  err instanceof Error ? err.message : String(err),
].join('\n');                                                // NOT CAUGHT
```

Both compile, both are what Prettier produces from an ordinary long line. Rule 2 should reuse `readExpression()` rather than a line-bounded class.

### N3. Rule 2's alternation is a name allowlist, and it disagrees with rule 3's in the same file

Round 3 closed `message`/`msg`. The category is still open, and the two rules in this one file now use different sets. Rule 3's `ERRORISH` accepts `ex|caught|reason`; rule 2's alternation does not. Injected as `catch (X) { preview.textContent = String(X); }`:

| catch variable | rule 2 |
|---|---|
| `err`, `Err`, `error` | CAUGHT |
| `e`, `ex`, `caught`, `reason`, `problem`, `thrown` | **NOT CAUGHT** |

And an intermediate `const` whose name is off the list — the ordinary refactor rule 3 explicitly covers:

```ts
const detail = err instanceof Error ? err.message : String(err);
preview.textContent = detail;                                 // NOT CAUGHT
```

Same for `reason`, `failure`, `problem`, `text`, `bodyText`, and a `let out = 'Save failed. '; out += …` accumulator. Word boundaries also drop compounds the codebase actually writes: `errMsg`, `errMessage`, `sfError`, `errorText` — all **NOT CAUGHT**, even though `errorMsg` is on the list.

Cheap structural fix that closes the class rather than adding names: rule 2 already computes `consoles` per element; do the same for "names bound to a flattened error" using rule 3's existing `flattensAnError()` + binding scan, and flag `console.textContent = <any such name>` regardless of spelling. That also picks up the two-hop in-file funnel, which is currently open for rule 2 and only documented for rule 3:

```ts
const show = (m: string): void => { preview.textContent = m; };
show(err instanceof Error ? err.message : String(err));       // NOT CAUGHT
```

### N4. Rule 3's "plausibly an error" narrowing has the same allowlist shape

`ERRORISH = /\b(?:err|error|errors|errorMsg|ex|caught|reason)\b/i`. Through the exported matcher:

```
renderSfError(problem.message)  → NOT CAUGHT
renderSfError(detail.message)   → NOT CAUGHT
renderSfError(thrown.message)   → NOT CAUGHT
renderSfError(failure.message)  → NOT CAUGHT
renderSfError(cause.message)    → NOT CAUGHT
renderSfError(String(problem))  → NOT CAUGHT
renderSfError(err.message)      → CAUGHT
```

**Not a regression** — round 2's `STRINGIFIED_ERROR` was narrower still (`err|error|e` only), so round 3 is a net widening here and the `String(jobId)` false positive it was written for is real. Worth a line in the "holes it does not close" test so it is a decision rather than a surprise.

### N5. `CHANGELOG.md` is now stale on the exact point round 3 fixed

The Added entry still reads: *"the sweep excludes the artifacts that define it: the helper, **the stylesheet declaring the classes**, and this file"*. `4ae470d` removed `lib/ui-styles.ts` from `DEFINING_ARTIFACTS` — that is the N3 fix. `UI-SYSTEM.md` was updated; the changelog was not. (`test/` isn't in the list either — it's excluded by not being scanned.)

### N6. The PR description's gate numbers are for `edb5829`, not for the head commit

The body says "Gates — commit `edb5829`" and quotes 109/1910, root 313/5266, 1.36 MB. Head is `4ae470d`; I measured 111/2000, root 315/5356, 1.38 MB — which is exactly what the head commit *message* claims, so nothing is wrong, but the body is two commits behind and a reader checking it will get a mismatch.

---

## Re-verified round 2's clearances (branch has been rebased since)

**The ten claimed-closed rule-1 forms — all confirmed against a real file, not fixture strings.** Injected each into `features/event-monitor.ts` (untouched by this diff), Prettier-formatted, `tsc` exit 0:

split `classList.add` · `setAttribute('class', …)` · `className +=` · const array + spread · `classList.toggle` · `Object.assign` · template interpolation · wrapped assignment · `let` accumulator · const identifiers — **10/10 CAUGHT**.

**The three documented-open forms behave as documented.** `panels[0].className = 'sfdt-console sfdt-error'` → rule 1 blind, and nothing else catches it (**NOT CAUGHT**, as the comment says). The alias split and `'sfdt-' + kind` are rule-1 blind as documented, but both happened to be caught by rule 2 in my injections — the backstop held.

**The stale-exclusion rewrite (round 2's N3) genuinely bites — it is not theatre.** I added two separate decorative entries to `DEFINING_ARTIFACTS` and the suite went red both times:

```
× every exclusion would actually fail without it
AssertionError: lib/ui-styles.ts is excluded but trips no rule — that is a hole,
not an exclusion. Delete the entry.
```
```
× every exclusion would actually fail without it
AssertionError: lib/icons.ts is excluded but trips no rule …
```

**The `unknown`-funnel audit is still at zero** (rule 3 tree-wide offenders `[]`), **rule 3 still bites on its target** (all 14 `bad` cases fire, all 6 `good` cases stay clean), and the funnel-set test still pins the six funnels and excludes the health modal.

**Rule 1's `/i`-adjacent widening produced no compensating narrowing I could find beyond the two examined above** — rule 1 carries no `/i`; `error-render-newlines` already had it. The only two narrowings added in `4ae470d` are `LITERAL_ONLY` (B1) and `ERRORISH` (N4).

---

## Gates — both sides, same worktree and install

| | `origin/develop` @ `cbffd36` | head @ `4ae470d` |
|---|---|---|
| `extension/ npx vitest run` | 111 files / **1983** passed | 111 files / **2000** passed |
| `extension/ npx tsc --noEmit` | exit 0 | exit 0 |
| `extension/ npx eslint .` | exit 0 | exit 0 |
| `extension/ npx wxt build` | Σ 1.38 MB | Σ 1.38 MB |
| root `npm test` | 315 files / **5339** passed | 315 files / **5356** passed |
| root `npm run check:all-contracts` | exit 0 | exit 0 |

Built `chrome-mv3/manifest.json` is **byte-identical** on both sides (`sha256 2a302f6a1b97539b0857f162e24fb498774b602f8840b515d865ecc9e8359c7d`).

**Diff hygiene:** `cbffd36` is an ancestor of `4ae470d` (no rebase drift). 20 files, all under `extension/`. `git diff cbffd36..4ae470d -- generated/ FEATURES.json package.json package-lock.json extension/package.json extension/wxt.config.ts` is **empty**. No already-merged #325 content re-appears — every hunk is follow-up.

**Principle #12:** the one remaining exclusion (`ui/panels.ts`) is proved load-bearing by two tests, one generic and one naming the specific rule. The *removal* of `lib/ui-styles.ts` is the right call and is itself pinned. The gap is that the same discipline was not applied to `LITERAL_ONLY` (B1).

---

## What I could not verify

- **No browser.** happy-dom only. I did not load `.output/chrome-mv3` against an org, so the a11y claim (`role="alert"` actually announced) and the two-tone rendering in light/dark remain CSS-rule assertions.
- **Whether the injections I did not think of are caught.** 33 rule-2 spellings, 13 rule-1 forms, 3 Prettier-wrap shapes, 25 identifier names, 10 rule-3 argument shapes. That is not the space. The pattern across three rounds is that every hole was found by construction, and I found more in one sitting than the previous two rounds combined — I would not read that as "the space is nearly exhausted".
- **`structuredClone`/`postMessage` loss of `.userFacing` (N4 in the commit message).** I read the type and the comment; I did not drive an error across the worker boundary to confirm the property is dropped as described.
- **The runtime behaviour of the `sf-error-guidance` restructure** beyond what `panels.test.ts` and `org-limits.test.ts` assert — I reviewed the diff but my testing effort went to the sweep, per the brief.
- **Whether the six "legitimate inline error string" sites rule 2 deliberately scopes out are all still legitimate.** I took that claim on the PR's word.


---

scoobydrew83 (commented):
**REQUEST CHANGES** — three blocking findings. The first is a regression this round introduces, of exactly the shape rounds 2 and 3 introduced: the new machinery hides the canonical #308 defect, and I found it by construction, not by reading.

> Filed as a `COMMENT` because GitHub refuses `REQUEST_CHANGES` on a same-account PR. Line 1 is the verdict.

Round 4, independent review at `1f802d2`, base `cbffd36`. Own worktree under `.claude/`, nothing pushed, `FEATURES.json` untouched, tree clean at exit.

---

## How I looked

Reading this code and finding it reasonable has a perfect record of being wrong, so I did not read my way to a verdict. I built a probe that drives the real exported matchers (`buildsSfErrorPanel`, `rendersErrorIntoConsole`, `passesStringifiedError`) plus a faithful replica of `error-render-newlines`' per-file scan, splices a candidate block into the **real source of a real file**, and reports which of the four guards fire.

- **102 constructed spellings** across three waves, injected into **`features/rest-explore.ts`** (`responsePane`, a real `.sfdt-console`) and **`features/apex-anonymous.ts`** (`logPane`, ditto) — two files no round has used for injections, neither of them in your burned list.
- **7 end-to-end injections** written into the file on disk, Prettier-checked against the project config, `npx tsc --noEmit`, `npx eslint <file>`, both guard suites, then `git checkout`. Every revert verified clean.
- The five e2e blocks all reported `prettier` **block-clean** and `tsc=0 eslint=0`. I confirmed your warning first: `features/rest-explore.ts`, `apex-anonymous.ts`, `event-monitor.ts`, `metadata-retrieve.ts` are **not** Prettier-clean at base (`npx prettier --check` warns on all four), so I never ran `--write` on a whole file — only the added block, through `prettier --parser=typescript`.

**Result: 21 of 102 NOT CAUGHT.** One is your documented two-hop funnel. Three are categories worth blocking on.

---

# BLOCKING

## B1 — `dynamicParts()` reads every backtick as a template delimiter, so one stray backtick blinds **both** guards over an arbitrary span. Regression vs `4ae470d`.

`dynamicParts()` is new in this round and both guards now scan its output instead of raw source. Its template branch is:

```ts
if (ch === '`') {
  i++;
  while (i < n && source[i] !== '`') { … blank(i, i + 1); i++; }
```

Unlike the `'`/`"` branch two lines up, **it is not newline-bounded**. A backtick that does not open a template literal — in a comment, in a regex character class — starts a mask that runs to the next backtick in the file, or to EOF.

In real TypeScript a template literal cannot be left unterminated, so an unmatched backtick can only come from a comment or a regex — and `dynamicParts()` parses neither.

### Demonstrated, end to end, in `features/rest-explore.ts`

The injected block is the canonical #308 defect with **one ordinary comment line** above it. `responsePane` is a `.sfdt-console` with no `.sfdt-error` and no `role="alert"`:

```ts
// The org sometimes returns a ` fenced block in its message.
responsePane.textContent = err instanceof Error ? err.message : String(err);
```

```
── CTRL   (the same two lines, without the comment)
   prettier=block-clean  tsc=0  eslint=0
    Test Files 1 failed | 1 passed (2)  Tests 1 failed | 36 passed (37)
   CAUGHT BY:
     × no console pane is handed a caught error directly 331ms
── R01   (with the comment)
   prettier=block-clean  tsc=0  eslint=0
    Test Files 2 passed (2)  Tests 37 passed (37)
   *** NOT CAUGHT BY ANY GUARD ***
```

### It is a regression, not a pre-existing gap

Same two injections, same harness, against your previous head:

```
── CTRL @4ae470d      Tests 1 failed | 28 passed (29)
   CAUGHT BY: × no console pane is handed a caught error directly
── R01  @4ae470d      Tests 1 failed | 28 passed (29)
   CAUGHT BY: × no console pane is handed a caught error directly
```

`4ae470d` catches it because round 3's rule 2 matched against raw source. Round 4 masks first, and the mask eats the assignment. Rounds 2 and 3 each closed a hole and opened a wider one; this is the third.

### Blast radius, measured on a real file

One backtick appended to the **existing** header comment at `features/rest-explore.ts:38`:

```
backtick injected at line 38
masking runs to line 108 (of 357)
characters blanked: 7588
.textContent= occurrences in that span, raw: 1, after masking: 0
```

Seventy lines of a live feature file go dark in rules 2 and 3 and in `error-render-newlines`, silently, with no test failing. Rule 1 is unaffected — it still reads raw source — which is why nothing tips you off.

### The trigger is not exotic; it is this codebase's comment style

```
93 of 124 scanned files carry backticks inside comments; 1819 backticks total
```

Every one of those 1,819 is currently balanced by convention. It only takes one — a markdown fence, a `` don`t `` typo, a ``/[`]/`` in a regex — and the guard goes quiet over a region nobody can see.

I checked whether it is already biting. It is not, yet:

```
TOTAL CODE TOKENS LOST TO MASKING: 5
features/soql-bulk-delete.ts   /\brenderSfError\s*\(/g   raw=1 masked=0
ui/panels.ts                   /\brenderSfError\s*\(/g   raw=2 masked=1
lib/sf-error-guidance.ts       /\brenderSfError\s*\(/g   raw=1 masked=0
lib/ui-controls.ts             /\.textContent\s*=/g      raw=4 masked=3
lib/ui-styles.ts               /\brenderSfError\s*\(/g   raw=1 masked=0
```

All five are mentions inside comments, harmlessly masked. `dynamicParts()` is length- and newline-preserving on every scanned file — no drift, so the offset claim holds. This is a landmine, not a live fire.

**The fix is small.** Strip comments before masking (or at minimum bound the backtick scan the way the quote scan is bounded, and skip `//`-to-newline and `/*…*/` first). Whatever you choose, the property to pin is the one I measured: *masking must never remove a code token that exists in the raw source*. That assertion, run over the tree, is a check that cannot be satisfied by a pattern that guesses.

The probes that share the cause, for completeness:

| | |
|---|---|
| comment with ONE backtick, then the canonical defect | **NOT CAUGHT** |
| comment with THREE backticks, then the canonical defect | **NOT CAUGHT** |
| ``'x'.replace(/[`]/g, '')``, then the canonical defect | **NOT CAUGHT** |
| one backtick in a comment, defect **40 lines** later | **NOT CAUGHT** |

---

## B2 — the catch scan misses `.catch((x) => …)`, which is the form Prettier emits and the form this codebase writes

The round's headline claim is that spelling no longer matters: *"every `catch (X)` binding — X **is** the thrown error, whatever it is called"*. The implementation is

```ts
for (const m of code.matchAll(/\bcatch\s*\(\s*([A-Za-z_$][\w$]*)/g)) holdsError.add(m[1]!);
```

`\s*` does not match `(`. So `catch (err)` is seen and **`.catch((err) => …)` is not** — and with `arrowParens: "always"` (Prettier 3's default, and this repo's `.prettierrc` has no override), the parenthesised form is the only one Prettier will ever produce for a typed or plain single parameter. The tree already writes it: `lib/spa-router.ts:40`, `entrypoints/background.ts:469`.

Those two happen to be spelled `err`, so the *spelling* fallback covers them. Rename one and the guard is blind.

Demonstrated end to end in `features/rest-explore.ts`:

```ts
void Promise.resolve().catch((e) => {
  responsePane.textContent = String(e);
});
```

```
── P01
   prettier=block-clean  tsc=0  eslint=0
    Test Files 2 passed (2)  Tests 37 passed (37)
   *** NOT CAUGHT BY ANY GUARD ***
```

| probe | |
|---|---|
| `.catch((e) => { pane.textContent = String(e); })` | **NOT CAUGHT** |
| `.catch((reason) => …)` | **NOT CAUGHT** |
| ``.catch((e) => { pane.textContent = `Could not refresh: ${String(e)}`; })`` | **NOT CAUGHT** |
| `.then(() => …, (oops) => …)` rejection handler | **NOT CAUGHT** |
| `try { … } catch (e) { … }` — statement form, control | CAUGHT |

The statement form and the promise form are the same binding to a reader and to the language. Closing this is `catch\s*\(\s*\(?\s*([A-Za-z_$][\w$]*)`, plus a case in `rule 2 knows an error by its binding, not only by its spelling`. It is the same one-token class of fix as round 3's blocker — and, like that one, it is in the mechanism the round advertises as having closed the category.

---

## B3 — `every exemption would actually be flagged without it` does not check that. It accepted a decorative entry.

You added this as *"the check whose absence let `LITERAL_ONLY` in"*. I added a decorative entry to `EXEMPT`:

```ts
{
  file: 'ui/toast.ts',
  name: 'toast',
  expression: 'message',
  because: 'decorative probe entry added by the reviewer to test that the bite check actually bites here',
},
```

```
--- decorative EXEMPTION ---
      Tests  12 passed (12)
```

Green. `ui/toast.ts` is provably **not** an offender — `toast.style.cssText` carries `'white-space: pre-line'` (`ui/toast.ts:78`), and the suite is green at head with no such exemption. So the entry buys nothing and costs an assignment's worth of coverage, permanently and invisibly.

The check asserts two things: that the assignment still exists in the source, and that `rendersAnErrorValue(entry.expression)` is true. Neither is the property an exemption claims. It never asks whether the site would be **reported** — it does not run `setsWhiteSpaceDirectly`, `setsWhiteSpaceByClass`, `fromPanelBuilder`, or the `cssText` test. Any site whose expression is error-ish and which is already covered by one of those four sails through.

That is the same weaker-property mistake as the old stale-exclusion test that vouched for `lib/ui-styles.ts` for two rounds, in the check written this round to prevent that mistake. **The sibling check on the sweep is correct** — I added a decorative `lib/icons.ts` to `DEFINING_ARTIFACTS` and it went red:

```
--- decorative EXCLUSION ---
     × every exclusion would actually fail without it 12ms
      Tests  1 failed | 24 passed (25)
```

The exemption check needs the same shape: run the offender loop with the entry removed and require the site to appear.

**No live hole today.** The one real entry is load-bearing — deleting it turns the suite red:

```
--- real exemption REMOVED ---
     × every element assigned an error message declares a white-space rule 120ms
      Tests  1 failed | 11 passed (12)
```

And **the exemption list gained no new entry this round** — confirmed, `EXEMPT` still has exactly one member.

---

# CAUGHT / NOT CAUGHT — 102 constructed spellings

Injected into `features/rest-explore.ts` (`responsePane`) and `features/apex-anonymous.ts` (`logPane`).

### The template-literal shape that defeated round 3 — all closed

| | |
|---|---|
| ``  `Could not send: ${err instanceof Error ? err.message : String(err)}` `` | CAUGHT (R2) |
| `` `${err}` `` | CAUGHT (R2) |
| `` `${String(err)} — not sent` `` | CAUGHT (R2) |
| `` const message = …; `${message}` `` | CAUGHT (R2+R3) |
| `` const message = …; `Failed to load: ${message}` `` | CAUGHT (R2) |

### The 19 rule-2 spellings round 3 found NOT CAUGHT — re-run, all closed

`catch (e)` · `catch (ex)` · `catch (caught)` · `catch (reason)` · `catch (problem)` · `catch (thrown)` · `const detail` · `const failure` · `const problem` · `const text` · `const bodyText` · `let out` `+=` accumulator · alias one hop · `const errMsg` · `const errMessage` · `const sfError` · `const errorText` · Prettier wrap with trailing `+` · Prettier wrap with array `.join()` — **19/19 CAUGHT (rule 2)**.

### The new structural path

| | |
|---|---|
| `catch (splat)` — a name off every plausible list | CAUGHT |
| `catch (q)` — single letter | CAUGHT |
| 1, 2, 3, 4, **5** hops of intermediate `const` | CAUGHT (all five — the fixed point holds past the 4-pass loop) |
| ternary at the assignment | CAUGHT |
| `?? 'Unknown failure'` fallback | CAUGHT |
| `\|\| 'Unknown failure'` fallback | CAUGHT |
| `const shown = (err as Error).message ?? '…'` then assign | CAUGHT |
| helper in another module, error passed as the argument | CAUGHT |
| IIFE returning the text | CAUGHT |
| `[String(err)].map(…).join('')` | CAUGHT |
| stashed on an object field, read back | CAUGHT |
| module-scope `let` written in the catch | CAUGHT |
| `for (const row of records) pane.textContent = row.message` | CAUGHT |
| `records[0]?.message` | CAUGHT |
| **array destructuring** `const [first] = [String(err)]` | **NOT CAUGHT** |
| **destructure with rename** `const { message: line } = err as Error` | **NOT CAUGHT** |
| **destructure then alias** `const { message } = err; const line = message` | **NOT CAUGHT** |
| **`const [firstLine] = String(err).split('\n')`** | **NOT CAUGHT** |
| pushed into an array, `.join()`d back | NOT CAUGHT |
| `Map.set` / `Map.get` | NOT CAUGHT |
| `catch (данные)` — non-ASCII identifier | NOT CAUGHT |
| **two-hop funnel through a function parameter** | **NOT CAUGHT** (documented) |

### Flattening idioms — all closed

`String(x)` · `` `${x}` `` · `(x as Error).toString()` · `(x as Error).message` · `(x as Error \| undefined)?.message` · `(x as Error \| undefined)!.message` · `JSON.stringify(err)` · `String(err).trim()` · `` `${(err as Error).message}` `` · `String(err).split().join()` · `'Failed: ' + String(err)` · `String(err).replace(…)` · `const flat = (err as Error).toString()` — **13/13 CAUGHT**.

### The known-open two-hop funnel — the test does assert it, and I judge it acceptable

`the two-hop in-file funnel is open, and that is a decision` asserts `rendersErrorIntoConsole(…)` `.toEqual([])` on the exact shape. It is a real assertion, not a comment, and it will go red if the hole is ever closed — a mild trap, but the honest one, and rule 3 already carries the same pattern. Leaving it open is fine: it needs parameter-level dataflow, and the shapes it covers are one refactor away from the binding scan anyway. **The two destructuring holes above are not in the same category** — they need no dataflow, only a pattern for `const {` / `const [`, and `const { message } = err` is far more ordinary authoring than a local funnel.

---

# Non-blocking

### N1 — rule 2 watches `X.textContent =` and nothing else. Six other ways to hand a pane an error walk through.

Demonstrated e2e in `features/rest-explore.ts`, `prettier=block-clean tsc=0 eslint=0`, `Tests 37 passed (37)`:

```ts
responsePane.append(err instanceof Error ? err.message : String(err));   // NOT CAUGHT
```

| | |
|---|---|
| `pane.append(<string>)` | **NOT CAUGHT** |
| `pane.appendChild(doc.createTextNode(String(err)))` | **NOT CAUGHT** |
| `pane.insertAdjacentText('beforeend', String(err))` | **NOT CAUGHT** |
| `pane.replaceChildren(String(err))` | **NOT CAUGHT** |
| `Object.assign(pane, { textContent: String(err) })` | **NOT CAUGHT** |
| `pane['textContent'] = String(err)` | **NOT CAUGHT** |
| `pane.innerText = String(err)` | **NOT CAUGHT** |

`.append(…)` and `appendChild(createTextNode(…))` are both used across this codebase (`features/bridge-tools.ts:247`, `ui/apex-log-console.ts:84`, `features/metadata-retrieve.ts:1259` …), and for a *log* pane appending is at least as natural as replacing. The `Object.assign` one is the sharpest: rule 1 already handles `Object.assign(el, { className })` through `OBJECT_ASSIGN`, so the guard knows the form exists and only reads it for classes. This is the "new category rather than a new spelling" you predicted; I would not block on it, but it is bigger than the destructuring gap.

### N2 — `error-render-newlines` still asks only the spelling question

The module header says both guards now share one definition of what an error is. They share the plumbing, but the newlines guard's entry point is `rendersAnErrorValue(expression)`, which takes **no `names` argument** and never consults `errorBoundNames()`. So every structural gain of this round — the catch scan, the binding scan, the fixed point — reaches rules 2 and 3 only. Measured:

```
S01  pane created in-file but classed elsewhere; catch (glitch) { pane.textContent = String(glitch) }
     rule2: scoped out (not a .sfdt-console)   NL: blind (`glitch` is not an error spelling)
     *** NOT CAUGHT ***
```

Not a regression — round 3's newlines guard was alternation-only too — but the header oversells the unification, and this is where the next reader will be surprised.

### N3 — `wordsOf()` splits on `[^A-Za-z0-9]+`, so digits stay glued on

`err2`, `error1`, `msg2` are not error names. It cost me a control case (`.catch((err2) => …)` came back NOT CAUGHT and I had to re-derive why). Numbered locals are common enough in a `catch` next to an outer `err` that this is worth a `\d` boundary in the camelCase splitter.

### N4 — `errorSinks()` over-includes much more widely than the comment implies

The comment describes an approximation of "forwards to the renderer". Measured, `features/soql-runner.ts` yields ten sinks: `renderSfError, setSfError, isPlainRecord, searchRecordsFrom, groupSearchRecords, formatCell, isRecord, findNodes, showError, fmt`. `features/field-impact.ts` yields `renderSfError, message`. Harmless — over-inclusion only ever adds failures to look at, and the tree-wide rule-3 offender set is `[]` — but "the exact set this PR widened" is not what the function returns, and `rule 3 finds the funnels this PR widened` only asserts `toContain`, so it would keep passing however wide the set grew.

### N5 — the empty-`Error('')` swallow is still live at this head, and the round-4 reply does not mention it

The bot review on `4ae470d` posted this inline on `ui/panels.ts:110`. Measured at `1f802d2` in happy-dom:

```
renderSfError(new Error(''))     -> outerHTML="<div class=\"\"></div>" role=null text=""
setSfError(pane, new Error(''))  -> class="sfdt-console" role=null text=""
control new Error('boom')        -> class="sfdt-console sfdt-error" role=alert text="boom"
```

`sfErrorParts()` falls back to `errorText(error)`, which is `''`, so `parts.length === 0` and `setSfError` takes the deliberate-null branch and calls `clearSfError`. A caught `new Error('')` — or a `throw new Error()` — renders as nothing at all, from every call site that forwards a caught error. It is the defect class this PR exists to close, arriving through the null contract this PR introduced. Not blocking because I could not show a live path that produces an empty message, but it deserves an answer rather than silence.

### N6 — `entrypoints/` is scanned by the sweep and not by the newlines guard

`error-render-newlines`' `SCANNED_DIRS` is `['features', 'ui']`; the sweep's is `['features', 'ui', 'entrypoints', 'lib']`. In the gap:

```
entrypoints/options/main.ts:682 (saveStatus) = err instanceof Error ? err.message : String(err)
```

A `<span class="status show error">` whose stylesheet rule at `entrypoints/options/main.ts:126` declares no `white-space`. An org error's guidance line collapses there today. Pre-existing on `develop`, outside this diff, so not this PR's bug — but it is the #308 defect and neither guard can see it.

---

# Verified — everything else held

**Gates, both sides, same worktree and install.** Baseline measured by checking out `cbffd36` in the same tree; `packages/flow-core` built once, `wxt prepare` re-run on each side.

| | `origin/develop` `cbffd36` | head `1f802d2` |
|---|---|---|
| `extension/ npx vitest run` | 111 files / **1983** passed | 111 files / **2009** passed |
| `extension/ npx tsc --noEmit` | `tsc exit=0` | `tsc exit=0` |
| `extension/ npx eslint .` | `eslint exit=0` | `eslint exit=0` |
| `extension/ npx wxt build` | `Σ Total size: 1.38 MB` | `Σ Total size: 1.38 MB` |
| root `npm test` | 315 files / **5339** passed | 315 files / **5365** passed |

Every number in the PR body reproduces exactly. Built manifest, both sides:

```
2a302f6a1b97539b0857f162e24fb498774b602f8840b515d865ecc9e8359c7d  .output/chrome-mv3/manifest.json
```

Byte-identical, and the same hash you and round 3 measured. `npm run generate:catalogs` → `git status --porcelain generated/` empty (no drift). `git diff cbffd36..1f802d2 -- generated/ FEATURES.json package.json package-lock.json extension/package.json` is **empty**.

**Diff hygiene.** `cbffd36` is an ancestor of `1f802d2` — no rebase drift. 24 files, all under `extension/`. Of the files #325 merged at `e493254`, the six this PR must not re-touch (`org-limits.ts`, `code-coverage.ts`, `trace-flags.ts`, `flow-quality.ts`, `schema-browser.ts`, `debug-log-viewer.ts`) are all **untouched**. Follow-up-only, confirmed.

**`LITERAL_ONLY` suppressed nothing — reproduced independently.** I checked out `4ae470d`, edited line 320 to `if (false && LITERAL_ONLY.test(whole!)) continue;`, and ran the whole suite:

```
 Test Files  111 passed (111)
      Tests  2000 passed (2000)
```

Zero false positives suppressed on rule 2, tree-wide. Deleting it rather than narrowing it was right, and it is the same sentence this PR added one round earlier about `lib/ui-styles.ts`.

**`readExpression()` replacing `[^;\n]*` — 15 shapes, all read correctly.** `;` inside a string on the RHS · `;` inside a template on the RHS · arrow body ending in `}` · assignment inside an object-literal argument · multi-line ternary with operators leading · with operators trailing · a comment between `=` and the value · a line comment containing an apostrophe · a block comment with an unbalanced quote · a regex containing a quote · nested template literals · an escaped backtick inside a template · an escaped quote in a single-quoted string · an apostrophe in template prose · both Prettier wrap shapes. **15/15 CAUGHT.** The `ENDS_OPEN`/`CONTINUES_LINE` pair genuinely handles both sides of a Prettier wrap; I could not construct a wrap it misreads.

**`dynamicParts()` — everything except backticks.** Nested quotes, escaped quotes, quotes inside templates, templates inside templates (the recursion works), a regex containing a `'` or `"` (the newline bound contains the damage), a comment with an unbalanced apostrophe (same), and the Python code template in `features/soql-runner.ts` — ``identifiersIn(dynamicParts('const py = `\nquery = """\n`;'))`` is `['const','py']`, and `soql-runner.ts` produces no phantom binding. Length- and newline-preserving on all 124 scanned files, verified by direct comparison, so the offset claim is real. Only the unbounded backtick breaks it (B1).

**Rule 1 — all twelve claimed-closed forms still closed after the rewrite:**

```
  split statements         CAUGHT      classList.toggle         CAUGHT
  setAttribute             CAUGHT      classList.replace        CAUGHT
  className +=             CAUGHT      Object.assign            CAUGHT
  template interpolation   CAUGHT      wrapped assignment       CAUGHT
  const array + spread     CAUGHT      let accumulator          CAUGHT
  const identifiers        CAUGHT      var accumulator          CAUGHT
```

**…and the three documented-open forms behave exactly as documented:**

```
  panels[i].className      MISSED (as documented)
  alias split              MISSED (as documented)
  runtime 'sfdt-' + kind   MISSED (as documented)
```

**The `unknown`-funnel audit is at zero.** Running all three rules over every scanned file, the only file that trips anything is `ui/panels.ts` (`r1=true r2=["el"] r3=[]`) — the one excluded defining artifact, which is exactly what `the helper trips rule 1 through its own constant` pins.

**Rule 3 bites on its target:** `renderSfError(err instanceof Error ? err.message : String(err))`, `setSfError(pane, err.message)`, `renderSfError(String(err))` and `const msg = String(err); renderSfError(msg)` all FLAGGED.

**The N3 stale-exclusion rewrite genuinely bites** — decorative `lib/icons.ts` entry → `× every exclusion would actually fail without it`, `1 failed | 24 passed (25)`. **`lib/ui-styles.ts` is back in the scan** and `the stylesheet is scanned like everything else` pins it.

**Both source fixes are load-bearing, mutation-tested.** Deleting `text.className = 'sfdt-msg';` from `metadata-retrieve.ts` → `× every element assigned an error message declares a white-space rule`, `1 failed | 11 passed (12)`. Reverting `features/event-monitor.ts` to `cbffd36` → three failures, `3 failed | 41 passed (44)`, including the behavioural `× announces a limits failure as an error, and stops announcing it on retry`.

**No false positives on the shipped rules.** Rule 2 tree-wide: `[]`. Rule 3 tree-wide: `[]`. The newlines guard: green. I also ran the **wide** variant of rule 2 with the `.sfdt-console` requirement dropped, to see the surface you would take on by widening further — **14 sites** now (round 3 measured 11 under the narrower name set), and I judge all fourteen correct code:

```
features/apex-test-runner.ts:145 (msg) = f.message          [has white-space: pre-wrap in cssText]
features/data-import.ts:722 (tdErrors) = row.errors || ' '
features/metadata-retrieve.ts:235 (text) = `${level.word} ${msg.text}`
features/schema-browser.ts:307 (countLabel) = global.error ? …
features/schema-browser.ts:521 (why) / :577 (why) = ` ${cell.error ?? ' '}`
ui/confirm-dialog.ts:90 · ui/health-modal.ts:174 · ui/toast.ts:80
ui/panels.ts:153 / :174 / :215
entrypoints/options/main.ts:328 (response.error) / :682  [N6]
```

Keeping the console scope and documenting the limit remains right.

**Removing `reason`/`problem`/`thrown`/`failure` from the spelling list reopened nothing.** Constructed with exactly those names: `catch (reason)`, `catch (problem)`, `catch (thrown)` all **CAUGHT** by the catch scan, and `const failure = err instanceof Error ? …` **CAUGHT** by the binding scan. The structural path really does cover the names spelling could not afford to claim. `${response.error}`, `${root}` in `schema-browser.ts` and `summary.message` in `apex-anonymous.ts` are all correctly declined, and `const parentJobId = String(jobId)` still does not fire.

**Adjudication — `metadata-retrieve.ts:229`: you are right, round 3's reviewer was wrong.** `addLog(level, text)` at `:197` renders through the span at `:235`, and `addLog('error', …)` interpolates `err.message` at `:450`, `:623`, `:796`, `:894`, `:1000` (plus `:395` at `info`). Since `lib/sf-error-guidance.ts` that message is two lines. The span carried no class and no `white-space`, so the guidance line collapsed into the org's text — the #308 defect, in the retrieve log. `.sfdt-msg` (`white-space: pre-line`, `lib/ui-styles.ts:670`) is the right class and fixing the source rather than exempting it is the right call. It is also mutation-tested load-bearing (above). Reading what `msg.text` can hold before writing an exemption is the discipline; calling it a false positive on the strength of the variable's name is what the whole spelling-versus-structure argument in this PR is about.

**Golden principle #12:** `DEFINING_ARTIFACTS` gained no entry (one member, `ui/panels.ts`), `EXEMPT` gained no entry (one member, `ui/panels.ts`/`el`/`message`), both live entries are proved load-bearing by mutation. The exclusion bite-check works; the exemption bite-check does not (B3).

---

## What I could not verify

- **No real browser.** happy-dom only, for the fourth round running. `role="alert"` landing on the element is a DOM fact I confirmed; whether an AT announces the limits pane when it toggles is not answerable here. The limits pane's `cssText` warning-tint under a new `.sfdt-console` is still reasoned-about, not looked at — your own caveat, and I agree with it.
- **Whether `new Error('')` is reachable from a real org.** I confirmed the swallow (N5) but did not trace `buildRequestError` against a response that would produce an empty message, so I cannot say whether the path is live or only latent.
- **Whether an unbalanced comment backtick will actually occur.** B1's regression and its blast radius are measured facts; the 1,819-backtick count is a measured fact; that one of them will eventually be odd is a judgement. I did not find one in the tree today.
- **The injections I did not think of.** 102 spellings across two fresh files and seven end-to-end runs is more than any previous round, and it is still not the space. Two of my three blockers are *categories* the structural rewrite does not cover — a different DOM API, a different binding syntax — which is your own prediction about where round 5 would find things, arriving one round early. I would not read the 21 misses as the tail.
- **The runtime behaviour of the `sf-error-guidance` restructure** beyond what `panels.test.ts`, `org-limits.test.ts` and `event-monitor.test.ts` assert. My effort went to the sweep, per the brief.

Nothing pushed, no `FEATURES.json` touched, all probe files deleted, worktree clean at `1f802d2` with `111 passed (111) / 2009 passed (2009)`.

---

PR #328, opened 2026-08-05
feat(extension): NL→SOQL generation in the SOQL runner (C-P4-5)

scoobydrew83:
**C-P4-5 · NL→SOQL in soql-runner · M · Depends: P2-1 (done).**

Adds **Generate query** to the SOQL runner's toolbar: a disclosure panel above the editor where you describe a query in plain English, and the generated SOQL lands **in the editor** for review.

Docs PR: **scoobydrew83/sfdt-site#48**

---

## Review follow-ups (post-approval, commit `b0eccc9`)

The review approved with nothing blocking but found four things worth fixing. All four are in, each with a test that was checked by mutation (revert the fix, the test goes red):

| # | Finding | Fix |
|---|---|---|
| **N1** | A short description defeated the record-leak gate entirely. The user's request was subtracted from the prompt with `prompt.split(request).join('\n')` — a **global** find-and-replace — so a one-character description like `o` deleted every `o` in the prompt, took the leaked value apart with it, and the scan matched nothing. | Subtract the **occurrence**, not the characters. `buildGeneratePrompt()` assembles by index and reports `requestRange`; `generateSoql()` passes it through the gate context; the gate masks that span. A minimum-length floor was **rejected** — it just moves the bypass, which the `Zenith P` case (8 chars, over the gate's own floor) pins. |
| **N2** | `LEAK_MAX_DEPTH = 3` reached two relationship hops. | The walk is depth-independent now, with a cycle guard for termination. Rationale is in the function's comment; tested at five hops (`A__r.B__r.C__r.D__r.E__r.Name`), through a parent-to-child subquery envelope, at 40 levels, and against a cyclic row. |
| **N4** | A stale success path overwrote the editor: `nlGenerateBtn.disabled` was checked at dispatch but not after the `await`, so an answer arriving after the user switched to SOSL destroyed the FIND query they had typed and flipped the language back. | `nlGenerationStale()` compares view identity, panel state and language against a snapshot taken at dispatch; the whole outcome is abandoned, silently. |
| **N5** | `aria-haspopup="true"` on a `role="group"` disclosure promised a menu and delivered a form. | Removed. `aria-expanded` + `aria-controls` was already the complete pattern. |

**N3** is a PR-body correction, not a code change — see the withdrawn `host_permissions` claim under "What I could NOT verify"; the same retraction is now in the source too (see the re-verify follow-ups below). **N7** is fixed in the docs PR. **N6** is fixed below. **N8/N9** are recorded rather than changed; see the closing comment for why.

---

## Re-verify follow-ups (commit `666b5fd`)

The independent re-verification of `b5a3e90..b0eccc9` cleared it to merge with nothing blocking, and listed five small items. All five are in. Nothing below changes behaviour the review measured; two are comment-accuracy fixes and one is a test-only refactor.

| # | Finding | Fix |
|---|---|---|
| **F1** | `subtractRequest()`'s no-span fallback failed **OPEN**. With no `## Request\n` marker it searched from index 0, so `indexOf(request, 0)` could delete the first occurrence of the request text from *inside the leaked value* — unmasking nothing and hiding the leak. N1's failure mode in miniature. Unreachable today (`soql-runner.ts` is the only caller and always passes a verified span), but it was the one path in this function that did not fail closed. | No marker ⇒ **no guess**: return the prompt whole and scan all of it. The accepted cost is a possible false block on the user's own words, asserted rather than hidden — a false block is a visible refusal, a missed leak is silent. |
| **F2** | `nlGenerationStale()` was a 4-clause guard with **3 untested clauses**. Dropping each separately reddened a test only for `!nlPanelOpen`; the guard could be reduced to that single clause with the suite still green. The other three are unreachable through the UI only because `setLang()` and `presentView`'s `onClose` both call `closeNlPanel()` first — a structural accident an ordinary refactor could remove silently. | The rules move to an exported pure predicate, `nlGenerationIsStale(at, now)`; the closure reads its own state and delegates. **Behaviour unchanged.** Each clause now has a case that varies only that clause. |
| **N3 (source)** | The withdrawn `host_permissions` claim was retracted from this description and **left in the code**: `test/no-llm-endpoints.test.ts`'s header still told the next reader that a new endpoint "would need … a new permission". The description is read once; that comment ships. | Retracted there too, in the same terms — no `content_security_policy`, so no `connect-src`; a cooperating `ACAO: *` host is not blocked; what guards this today is a source-level lint over two files plus review. The `host_permissions` assertion stays, described as a tripwire on the permission ledger rather than an egress control. |
| **N6** | Four lines away in the same header: "this file, **and the docs that describe the rule**, are excluded from their own scan". `DEFINING_ARTIFACTS` holds exactly one entry. | Corrected to what the array actually contains: only this file, by name and with a reason. The docs are not excluded — they are outside `SCANNED_DIRS`. Golden principle #12. |
| **`attributes.url`** | Confirmed leakable and missing from the honest-limits list, which is presented as complete. | Added to the list below. |

**Mutation-checked, both code items:**

```
F1 — revert the fallback to `indexOf(request, from)` with from = 0 when there is no heading:
     × subtracts nothing when the prompt has no Request heading (F1)
      Tests  1 failed | 56 passed (57)

F2 — drop each clause of nlGenerationIsStale() separately:
     drop `at.view !== now.view`  → × view identity — the runner was closed, or closed and reopened
     drop `!now.panel`            → × panel existence — there is no panel to have asked from
     drop `!now.panelOpen`        → × panel-open state — the user pressed Close
                                     × does not overwrite the draft after the user closed the generator
     drop `now.lang !== at.lang`  → × language — the editor is in SOSL now
```

Each of the four clauses now reddens a **distinct** test, which is what it did not do before. Clause 3 additionally reddens the pre-existing integration test — the one that was already carrying the whole guard.

No CSP, no eslint fetch ban, no other hardening was added: that gap is disclosed below and stays out of scope.

---

## The two safety properties, and where each is enforced

Both ACs that matter here are negative, so they are built as code *shape* rather than as behaviour a caller has to remember. A reviewer mutation-testing them should find the mutation caught by a test, not by a comment.

### AC-2 — a generated query cannot run itself

`generateSoql()` in `extension/features/soql-nl-generate.ts` is the only orchestrator, and its `GenerateDeps` interface is:

```ts
describeObject(name): Promise
knownObjects(): Promise | readonly string[]
inspectPrompt?(prompt, requestText, ctx): string | null
askAi(prompt): Promise
```

There is **no executor in it**, and the success outcome is `{ status: 'generated', soql: string, … }` — a string. So there is no path from "the model answered" to "the org ran it" for a later change to take by accident: the module that talks to the model cannot reach the org, and the runner's success handler — once it has confirmed it is still the runner that asked (N4) — does five things (`editor.setValue`, resync the language toggle, resync autocomplete, close the panel, focus the editor) and calls no execute.

Proved three ways:

1. **Structurally, against mutation.** `test/soql-nl-generate.test.ts` wraps the deps object in a `Proxy` and pins the exact set of property names the function ever reads:
   ```
   expect([...reads].sort()).toEqual(['askAi','describeObject','inspectPrompt','knownObjects']);
   ```
   Add a `deps.run` and call it — this fails, naming the extra dep.
2. **At the call site.** The wiring test drives a whole generation and asserts `api.query`, `api.toolingQuery` and `api.apiRequest` were **not called**, and that every `apiGet` was a describe.
3. **Non-vacuously.** The next test presses **Run** and asserts `api.query` was called exactly once, with the generated text. So assertion 2 means "it did not run", not "the harness runs nothing".

Plus: the panel contains exactly two buttons (`['Generate','Close']`, asserted), and a generated string that fails the local SOQL checks is *reported* rather than pasted — asserted against a pre-existing editor draft that survives untouched.

### AC-4 — record data cannot reach the prompt

The guarantee is the allowlist, not a filter:

- `schemaFieldsForPrompt()` **rebuilds** each field property by property from `PROMPT_FIELD_PROPERTIES = ['name','label','type','nillable','inlineHelpText']`. An input property not on the list has no route to the output, not even a wrong-typed one.
- `buildGeneratePrompt({ request, schemas })` has **no records parameter** and no client it could fetch one with.
- The rendered table comes from `buildSchemaMarkdown()` — the same function `export-for-prompt` (P2-1) copies to the clipboard, so the format cannot drift into two dialects.

On top of that sits a **gate the orchestrator must clear before it may reach `askAi`**, shaped like bulk delete's backup gate: `recordValueLeak()` scans the assembled prompt for any value from the on-screen result set, and a hit returns `{ status: 'blocked' }` with `askAi` never called. A gate that *throws* is a refusal, not consent.

The user's own request text is excluded from that scan, because their words are theirs — someone typing "accounts named Universal Containers" while that name is in the results table has not suffered a leak. It is excluded **by index**: `buildGeneratePrompt()` reports the span it interpolated the request at (`requestRange`) and the gate masks exactly that region. It used to be excluded by text, which is what N1 was.

Proved:

- Unit: a describe polluted with `records: [...]`, `sampleValue` and `lastRecordId` yields a prompt containing none of them but still containing the field row.
- Unit: `recordValueLeak` finds top-level and nested (relationship) values at any depth; ignores the object/field API names the schema legitimately carries; does not flag the user's own words back at them.
- Unit (N1 regression): a one-character and a two-character description, and an 8-character one that clears the gate's own length floor, all still catch a value spliced into the schema half. Each of these returns `null` against the pre-fix code.
- Unit: gate returns a message ⇒ `expect(d.askAi).not.toHaveBeenCalled()`.
- **End-to-end**: the runner test runs a query returning `Zenith Prosthetics Consortium` / `001000000000001AAA`, confirms both are on screen, then generates and asserts the captured prompt contains **neither** — while still containing `# Schema: Account`, so the assertion isn't passing on an empty prompt.

**Honest limits** (documented on the function, and I want them on the record). The backstop is a substring search over the values it can see, so it is blind to:

- values under 8 characters — a `true`, a `42`, a picklist code; those aren't distinguishable from ordinary prompt text;
- a value reformatted between the row and the prompt (a date re-rendered, a number re-grouped);
- rows in the result set that are not currently on screen;
- **the `attributes` envelope, which `collectRecordValues` skips whole** — so `attributes.url` is never scanned, and it embeds the record's own Id (`/services/data/v62.0/sobjects/Account/001000000000001AAA`). Found by the re-verify and confirmed at this head: `attributes.url leak detected? null`. Same severity class as the three above — it only matters once the structural guarantee is already broken, and there is no route by which such a URL reaches the prompt — but the list is presented as complete, so it belongs on it.

It is a fail-closed net under the structural guarantee, not the guarantee.

Two limits were **missing from this list** on the first pass, and review found both. Neither is disclosed — both are fixed:

- ~~a short description switched the gate off entirely~~ → **fixed (N1)**, the request is subtracted by index. This was the bigger of the two and it was not on the list at all.
- ~~anything nested deeper than three levels was invisible~~ → **fixed (N2)**, the walk has no depth cap.

And one real limit that belongs on the list and wasn't there (N9): **`label` and `inlineHelpText` travel wholesale.** That is deliberate — they are what make a schema legible to a model — but it means admin-authored text is part of the egress. Markdown-table breakout is prevented (pipes escaped, newlines collapsed, verified by review), and the residual prompt-injection surface is bounded three ways: the CLI-side anti-injection preamble, SELECT-only validation of the reply, and never auto-running the result.

### AC-1 (hard) — zero direct LLM endpoints

Kept by **adding no transport**. Generation reuses the plumbing `features/ai-assistant.ts` already had: `createBridgeClient(...).call({ kind: 'ai', prompt }, { timeoutMs: LONG_RUNNING_TIMEOUT_MS })`. The provider, the API key, the redaction pass and the anti-injection preamble all stay CLI-side (`src/lib/bridge/routes.js` → `src/lib/ai.js`). **No new runtime dependency. No new manifest permission. No new host.**

New guard `extension/test/no-llm-endpoints.test.ts`:
- bans 13 model-provider hostnames across every shipped `.ts` in `features/ ui/ lib/ entrypoints/`;
- bans key-shaped credentials (`ANTHROPIC_API_KEY`, `x-api-key`, `apiKey:`, `sk-…`, …);
- forbids `fetch(` / `XMLHttpRequest` / `WebSocket` / `EventSource` **inside both AI features**;
- asserts the positive half — the call is the contract's existing `ai` kind;
- asserts `host_permissions` in `wxt.config.ts` still matches only Salesforce + localhost;
- has a floor assertion (`files.length > 80`) so a broken glob can't pass vacuously, and excludes only itself, by name with a reason (golden principle #12).

Verified against the emitted bundle too, not just source:
```
$ grep -rEo 'api\.(anthropic|openai|cohere|mistral|groq|deepseek|perplexity)\.(com|ai)|generativelanguage\.googleapis\.com|openrouter\.ai' .output/chrome-mv3/
(no matches)
```
and the built `manifest.json` `permissions`/`host_permissions` are byte-identical to `develop` (`wxt.config.ts` is untouched in the diff) — independently reproduced by review.

### AC-3 — gate + unavailable state

Own registry feature id **`soql-nl-generate`**, `enabledByDefault: false`, following C-P4-2's not-built-at-all posture: when off, *neither the toolbar button nor the panel is constructed* — absent, not hidden.

> **Interpretation flag.** AC-3 says "hidden when `features.ai` is off". There is no `features.ai` key in the extension — `settings.features` is keyed by feature id — so, per the C-P4-2 precedent in the brief, the AI gate for this control is its own registry id. `features.ai` *does* exist, but on the **CLI** side, in the project's `.sfdt/config.json`, enforced bridge-side: when it is false the bridge answers `REQUEST_INVALID` with *"AI features are disabled for this project. Set \"features.ai\": true…"*, and that is routed to the unavailable state below. **Review accepted this and recommended amending the AC text rather than the code**, on the grounds that the literal reading is self-contradictory with AC-3's other half: to hide the control you would have to round-trip the bridge before deciding whether to build it, leaving nowhere to render the unavailable state that the same AC demands. Review also confirmed the extension's `ai-assistant` toggle should **not** additionally gate this — an invisible cross-feature dependency is a "the button isn't there and I can't work out why" bug report waiting to happen.

Gate verified against the emitted bundle, per the C-FIX-2 warning that a dead lookup and "resolves enabled" look identical:
```
$ grep -o '"soql-nl-generate"[^}]*' .output/chrome-mv3/content-scripts/content.js
"soql-nl-generate","name":"Generate SOQL from a description (AI)",…,"enabledByDefault":false,…,"bridgeRequired":true

$ grep -o 'soql-nl-generate' .output/chrome-mv3/content-scripts/content.js | wc -l
2      # the manifest seed + the runner's isFeatureEnabled() literal — same string, live lookup
```
and at runtime by a test that calls `createSoqlRunnerFeature()` with **no `patchSettings` at all** (the genuine out-of-the-box state) and asserts the trigger is undefined, plus `feature-manifests.test.ts`'s `SHIPS_OFF_BY_DESIGN` allowlist, which I extended with a reviewed reason.

**Unavailable state.** `BRIDGE_OFFLINE` / `BRIDGE_UNAUTHORIZED` / `BRIDGE_FORBIDDEN` / `REQUEST_INVALID` are treated as "not set up yet": the panel keeps the bridge's own diagnosis and appends our steps (`sfdt ui` → pair on the options page → `"features.ai": true`) as a separate node via `renderSfError`'s `guidance` channel. A plain provider failure (a rate limit, `INTERNAL_ERROR`) deliberately gets **no** setup advice — asserted — because telling someone to configure what they already configured is worse than silence.

### Why the default is OFF

Bulk delete ships off because it destroys data. This ships off for a different reason: it **moves org schema out of the browser** — object and field API names, labels, types and inline help text — through the bridge to whatever provider the CLI is configured with. That is an egress decision that belongs to the user. It is also inert without a running bridge, so shipping it on would mostly manufacture an "unavailable" control for everyone. `soql-bulk-delete.ts`'s "the one feature that does" comment is corrected, and `SHIPS_OFF_BY_DESIGN` (the list of record) now has two entries.

---

## Reuse, not reinvention

- Error rendering: `renderSfError`/`setSfError`/`clearSfError`, always given the **error object** (`showNlError(error: unknown, guidance?)`), never `err.message` — so #327's rebase should be a no-op and the guidance lines survive as separate nodes. This file builds no `.sfdt-console`+`.sfdt-error` pair; `sf-error-panel-contract.test.ts` is green.
- Schema format: `buildSchemaMarkdown()` from P2-1's `export-for-prompt`, not a second serialiser.
- Local validation: mirrors `validateLocal()` in `src/lib/soql-runner.js` (D-1's `sfdt soql validate`), the same way `isSoslQuery` already mirrors its FIND rule. **I deliberately did not auto-fire Explain/plan after generation** — no follow-on org call is what makes "it never runs anything" cheap to keep true, and Explain is one click away once the query is in the editor.
- Describe data: the shared `DescribeCache` (usually already warm from autocomplete), wrapped in a small `awaitDescribe()` promise adapter with a 20s floor so a stuck describe can't leave the button spinning.
- Controls: `button()`, `setLabel()`, `.sfdt-field.sfdt-tall`, `.sfdt-frame`, `.sfdt-stack`, `.sfdt-note`. No second toolbar factory; `createResultActions()` is untouched.

## A11y checklist (CONVENTIONS.md), item by item

| # | Item | Status |
|---|---|---|
| 1 | Esc closes | **MET** — capture-phase `document` listener, removed on close *and* on view close. It claims Escape **only when focus is inside the panel** (via `composedPath()`, per item 13); with focus in the editor Escape still closes the runner exactly as before — asserted both ways. |
| 2 | Click-outside | **N/A** — a persistent surface holding unsaved user input (your description). Per item 2 it must *not* click-outside-dismiss. |
| 3 | Focus trap | **N/A** — an inline disclosure inside the runner, not a modal. It does not overlay anything; Tab out to the editor is correct. |
| 4 | Focus restore | **MET** — Esc and **Close** both return focus to the trigger, asserted with `activeElement`. **The trigger is never disabled**, and `closeNlPanel` re-checks `!disabled` before `.focus()` — the #326 bug, guarded rather than commented. On the SOSL path the panel is closed *before* the trigger is disabled, and a test asserts focus did not end up on it. On success focus goes deliberately to the **editor** (the thing you now have to read), not to the trigger. |
| 5 | Dialog semantics | **N/A** — not a dialog. `role="group"` + `aria-label`, with `aria-controls`/`aria-expanded` on the trigger, and **no `aria-haspopup`**: it is defined as "opens a menu" and this opens a form, so promising one and delivering the other is worse than promising nothing (N5, asserted). |
| 6 | Close affordance | **MET** — a visible, focusable **Close** button in addition to Esc. |
| 7 | Pin affordance | **N/A** — lives inside the runner's own view and dies with it. |
| 8 | Full keyboard path | **MET** — native `textarea` / `input` / `button` elements; focus moves to the description box on open; Enter submits from the Objects box, Ctrl/Cmd+Enter from the description box (matching the editor); Esc cancels. |
| 9 | Native controls | **MET** — no click-wired `div`s wired to click. |
| 10 | Every control labelled | **MET** — `label for=…` + `aria-label` on both inputs, `ariaLabel` on both buttons, `aria-describedby` pointing the description box at the disclosure text. Asserted. |
| 11 | State exposed | **MET** — `aria-expanded` on the trigger; `role="status"`/`aria-live="polite"` progress line; errors carry `role="alert"` via the shared funnel. |
| 12 | Zero innerHTML | **MET** — `createElement` + `textContent` only. |
| 13 | Shadow-root mounting | **MET** — renders inside the runner's `presentView` body, already in the shared root; the Escape check uses `composedPath()` for exactly that reason. |
| 14–15 | Tokens, both themes | **MET** — the panel's only inline style is `display`/`padding`/`margin`; every colour comes from `.sfdt-frame` / `.sfdt-field` / `.sfdt-note` / `.sfdt-muted`. A test walks every node in the panel and fails on any `#hex` or `rgb()`. |

Telemetry: local-only, `recordActivity` records the **object names**, never your free text. Secrets: no new SID path; `sid-never-leaves-worker.test.ts` green.

---

## Gates — actual output

Baseline established on `origin/develop` (cbffd36) in the same worktree, then head measured at `666b5fd`, so nothing below is a pre-existing failure. The `b0eccc9` column is what review reproduced exactly; the `666b5fd` column is the same run after the five re-verify follow-ups (+6 tests: one for F1, five for F2).

| Gate | `origin/develop` (cbffd36) | `b0eccc9` | head (`666b5fd`) |
|---|---|---|---|
| `extension` `npx vitest run` | 111 files / 1983 tests, =0 | 114 / 2089, =0 | **114 files / 2095 tests, =0** |
| `extension` `npx tsc --noEmit` | 0 | 0 | 0 |
| `extension` `npx eslint .` | 0 | 0 | 0 |
| `extension` `npx wxt build` | 0 (1.38 MB) | 0 (1.4 MB) | 0 (1.4 MB) |
| root `npm test` | 315 files / 5339 tests, =0 | 318 / 5445, =0 | **318 files / 5451 tests, =0** |
| root `npm run check:all-contracts` | 0 | 0 | 0 |
| root `npm run generate:catalogs` | — | no drift | no drift (`git status` clean after) |

**From `extension/` at head (`666b5fd`):**

```
$ npx vitest run
 Test Files  114 passed (114)
      Tests  2095 passed (2095)
   Duration  32.45s
vitest=0

$ npx tsc --noEmit
tsc=0

$ npx eslint .
eslint=0          (no output)

$ npx wxt build
Σ Total size: 1.4 MB
√ Finished in 1.628 s
wxt=0
```

**From the repo root at head (`666b5fd`):**

```
$ npm run generate:catalogs
catalogs=0        → no diff; generated/chrome-features.json + summary.json already current

$ npm test
 Test Files  318 passed (318)
      Tests  5451 passed (5451)
npm test=0

$ npm run check:all-contracts
contracts=0
```

`FEATURES.json` untouched. `generated/` regenerated, never hand-edited.

---

## What I could NOT verify

- **No live org.** Every describe and query is a stub. Object inference against a real global describe (managed-package prefixes, a 900-object org), the 20s describe timeout, and describe permission errors are untested against reality.
- **No running bridge, and no model was ever called.** Every bridge reply in the tests is a fixture. **Whether the prompt actually produces good SOQL is entirely unverified** — the prompt text is asserted for *content*, never for *quality*. Someone should run it against a real `sfdt ui` before this is called finished.
- **The CLI's `ai` route was read, not exercised.** The `REQUEST_INVALID` mapping for `"features.ai": false` is derived from `src/lib/bridge/routes.js`; I did not run the CLI with AI disabled to confirm the code on the wire.
- **Not driven in a loaded unpacked build.** `.output/chrome-mv3` builds and I verified the gate and the endpoint scan against the emitted bundle, but I never clicked the panel in Chrome. Visual density inside the runner modal, and the panel's behaviour in the docked side panel at narrow widths, are unconfirmed.
- **Both themes** are argued from tokens + an automated no-raw-colour assertion, not from looking at it in dark mode.
- **Provider-host list is not exhaustive, and nothing enforces egress at runtime.** `no-llm-endpoints.test.ts` bans 13 known hosts; a novel provider host passes it. The `fetch`/XHR ban is the broader net but it is scoped to the two AI features — review demonstrated that the identical `fetch` added to a non-AI feature (`features/inspect-record.ts`) passes `vitest`, `tsc` and `eslint`.
  **An earlier version of this section claimed the manifest's `host_permissions` allowlist was "the real backstop". That claim was wrong and is withdrawn.** The built manifest declares no `content_security_policy`, so there is no `connect-src` restriction, and a `fetch` to a cooperating host returning `Access-Control-Allow-Origin: *` succeeds from a content script or the worker regardless of `host_permissions` — that list governs credentialed / CORS-bypassing access to protected origins, not egress to an endpoint that wants the data. `host_permissions` is a real backstop against *reading* origins you have no permission for; it is not one against a deliberately-added exfil call. What actually guards this today is a source-level lint over two files, plus review. Closing it properly means a scoped `no-restricted-globals` fetch ban over `features/**` + `ui/**`, or a `connect-src` CSP — a **pre-existing gap this PR does not introduce**, and deliberately does not widen its scope to fix. **The retraction now also lives in `test/no-llm-endpoints.test.ts`'s header**, where the wrong claim had survived the first correction (see the re-verify follow-ups).
- **The `soql-bulk-delete` catalog row** rides along in the docs-site sync (it merged upstream before this and had not been synced), so the site's feature count moves 44 → 46 rather than 45 → 46. Review caught that the docs PR's new off-by-default table then read as exhaustive while omitting that row; it is added in sfdt-site#48.

---


scoobydrew83 (commented):
## Verdict: **APPROVE** — nothing blocking found

_Filed as a COMMENT because GitHub refuses `REQUEST_CHANGES` on a same-account PR. Read the first line as the verdict._

I did not write this code and I did not take the PR body's word for anything. I built my own worktree off `develop`, established a baseline, measured the head, and then spent most of the time trying to break the two safety properties by mutation rather than by reading. **16 mutations, 15 caught, 1 not caught** (and the one that isn't is a pre-existing scope gap, not a regression). The claims in the body hold up under attack better than most PRs I've reviewed; the notes below are all non-blocking, and several of them are corrections to the *honest-limits list* rather than to the code.

---

## Gates — my numbers vs yours

Own worktree (`.claude/worktrees/review-328`), one install, `wxt prepare` + a `packages/flow-core` build first (the suite does not run without `flow-core/dist` — worth a line in CONTRIBUTING, it cost me two red runs that were not the PR's fault).

| Gate | `origin/develop` (cbffd36) | `b5a3e90` | You claimed |
|---|---|---|---|
| `extension` `npx vitest run` | **111 files / 1983 tests, =0** | **114 files / 2070 tests, =0** | 111/1983 → 114/2070 ✅ |
| `extension` `npx tsc --noEmit` | 0 | 0 | 0 ✅ |
| `extension` `npx eslint .` | 0 | 0 | 0 ✅ |
| `extension` `npx wxt build` | 0 (1.38 MB) | 0 (1.4 MB) | 0 ✅ |
| root `npm test` | 315 / 5339, =0 | **318 / 5426, =0** | 318/5426 ✅ |
| root `npm run check:all-contracts` | 0 | **0** | 0 ✅ |
| root `npm run generate:catalogs` | — | **no drift** (`git status` empty after) | ✅ |

Every number in your gates section reproduces exactly. Contract lints match line for line (11 catalogs, 7 manifests, 53 auth docs, 118 paths).

**Dependency / permission diff (AC-1's third leg):**
```
$ git diff --stat cbffd36 b5a3e90 -- package.json package-lock.json extension/package.json extension/wxt.config.ts
(empty)
$ diff <(jq -S . manifest-develop.json) <(jq -S . manifest-head.json)
MANIFESTS IDENTICAL
```
Both manifests built by me, not copied from the PR. No new dependency, no new permission, no new host. Confirmed.

---

# P1 — AC-2: a generated query cannot auto-run

### Mutation results

| # | Mutation | Result |
|---|---|---|
| **M1** | Add `run?(soql)` to `GenerateDeps`, call it before returning `generated` | **CAUGHT** |
| **M2** | Wire that `run` in the runner to `runQuery(api, …)` | **CAUGHT** ×2 |
| **M4** | Neuter the Run button (`runBtn` click does nothing) | **CAUGHT** ×2 |
| **M16** | `invalid` case `break`s instead of `return`s, so the bad query is pasted | **CAUGHT** |

**M1** — the Proxy test does exactly what you say, and it *names the offender*:
```
AssertionError: expected [ 'askAi', 'describeObject', …(3) ] to deeply equal [ …(2) ]
  [ "askAi", "describeObject", "inspectPrompt", "knownObjects",
+   "run",
  ]
```

**M2 — "delete or weaken the Proxy assertion; does anything else catch it?" Yes.** With the executor actually wired at the call site, the Proxy test is redundant:
```
× puts the query in the editor and makes ZERO org calls doing it
  AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
× is not vacuous: the SAME query runs the moment the user presses Run
```
So AC-2 has two independent nets, and the second one does not depend on the first.

**M4 — the non-vacuity claim is genuine, and stronger than you argued.** Neutering Run fails *two* tests, not one:
```
× is not vacuous: the SAME query runs the moment the user presses Run
  AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times
× carries NO value from the result set that is on screen
  AssertionError: expected 'SOQL Query Runner…' to contain 'Zenith Prosthetics Consortium'
```
The AC-4 end-to-end test's precondition (`expect(document.body.textContent).toContain('Zenith…')`) is load-bearing — it cannot pass unless the harness really did run a query and really did paint rows. That is a better non-vacuity anchor than the one you claimed.

### Path-from-string-to-execution hunt

I went looking for the shape you'd expect to find and did not find it:

- **`editor.setValue` does not dispatch an event.** `lib/code-editor.ts:233` assigns `input.value = value` then `paint()/sync()` — no synthetic `input`/`change`. The runner's `textarea.addEventListener('input', …)` therefore never fires from a generation, so the `syncLangFromText` + `runAutocomplete` on that listener are dead for this path.
- **The success path calls `runAutocomplete()` explicitly**, which *can* reach `api.query` — but only inside `if (ctrlSpace)` (`soql-runner.ts:2809–2832`), and the success path calls the zero-arg form (`ctrlSpace = false`). No org call. The suite's `for (const call of apiGet.mock.calls) expect(...).toMatch(/describe/)` covers the describe traffic it *does* make.
- Checked and cleared: `exportAllBtn` (its `runQuery` is user-initiated), history/saved-query restore, `loadMore`, `explain`. None is reachable from generation.
- Escape-listener leak after view close: **none** (probe E3 — a second `Escape` after the runner closes has `defaultPrevented === false`, so no stale capture handler survives).

### Invalid-SOQL path

Reports, does not paste, and the draft survives — and M16 proves the test bites. `validateGeneratedSoql` also rejects `DELETE …`, `UPDATE …` and `SELECT …; DROP …` (I ran all three), while still accepting a subquery (`SELECT Id, (SELECT Id FROM Contacts) FROM Account`), so it isn't over-tight either.

**P1 verdict: the claim holds, and it holds structurally.** ✅

---

# P2 — AC-4: record data cannot reach the prompt

### I tried to get a record value in. I could not.

I wrote my own probe suite against the allowlist (deleted afterwards; tree is clean):

| Probe | Attack | Result |
|---|---|---|
| A1 | `picklistValues`, `defaultValue`, `compoundFieldName`, `referenceTo`, `records:[…]` on the describe | **all dropped**; the `Industry` row still renders |
| A2 | A `toJSON` on the describe field returning `{leak: …}` | **dropped** — output keys are exactly the five |
| A6 | `../../../etc/passwd`, `Account/describe?x=1`, `Account'--` as object names | **all filtered** by `SOBJECT_NAME_RE` before they reach a describe URL |

`buildSchemaMarkdown()` (`export-for-prompt.ts:68`) reads only `name/label/type/nillable/inlineHelpText` off the field objects it is handed, and it is handed already-rebuilt `PromptField`s — so it cannot be the wholesale-copy hole either. **The structural guarantee is real: `buildGeneratePrompt` has no records parameter, `schemaFieldsForPrompt` rebuilds, and nothing downstream widens it.**

### Gate mutations

| # | Mutation | Result |
|---|---|---|
| **M5** | `recordValueLeak` returns `null` unconditionally | **CAUGHT** (2 tests) |
| **M6** | `recordValueLeak` throws unconditionally | **CAUGHT** (20 tests) |
| **M7** | Remove the `try/catch` around `deps.inspectPrompt` | **CAUGHT** |
| **M8** | Observe the gate result but drop the `return { status: 'blocked' }` | **CAUGHT** (2 tests) |

**"A gate that throws is a refusal" — proven at the call site, not just asserted.** Under M6 I ran the wiring test in isolation:
```
$ npx vitest run test/soql-runner-nl-generate.test.ts -t "routes the prompt through"
AssertionError: expected [] to have a length of 1 but got +0
```
`calls` is empty. Nothing reached the bridge. A throwing gate is a refusal in fact, not by convention.

**`expect(d.askAi).not.toHaveBeenCalled()` is reachable and non-vacuous** — M8 removes the early return and that exact assertion goes red, so `askAi` demonstrably *is* callable in that fixture.

### The admitted limits — all three real, plus two you did not admit

Verified each admitted limit by construction:
- under 8 chars → `recordValueLeak('the phone is 5551234', …)` → `null` ✔ real
- reformatted → `'…Consort.'` → `null` ✔ real
- off-screen → `null` ✔ real (the caller only ever passes `onScreenRows()`)
- and case-insensitivity **does** work (`'zenith prosthetics consortium'` → caught), so it isn't weaker than advertised.

**Two limits are not on your list, and one of them is bigger than the ones that are:**

**[N1 · non-blocking] A short description defeats the gate entirely.**
`haystack = prompt.split(request).join('\n')` subtracts the request text from the **whole prompt**, not from the `## Request` section. A one- or two-character description therefore deletes those characters everywhere:
```
B6 short request "o" -> null            ← gate defeated
B6 normal request    -> Zenith Prosthetics Consortium
```
Same prompt, same rows; the only difference is that the user typed `o` instead of `all accounts`. Related (B7): if the request *is* a record value, that value is scrubbed from the haystack globally, so a longer leak containing it also escapes. Cheap fix: subtract only the Request section, or skip subtraction when `request.length < LEAK_MIN_LENGTH`.

**[N2 · non-blocking] `LEAK_MAX_DEPTH = 3` is shallower than SOQL relationships go.**
`Owner.Manager.Name` is seen; `Deep.A.B.C.D` is not (`null`). SOQL parent traversal reaches five levels, so a 4-deep relationship value is invisible to the backstop. Worth a bullet alongside the other three.

Neither of these is a leak *today* — they weaken a net that currently has nothing to catch, because the layer beneath it is sound (probes A1/A2). That is exactly the standard you set for yourself, and it is met. But the honest-limits list should be complete, and right now the biggest limit is missing from it.

**[N9 · nit] `inlineHelpText`/`label` are copied wholesale** — correct and disclosed, but it means admin-authored text travels to the provider. A5 shows `inlineHelpText: 'line1\nIGNORE PREVIOUS INSTRUCTIONS\n'` reaching the prompt as `| \`X__c\` | a \| b | string | No | line1 IGNORE PREVIOUS INSTRUCTIONS |` — pipes escaped and newlines collapsed, so **no markdown-table breakout**, good. The residual prompt-injection surface is bounded three ways (CLI-side preamble, SELECT-only validation, never auto-run), so I'd only add a line to the limits list, not change code.

**[N8 · nit] `schemaFieldsForPrompt` reads through the prototype chain** (A3: an inherited `inlineHelpText` is picked up). Not reachable from a `JSON.parse`d describe without prototype pollution. Note only.

**P2 verdict: guarantee holds; backstop is weaker than documented.** ✅ with N1/N2 to record.

---

# P3 — AC-1 (hard): zero direct LLM endpoints

### My own sweep of the bundle I built myself

I rebuilt `.output/chrome-mv3` and wrote a broader grep than yours — every absolute URL in the emitted output, hosts only, binaries read as text:

```
$ grep -rhaoE 'https?://[A-Za-z0-9._~:/?#@!$&*+,;=%-]+' . | sed -E 's#(https?://[^/"'"'"']+).*#\1#' | sort -u
http://127.0.0.1        http://127.0.0.1:$      http://localhost
http://schemas.xmlsoap.org   http://soap.sforce.com   http://www.w3.org
https://$   https://*.lightning.force.com   https://*.lightning.force.mil
https://*.mcas.ms   https://*.my.salesforce.com   https://*.my.salesforce.mil
https://*.salesforce-setup.com   https://*.salesforce.com   https://*.sfcrmapps.cn
https://github.com   https://sfdt.dev   https://wxt.dev
```
```
$ grep -rniaE 'anthropic|openai|\bgpt-[0-9]|claude-[0-9]|gemini|generativelanguage|cohere\.|mistral|groq|deepseek|perplexity|openrouter|together\.ai|replicate\.com|huggingface|bedrock|azure.*openai|vertexai|aiplatform\.googleapis|x\.ai|ollama|localai|llama' . -o
(no output)
```
Credential sweep (`sk-…`, `x-api-key`, `api_key`, `ANTHROPIC`, `OPENAI_`) turns up only the Salesforce `Authorization: Bearer` plumbing. **Clean.**

### Vacuity of `no-llm-endpoints.test.ts`

- **Glob resolves:** 125 `.ts` files across `features/ui/lib/entrypoints`.
- **The `> 80` floor does work.** M9 (narrow `SCANNED_DIRS` to `['features']`): `AssertionError: expected 46 to be greater than 80` → **CAUGHT**. Even dropping three of four directories trips it.
- **Golden principle #12: satisfied.** `DEFINING_ARTIFACTS` holds exactly one entry, the checker itself, with a reason, and explicitly explains why `lib/sfdt-bridge.ts` is *not* excluded. Correct.
  **[N6 · nit]** The header comment says "this file, **and the docs that describe the rule**, are excluded from their own scan by name and with a reason." The docs aren't excluded — they're simply outside `SCANNED_DIRS`. Trim the clause so the comment matches `DEFINING_ARTIFACTS`.

### Defeating the guard

| # | Mutation | Result |
|---|---|---|
| **M10** | `fetch('https://api.brand-new-llm-vendor.example/…', {headers:{'X-Auth':'sk-…'}})` inside `soql-nl-generate.ts` | **CAUGHT** (fetch ban *and* the `sk-` key pattern) |
| **M11** | Same fetch inside `features/inspect-record.ts` (non-AI) | **NOT CAUGHT** |

M11 in full: `npx vitest run` → **2085 passed**, `npx eslint .` → **0**, `npx tsc --noEmit` → **0**. A novel-host exfil endpoint added anywhere outside the two named AI features passes every gate in the repo.

**[N3 · non-blocking] The fallback claim is overstated.** You wrote that "the manifest's `host_permissions` allowlist is the real backstop." It isn't, for this threat: the built manifest declares no `content_security_policy`, so there is no `connect-src` restriction, and a `fetch` to a host that returns `Access-Control-Allow-Origin: *` succeeds from a content script or the worker regardless of `host_permissions` — that list governs *credentialed / CORS-bypassing* access, not egress to a cooperating endpoint. `host_permissions` is a real backstop against reading protected origins; it is not one against a deliberately-added exfil call.

This is a **pre-existing scope gap, not a regression** — this PR strictly improves the situation by adding the first such guard at all. Two options, both cheap: soften the sentence, or add `no-restricted-globals: ['fetch','XMLHttpRequest','WebSocket','EventSource']` scoped to `extension/features/**` + `extension/ui/**` in `eslint.config.js`, alongside the existing sid rules that already live there. I'd take the eslint rule; it turns M11 from NOT CAUGHT into CAUGHT for every feature at once.

**P3 verdict: AC-1 met and verified against the bundle.** ✅ with N3.

---

# P4 — AC-3: not-built-at-all, and the interpretation flag

### The C-FIX-2 dead-lookup trap

First I confirmed the trap is live: `UNKNOWN_FEATURE_DEFAULT = true` (`lib/feature-defaults.ts:59`), so a typo'd id resolves **enabled**. Then:

| # | Mutation | Result |
|---|---|---|
| **M12** | `isFeatureEnabled(settings, 'soql-nl-generatee')` | **CAUGHT** ×2 |

```
× builds no trigger and no panel at all for a user who has not opted in
× is off for a user with no stored preference at all
  AssertionError: expected <button …> to be undefined
```
**The lookup is live, not decorative.** Both the opted-out test and the out-of-the-box test go red — which also confirms the out-of-the-box test really is the genuine no-`patchSettings` state (I read it: `createSoqlRunnerFeature({ api })` with no settings patch at all ✔).

Traced in the emitted bundle independently of your grep — `soql-nl-generate` appears twice in `content.js`, and the second occurrence resolves through minification to a real call:
```
var hu = `soql-nl-generate`;          ← the exported const
… let h = Ot(t, hu), …                ← the live call
function Ot(e,t){ if(Object.prototype.hasOwnProperty.call(e.features,t)) … }   ← isFeatureEnabled
```

### Not-built vs built-and-hidden — I queried the DOM

OFF/ON census of the whole document after `onActivate()`:
```
OFF : {"textareas":1,"inputsText":0,"buttons":20,"roleGroup":1,"sfdtFrame":2,"ariaHaspopup":2,"hiddenEls":14}
ON  : {"textareas":2,"inputsText":1,"buttons":23,"roleGroup":2,"sfdtFrame":3,"ariaHaspopup":3,"hiddenEls":16}
OFF mentions "Generate": false | "sfdt-nl-panel": false | "sparkle": false
```
The panel id, the label, the icon and every one of its nodes are **absent from the DOM**, not `display:none`. Note the `hiddenEls` count also rises by 2 in the ON state — i.e. this is not a case of a hidden node already being present in the OFF baseline. Confirmed.

`SHIPS_OFF_BY_DESIGN` carries a reviewed, specific reason ("it moves data OUT of the browser… an egress the user did not ask for") that is genuinely different from bulk delete's, and the `soql-bulk-delete.ts` "the one feature that does" comment is correctly amended. `BRIDGE_REQUIRED` gains the id with a reason too.

### The interpretation flag — **I accept it, and I think the AC text is what's wrong**

You asked for a real decision. Here it is, with the reasoning.

I verified both halves of your premise:
- `grep -rn "features.ai" extension/ --include=*.ts` → **no hits** outside the two `ai-assistant` imports. The key genuinely does not exist extension-side.
- `src/lib/bridge/routes.js:504–509` → `if (!config.features?.ai) return makeErrorResponse(…, 'AI features are disabled for this project. Set "features.ai": true in .sfdt/config.json.', 'REQUEST_INVALID')`. It exists CLI-side and answers exactly the code and string your fixtures use.

So the AC as written asks the extension to hide a control based on a key in another process's config file. **The literal reading is not merely inconvenient, it is self-contradictory with AC-3's other half.** To "hide when `features.ai` is off" the extension would have to round-trip the bridge before deciding whether to *build* a button — meaning the control's existence depends on network latency, it vanishes when the bridge is merely slow, and there is no longer anywhere to render the "unavailable state that explains how to enable it" that the same AC demands. You cannot both hide the control and use it to explain how to turn it on.

What you shipped satisfies the intent strictly better: absent by default, and present-and-explanatory only once the user has asked for it and the other end says no. **Accept.** I'd amend the AC text rather than the code.

On the follow-up question you raised yourself — should the extension's `ai-assistant` toggle *also* gate this? **No, don't couple them.** `ai-assistant` is a Flow feature; a user who switched it off has said nothing about SOQL generation, and an invisible cross-feature dependency is exactly the kind of thing that produces a "the button isn't there and I can't work out why" bug report. Two independent, badged toggles is the more discoverable design. Your instinct was right.

**P4 verdict: met, and the reinterpretation is the correct call.** ✅

---

# Global DoD, golden principles, a11y, reuse

**Golden principles**
- **#2** — `generated/chrome-features.json` + `summary.json` regenerated; `npm run generate:catalogs` on my side produced **zero diff**. Not hand-edited. ✔
- **#4** — no secret in the diff; the bridge token is read from `settings.bridge.token`, the same path `ai-assistant` already used. No new secret path, `sid-never-leaves-worker.test.ts` green. ✔
- **#12** — see N6; substantively satisfied. ✔
- **FEATURES.json untouched** ✔ (not in `git diff --stat`).

**A11y checklist — I mutation-tested the MET claims you'd most want falsified**

| # | Mutation | Result |
|---|---|---|
| **M13** | Delete `if (!inside) return;` — panel steals Escape globally | **CAUGHT** ×2 (including "leaves Escape to the modal shell") |
| **M14** | `background: #ff00aa` into the panel's inline style | **CAUGHT** — the no-raw-colour walk bites |
| **M15** | Disable the trigger during generation (the #326 shape) | **CAUGHT** ×2 |

- **Item 1 / 13 — Escape scoping is done right, and for the right reason.** `present-view.ts:184–193` listens in the **bubble** phase specifically so a capture-phase child can claim Escape first (its own comment cites `ui/menu.ts`). The panel registers capture + `stopImmediatePropagation`, matching that contract exactly. Escape in the editor still closes the runner; Escape in the panel closes only the panel. Both directions asserted, both mutations caught.
- **Item 4 — focus restore.** Restores to the trigger, and the trigger is never disabled; `closeNlPanel` re-checks `!disabled` before `.focus()`. M15 confirms the guard is tested, not commented. The SOSL path closes the panel *before* disabling. ✔
- **Item 12 — zero `innerHTML`:** `grep -n "innerHTML|outerHTML|insertAdjacentHTML|createContextualFragment"` across both touched files → no hits. ✔
- **Item 2 N/A** — I checked the carve-out against `CONVENTIONS.md:43–48` verbatim: "*a persistent surface (… anything holding unsaved user input) must not click-outside-dismiss*." The panel holds the user's typed description. The N/A is correct, not convenient. ✔

**[N5 · nit]** `nlGenerateBtn.setAttribute('aria-haspopup', 'true')` — `aria-haspopup="true"` is defined as *menu*, but the target is `role="group"`. For a disclosure, `aria-expanded` + `aria-controls` alone is the correct pattern; `aria-haspopup` makes a screen reader promise a menu and then deliver a form. Not in your a11y table (item 5 lists the other three attributes). One-line deletion.

**Reuse claims**
- `renderSfError` always given the error object, never `err.message`: confirmed — `showNlError(error: unknown)`, and the call sites use `outcome.error ?? outcome.message`, preferring the throwable. `askAiViaBridge`'s catch keeps `error: err`. `sf-error-panel-contract.test.ts` **green** at this head, so #327's rebase should indeed be a no-op. ✔
- `buildSchemaMarkdown()` shared with P2-1 rather than a second serialiser: confirmed by import and by the byte-for-byte header assertion. ✔

**[N4 · non-blocking] A stale success path can overwrite the editor.** `runNlGenerate` checks `nlGenerateBtn?.disabled` at the *start* but not after the `await`. Probe E1, driven end to end: open the panel → Generate (bridge held open) → click **SOSL** → type `FIND {Acme} IN ALL FIELDS` → release the bridge.
```
E1 after SOSL: trigger disabled = true | panel display = none
E1 editor after generation completed: "SELECT Id FROM Account LIMIT 1"
E1 SOSL button pressed? false
```
The abandoned generation still lands, destroying the SOSL query the user typed and silently flipping the language back. No org call and no egress, so AC-2 is untouched — but it contradicts your own "an unusable answer must not overwrite the draft the user already had", for a *usable* one. A `if (!nlPanel || nlGenerateBtn?.disabled) return;` immediately before `editor.setValue` closes it, plus a test.

---

# Docs PR — `sfdt-site#48`

Claim-by-claim against what the code actually enforces:

| Claim | Verdict |
|---|---|
| "it is **never run for you**" | **Accurate as an absolute.** Verified structurally + by M1/M2/M4. |
| "does not send record data … a check refuses the send if a value from the results table currently on screen appears" | **Accurate as written.** Framed as the mechanism, correctly scoped to on-screen rows. |
| "the extension holds **no AI endpoint and no API key**" | **Accurate.** Verified against the bundle I built, with a broader regex than the PR's. |
| "adds no browser permission" | **Accurate.** Built manifests byte-identical. |
| "API name, label, type, required, inline help text, **and nothing else**" | **Accurate** — exactly `PROMPT_FIELD_PROPERTIES`, and probes A1/A2 confirm nothing else survives. |
| "capped at 150 fields, and the prompt says so" | **Accurate.** |
| "matched against the org's own sObject list — never invented" | **Accurate.** |
| `#generate-query-soql-nl-generate`, `/cli/commands/metadata#sfdt-soql` | **Both resolve** (`## \`sfdt soql\`` exists at `metadata.mdx:218`). |

**[N7 · non-blocking, other repo] The new off-by-default table in `settings.mdx` is presented as exhaustive and isn't.** It replaces "a feature with no explicit toggle is enabled by default" with "*a small number ship **off** … Those rows are badged Off by default on the options page:*" followed by a one-row table containing only `soql-nl-generate`. But the same PR syncs `soql-bulk-delete` — also `enabledByDefault: false` — into `data/upstream/chrome-features.json`, and:
```
$ grep -rn "soql-bulk-delete\|Bulk delete" content/
(no output)
```
The site documents neither that feature nor its off-by-default status anywhere. So a table that reads as the complete list of off-by-default features is missing half of it, in the same commit that adds the missing entry to the catalog. Add the `soql-bulk-delete` row (and ideally a Highlights section for it — but the table row is the part this PR made wrong). Given the prior overclaiming block on this repo, an *under*-complete table is the same category of problem.

Everything else in #48 is scrupulously hedged and I'd approve it once that row is added.

---

# Summary

**BLOCKING: none.** I looked hard: 16 mutations across all four priorities, my own probe suites against the allowlist and the leak scan, my own bundle build and endpoint sweep, my own DOM census of the OFF state, independent baseline and head measurement, and a read of the CLI-side route the AC interpretation rests on. The two safety properties are enforced by code shape, both nets are independent, and every net I tried to cut bled a test.

**Non-blocking, in the order I'd fix them:**
1. **N1** — short-request haystack gutting defeats the leak gate; subtract only the Request section (or skip subtraction below `LEAK_MIN_LENGTH`), and add it to the honest-limits list.
2. **N3** — soften the "`host_permissions` is the real backstop" sentence, or (better) add a scoped `no-restricted-globals` fetch ban for `features/**` + `ui/**`, which would have caught M11.
3. **N4** — re-check `nlGenerateBtn.disabled` after the await, before `editor.setValue`.
4. **N7** (docs repo) — add the `soql-bulk-delete` row to the off-by-default table.
5. **N2 / N9** — two more bullets on the limits list (nesting depth 3; admin-authored `label`/`inlineHelpText` travel).
6. **N5** — drop `aria-haspopup="true"`.
7. **N6** — trim the "and the docs" clause from the guard's header comment.
8. **N8** — note only.

None of these changes a merge decision. **Approve.**

---

## What I could not verify

- **No live org.** Every describe and query in my runs is a stub, same as yours. Object inference against a real global describe, the 20s `awaitDescribe` timeout, and describe permission errors remain untested against reality. *Not worked around.*
- **No running bridge and no real model.** Every bridge reply is a fixture. I confirmed the CLI-side `REQUEST_INVALID` mapping by **reading `src/lib/bridge/routes.js:504–509` and matching the literal string against your test fixtures** — so the code path and the message agree — but I did not run the CLI with `"features.ai": false` on the wire either. *Partially worked around; the string match is real, the wire is not.*
- **Prompt quality is entirely unverified.** I asserted content, never usefulness, exactly as you said. Nobody has established that this produces good SOQL.
- **Not driven in a loaded unpacked build.** I worked around this as far as static analysis allows: I rebuilt `.output/chrome-mv3` myself, ran my own endpoint/credential sweep over the emitted bytes, diffed the built manifest against a `develop` build I also produced, and traced the feature-gate literal through minification to the live `isFeatureEnabled` call. I did **not** click the panel in Chrome. Visual density inside the runner modal and behaviour in the docked side panel at narrow widths are unconfirmed.
- **Both themes.** Argued from tokens plus M14 proving the no-raw-colour walk bites. Not looked at in dark mode.
- **`sfdt-site#48`'s build.** I reviewed its diff and verified its anchors and claims against the code and the site's `content/`; I did not run `npm run build` in that repo.
- **The N3 CORS reasoning** is from the MV3 spec and the absence of a `content_security_policy` in the built manifest, not from an observed cross-origin request. The conclusion that M11 passes every repo gate *is* observed.

Worktree was `.claude/worktrees/review-328` (gitignored), nothing pushed, `FEATURES.json` never touched, all probe files deleted — `git status` clean in both the worktree and `/home/user/sfdt`.

---

scoobydrew83 (commented):
## Verdict: **CLEAR TO MERGE**

_Independent re-verification of the delta `b5a3e90..b0eccc9` only, not a re-review of the PR. Own worktree (`.claude/worktrees/rv328`, since removed), own install, own baseline. Nothing pushed, `FEATURES.json` untouched, both trees clean._

I went after the leak-gate indexing hard — 11 span mutations, 9 regex-shaped requests, 7 multibyte/normalisation cases, three fallback-poisoning constructions — and **could not defeat it**. I did find one fail-open, in the no-span fallback, and one real coverage gap in the stale guard where **3 of 4 clauses are untested**. Neither is blocking. Details below; I've kept the parts that are simply fine to one line each.

---

## Gates — every number reproduces exactly

| Gate | `origin/develop` (cbffd36) | `b0eccc9` | You claimed |
|---|---|---|---|
| `extension` `npx vitest run` | **111 / 1983, =0** | **114 / 2089, =0** | ✅ exact |
| `extension` `npx tsc --noEmit` | 0 | 0 | ✅ |
| `extension` `npx eslint .` | 0 | 0 | ✅ |
| `extension` `npx wxt build` | 0 (1.38 MB) | 0 (1.4 MB) | ✅ |
| root `npm test` | **315 / 5339, =0** | **318 / 5445, =0** | ✅ exact |
| root `npm run generate:catalogs` | — | no drift, `git status` empty after | ✅ |
| root `npm run check:all-contracts` | — | 0 | ✅ |
| `sfdt-site` `npm run build` (5bd2b40) | — | 64 pages / 3659 words, =0 | ✅ exact |

Manifest, both built by me:
```
$ diff <(json-normalise manifest-develop.json) <(json-normalise manifest-head.json)
MANIFESTS IDENTICAL
$ git diff --stat cbffd36 b0eccc9 -- package.json package-lock.json extension/package.json extension/wxt.config.ts
(empty)
```

---

# P1 — I tried to defeat the leak gate again. I could not.

**MU1 — revert `subtractRequest()` to `prompt.split(request).join('\n')`:** **CAUGHT**, and your seven-test table reproduces verbatim, plus three of my own probes.
```
× catches the leak with a one-character description
× catches the leak with a two-character description
× catches the leak with a description that is a word of the value
× catches the leak with a description at the gate's own 8-character floor
× subtracts the request ONCE — a second copy elsewhere is still evidence
× falls back to the Request heading when the caller gives no span
× ignores a span that does not match the prompt it was given
× P2 request is a PREFIX of the leaked value (B7 variant)
× P6 mutated spans: off-by-one, negative, reversed, past-end, wrong-place
× P9 fallback: heading literal inside a field inlineHelpText
 Tests  10 failed | 62 passed (72)
```

### My own probe suite — all CAUGHT

```
PROBE P1 request==schema-text ("Account Name", present twice)      -> CAUGHT
PROBE P2 prefix-of-value ("Zenith Prosthetics")                    -> CAUGHT
PROBE P3 request="$&"  "$'"  "$`"  "$1"  ".*"  "[a-z]+"  "\d{8}"
         "(?:o)"  "^$"                                             -> CAUGHT ×9
PROBE P4 span-ok=true  "👨‍👩‍👦 accounts" | "café"(NFC) | "café"(NFD)
         | "日本語のアカウント" | "𝔘𝔫𝔦𝔠𝔬𝔡𝔢" | ZWSP | NBSP        -> CAUGHT ×7
PROBE P5 untrimmed requestText against a trimmed span              -> CAUGHT (fails CLOSED)
PROBE P10 empty request                                            -> CAUGHT
```

`$&`/`$'`/`` $` `` have no surface at all because the fix uses `slice`, not `replace` — that's a genuine structural improvement over any regex-based subtraction, and worth noting deliberately.

**Why the multibyte cases can't line up wrong**, which I think is the strongest thing about this fix and isn't argued in the body: `requestRange` is `[head.length, head.length + request.length]` where `head = PROMPT_INSTRUCTIONS + '\n\n' + REQUEST_HEADING` — both **static**. The span therefore depends on neither the request's contents nor the schema's, so a malicious `inlineHelpText`, a 900-field describe, or an astral-plane request cannot move it. `prompt.slice(start,end) === request` cannot fail on the real path. Combined with `generateSoql()` trimming **once** and handing the same trimmed string to both `buildGeneratePrompt()` and `inspectPrompt()`, the trim-divergence attack has no purchase either.

### What happens when the equality check fails — it fails **closed**, in all eleven mutations

```
PROBE P6 off-by-one start-1            -> CAUGHT
PROBE P6 off-by-one end+1              -> CAUGHT
PROBE P6 off-by-one start+1            -> CAUGHT
PROBE P6 negative start                -> CAUGHT
PROBE P6 reversed                      -> CAUGHT
PROBE P6 past end                      -> CAUGHT
PROBE P6 zero span                     -> CAUGHT
PROBE P6 whole prompt                  -> CAUGHT
PROBE P6 NaN                           -> CAUGHT
PROBE P6 float                         -> CAUGHT
PROBE P6 wrong-place but slice-equal   -> CAUGHT
```
The `wrong-place but slice-equal` case is the only one that gets past validation, and it's harmless by construction: it removes **one** span of exactly `request.length`, so it can never take a whole leaked value with it. The old bug needed a *global* deletion; nothing in the new code can produce one.

### The one thing I found — **[F1 · non-blocking] the no-span fallback fails OPEN when the prompt has no `## Request\n`**

```
PROBE P7 no-heading, leak first + request "o"            -> null :: NOT CAUGHT
PROBE P7 no-heading, leak first + request "Zenith P"     -> null :: NOT CAUGHT
PROBE P7 no-heading, leak first + request "Prosthetics"  -> null :: NOT CAUGHT
```
With `heading < 0`, `from = 0`, so `prompt.indexOf(request, 0)` finds the first occurrence **anywhere** — and when that occurrence sits inside the leaked value, it breaks the value apart before the scan, exactly the N1 failure mode in miniature. The doc comment ("removes exactly one occurrence — the one under `REQUEST_HEADING` **when there is a heading** — never all of them") is literally true, and that carve-out is the whole hole.

Not reachable today, and strictly better than the pre-fix state: `soql-runner.ts` is the only caller and always passes `context.requestRange`, whose validation cannot fail (static `head`). But the fallback exists precisely for callers that didn't, and for those it is weaker than the comment reads. One line: `if (heading < 0) return prompt;` — subtract nothing rather than the wrong thing.

The **heading-present** attacks all hold:
```
PROBE P8 two "## Request" headings, no span                       -> CAUGHT
PROBE P9 REQUEST_HEADING literal inside a field's inlineHelpText  -> CAUGHT (with span AND via fallback)
         heading count in prompt = 2
```
P9 is the injected-heading attack: even with an admin-authored `## Request\n` in the schema half, `indexOf` finds the real one first (the schema is always after), so the poisoned copy is unreachable.

### B7 and the original AC-4 attacks

B7 closed — confirmed independently (P2, and their `subtracts the request ONCE` test). Original AC-4 attacks re-run and all still held: `picklistValues`, `defaultValue`, `referenceTo`, `compoundFieldName`, `calculatedFormula`, a `toJSON` returning the value, `records:[…]`, `sampleValue`, `lastRecordId` — `prompt contains leak? false`, `contains Industry row? true`. Five-hop and subquery-envelope nesting caught.

---

# P2 — the other three fixes

### Depth — non-vacuous, terminates, fails closed under stress

**MU2 (reinstate `LEAK_MAX_DEPTH = 3`): CAUGHT**, three shipped tests + two probes red — the 5-hop / subquery / 40-level tests are not decorative.
**MU3 (remove the cycle guard): CAUGHT** — `× terminates on a cyclic row rather than hanging the tab / RangeError: Maximum call stack size exceeded`. The cyclic test bites.

Termination, beyond your table:
```
PROBE P16 2^40-path DAG (40 levels, each node shared twice) -> CAUGHT in 0ms
PROBE P14 50,000-deep chain -> THREW RangeError: Maximum call stack size exceeded in 12ms
PROBE P15 same, through generateSoql -> outcome status = blocked
```
The `seen` set makes a shared-reference DAG bomb linear rather than exponential — an unintended but real win. A pathological depth blows the stack rather than hanging, and `generateSoql`'s try/catch turns that into `status: 'blocked'`: **fail-closed**, no hung tab, and unreachable from a `JSON.parse`d response anyway. Uncapped-with-a-cycle-guard is the right call.

### **[F2 · non-blocking] `nlGenerationStale()` — three of its four clauses are untested, and only one is load-bearing**

You asked me to assume the #327 shape. This is the closest thing to it I found. Dropping each clause **separately**:

| Clause dropped | Result |
|---|---|
| `viewAtDispatch !== view` | **32 passed — NOT CAUGHT** |
| `!nlPanel` | **32 passed — NOT CAUGHT** |
| `!nlPanelOpen` | **1 failed — CAUGHT** (`× does not overwrite the draft after the user closed the generator`) |
| `lang !== langAtDispatch` | **32 passed — NOT CAUGHT** |

And reducing the whole guard to a single clause:
```
=== guard reduced to ONLY !nlPanelOpen ===
      Tests  34 passed (34)          # incl. my two extra probes
=== guard reduced to ONLY viewAtDispatch !== view ===
     × S2 language flips by TYPING a FIND, panel untouched
     × does not overwrite the SOSL query typed while it was in flight
     × does not overwrite the draft after the user closed the generator
      Tests  3 failed | 31 passed (34)
```

**The behaviour is correct** — I built the two scenarios the untested clauses are supposed to own and drove them end to end against unmutated code:
```
PROBE S1 close-and-REOPEN mid-flight: new-panel-open=true editor="SELECT Id FROM Contact" toast=false
PROBE S2 lang flipped by TYPING a FIND: panel-open-after-typing=false editor="FIND {Acme} IN ALL FIELDS"
```
Both abandoned correctly. The reason the other clauses never fire is structural, and I checked it rather than inferred it: every route to `lang === 'sosl'` goes through `setLang()`, which calls `closeNlPanel({restoreFocus:false})` **first** — including the typing route (`syncLangFromText` → `setLang`), which is why S2 shows the panel already shut. And `presentView`'s `close()` always invokes `opts.onClose`, which calls `closeNlPanel()` — so close, and close-then-reopen, both clear the **dispatching closure's** `nlPanelOpen` (`runNlGenerate` closes over the previous `open()` invocation's locals, not the new one's).

So this is not a hole; it's belt-and-braces, and I'd keep all four. But the PR body sells "**view identity** (so close-and-reopen is caught, not just close)" as a distinguishing property of the fix, and nothing in the suite would notice if that clause vanished. If someone later stops `onClose` from calling `closeNlPanel()` — a plausible refactor — the guard silently degrades to nothing and the suite stays green. Either a test per clause, or a comment saying the other three are deliberately redundant today and why. Non-blocking.

Non-vacuity confirmed the other way too: the "user waits" test genuinely lands (`SELECT Id FROM Account LIMIT 1` in the editor), and the reduce-to-`!nlPanelOpen` run above shows the three guard tests aren't passing for a "never write" reason.

**On the judgement call — a manual Close abandoning a late success: I agree.** Close is an explicit dismissal, and a `SELECT …` materialising in the editor after you dismissed the thing that makes them is the same class of bug as N4, just with better manners. The cost is genuinely small: `closeNlPanel()` only sets `display:none`, so the typed description survives; reopening and pressing Generate is two clicks. The one thing I'd note is that a user who pressed Close *in order to look at the editor while waiting* gets no signal at all — the re-enabled trigger is the only tell. Acceptable as shipped; the alternative is worse.

### `aria-haspopup`

**MU7 (re-add it): CAUGHT.** **MU8 (delete `aria-expanded` instead): CAUGHT ×2** — so the test asserts the right thing, not merely an absence. `aria-controls` still pinned to `panel().id`.

---

# P3 — regression sweep on what was already cleared

- **AC-2.** **MU10** (add `run?(soql)` to `GenerateDeps`, call it before returning `generated`): **CAUGHT** — `expected [...] to deeply equal [...]  + "run"`. **MU11** (wire that `run` to `api.query` at the call site): **CAUGHT ×2** — `× puts the query in the editor and makes ZERO org calls doing it` / `× is not vacuous: the SAME query runs the moment the user presses Run`, both `expected "vi.fn()" to not be called at all, but actually been called 1 times`. The dep-name pin still holds and still names the offender; the second net is still independent. The stale rework adds only early `return`s — no new call path.
- **AC-3.** **MU9** (typo the id at the lookup): **CAUGHT ×2** — the opted-out test and the out-of-the-box test both go red, so the lookup is still live, not decorative. Emitted bundle: `"soql-nl-generate"…"enabledByDefault":false…"bridgeRequired":true`, 2 occurrences. Panel absent, not hidden.
- **AC-1.** Rebuilt `.output/chrome-mv3` myself. 20-pattern provider-host regex → **no output**. Credential sweep (`sk-…`, `x-api-key`, `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `api_key:`) → **no output**. Full absolute-URL host census identical to the first review's list (Salesforce + localhost + github.com/sfdt.dev/wxt.dev/w3.org/xmlsoap). Manifests identical, dependency diff empty.
- **M11 still open, and slightly narrower than reported.** With an `sk-…` literal it *is* caught by the key rule (`× holds no provider API key or key-shaped credential`). Without one — the realistic shape, a token read from storage — it is not:
  ```
  === M11b: novel-host fetch, no literal key, in features/inspect-record.ts ===
        Tests  18 passed (18)
  eslint=0   tsc=0
  ```
  Pre-existing, correctly disclosed, correctly out of scope.

---

# P4 — PR body and docs

- **N3 replacement text: accurate.** Verified independently rather than taking it — the built manifest's keys are `[action, background, commands, content_scripts, description, host_permissions, icons, incognito, manifest_version, name, options_ui, permissions, side_panel, version]`. **No `content_security_policy` key at all**, so no `connect-src`. "What actually guards this today is a source-level lint over two files, plus review" matches what M11b shows.
- **Honest-limits list: correctly reflects the fixed state.** N1 and N2 are struck through and marked fixed, not still described as live; N9 added. **One omission remains** (the bot's finding at this head, so outstanding rather than ignored, and I confirmed it): `collectRecordValues` skips the whole `attributes` object, and `attributes.url` embeds the record's own Id.
  ```
  PROBE P13 attributes.url leak detected? null
  ```
  `/services/data/v62.0/sobjects/Account/001000000000001AAA` is squarely "record data currently on screen", and it's a fourth blind spot alongside the three that are listed. Same severity class as the others — only matters once the structural guarantee is already broken — but the list is presented as complete. Either skip only `attributes.type`, or add the bullet.
- **`sfdt-site#48` (`511827c..5bd2b40`): accurate, and the table is now genuinely complete.** I checked the row's claims against code rather than against the docs: `feature-manifests.json` has **46 features and exactly two** with `enabledByDefault:false` — `soql-bulk-delete`, `soql-nl-generate` — so the two-row table is exhaustive as of this head. CSV-backup-first confirmed (`downloadDeleteBackup`, with `backupCsvCoversPlan` refusing a backup that doesn't cover the plan); the phrase is `` `DELETE ${plan.ids.length} ${plan.sobject}` `` at `soql-bulk-delete.ts:219`, so `DELETE <count> <Object>` is right. "It has no page on this site yet" is true — `grep -rn "soql-bulk-delete\|bulk-delete" content/` returns only the CLI MCP-tool row and this new row. The softened intro no longer reads as exhaustive: it names the options-page badge as the authority. Row present in the **rendered** `out/chrome-extension/settings.html`, not just the MDX. Build green.
- **N6: confirmed inaccurate — and there is a worse inaccuracy four lines above it in the same comment block.**
  `DEFINING_ARTIFACTS` holds exactly one entry (`test/no-llm-endpoints.test.ts`, with a reason). The docs are outside `SCANNED_DIRS`, not excluded. So "this file, **and the docs that describe the rule**, are excluded from their own scan by name and with a reason" overstates what the array does. On its own: a trivial comment fix, and **#12 is substantively satisfied** by the one honest entry.
  What makes it worth more than a nit is its neighbour. The same header still reads:
  > *"Adding an LLM endpoint to the extension would need a provider host, a key, and (because of the manifest's `host_permissions` allowlist) a new permission"*

  That is the **exact claim you withdrew from the PR body as wrong**. The retraction landed in the description and not in the source — `git diff --stat b5a3e90..b0eccc9 -- extension/test/no-llm-endpoints.test.ts` is empty. The PR body is read once; this comment ships and is the first thing the next person to touch AC-1 will read, and it tells them a backstop exists that doesn't. Fix both clauses in one edit. **Still not blocking** — it's a comment, and the withdrawal is on the record here.

---

## Summary

**BLOCKING: none.** Every fix in `b0eccc9` does what it claims, each is mutation-checked and each mutation I re-ran reproduced your reported result exactly. The N1 fix in particular is better than "a different subtraction": moving to a static-offset span makes the exclusion independent of both user input and schema content, which is why nine regex shapes, seven multibyte/normalisation cases and eleven span mutations all bounced off it.

Non-blocking, in the order I'd fix them:
1. **F2** — one test per `nlGenerationStale()` clause, or a comment that three are deliberately redundant. Today the guard is 3/4 untested and would degrade silently.
2. **N6+** — trim the "and the docs" clause **and** delete the withdrawn `host_permissions` sentence from `no-llm-endpoints.test.ts`'s header, in the same edit.
3. **`attributes.url`** — one more bullet on the honest-limits list, or skip only `attributes.type`.
4. **F1** — `if (heading < 0) return prompt;` in `subtractRequest`'s fallback.

Ship it.

---

## What I could not verify

- **No live org, no running bridge, no model called.** Every describe, query and bridge reply in my runs is a stub, same as both previous passes. The 20s `awaitDescribe` floor, object inference against a real global describe, and describe permission errors remain untested against reality. **Prompt quality is still entirely unverified** — I asserted content, never usefulness, and nobody has yet established this produces good SOQL.
- **Never clicked in a loaded unpacked build.** Worked around as far as static analysis goes — I rebuilt `.output/chrome-mv3` myself, swept the emitted bytes, diffed the built manifest against a `develop` build I also produced, and traced the gate literal in `content.js`. I did not open the panel in Chrome. Visual density in the runner modal and the docked side panel at narrow widths are unconfirmed.
- **Dark mode** — argued from tokens and the no-raw-colour walk, not looked at.
- **F1's unreachability** is argued from `soql-runner.ts` being the only caller of `recordValueLeak` in the tree today. A future caller assembling its own prompt is the case it applies to, and there is no test or type that stops one appearing.
- **The `attributes.url` finding** is confirmed as a scan blind spot; I did **not** construct a path by which such a URL reaches the prompt — the allowlist still has no route for it, which is why this stays a limits-list item and not a defect.
- **The N3 CORS reasoning** — I verified the manifest has no CSP directly, but the conclusion that a cooperating `Access-Control-Allow-Origin: *` host would receive the data is from the MV3 spec, not an observed request. M11b passing every repo gate *is* observed.
- **`sfdt-site#48`** — I built it and checked the rendered HTML for this table, but I did not read the whole site for other staleness.

Worktree `.claude/worktrees/rv328` (gitignored) plus a detached `sfdt-site` worktree, both removed; all probe files deleted; nothing pushed; `FEATURES.json` never touched. `git status` clean in `/home/user/sfdt` and `/home/user/sfdt-site`.

---

PR #329, opened 2026-08-05
fix(extension): the C-FIX-4 sweep can be blinded by one backtick, and cannot see `.catch((e) => …)`

scoobydrew83:
Round 5 of C-FIX-4. Closes the three blocking findings from the round-4 independent review of #327 (review `4869214853`, `REQUEST CHANGES`, 2026-08-05 22:08:50Z), plus four of that review's non-blocking findings.

**#327 was merged into `develop` at `d42b7c1` (22:43:31Z) with that review open, at the exact head — `1f802d2` — the review requested changes on.** So B1, B2 and B3 are live defects on `develop`, not proposed ones. This is the second time on this line of work: #325 merged at `102f5dd` the same way, which is what #327 existed to fix. Stated once, as fact.

Base for this PR is therefore `origin/develop` @ `d42b7c1`, which *includes* #327. Branch `ext/c-fix-4-round-5`, head `ed1a878`. `FEATURES.json` untouched. Not merged.

---

### B1 — one backtick could blind every rule, over an arbitrary span

`dynamicParts()` is the mask **both** guards read the tree through. It treated every backtick as a template delimiter, and — unlike its `'`/`"` branch two lines above — it was not newline-bounded. A backtick that opens no template can only come from a comment or a regex, and that scanner parsed neither. So one stray backtick started a mask running to the next backtick anywhere in the file.

Review measured the blast radius on `features/rest-explore.ts`: a backtick appended to the header comment at line 38 blanked **7,588 characters through to line 108** and took the file's only `.textContent =` with it. Rules 2 and 3 and the whole newlines guard went silent over seventy lines of a live feature file, with nothing failing. 93 of 125 scanned files carry backticks inside comments — 1,819 of them — balanced today by convention and nothing else. It is a regression against `4ae470d`, where rule 2 still matched raw source.

**The fix asks the compiler.** Bounding the backtick scan at a newline was not available: the CSS sheet and the Python template in `soql-runner.ts` are real multi-line templates, and stopping at a newline puts the phantom-binding class straight back. Handling comments needs a scanner that knows strings; handling a backtick inside a regex character class needs one that knows regex-from-division. Each of those hand-rolled is one more pattern that has to guess, which is this guard's entire failure history.

So `dynamicParts()` now runs `ts.createSourceFile()` and blanks the literal spans the parser reports. That is not an approximation of the grammar, it is the grammar, and it cannot be defeated by an odd backtick because it never has to decide what one means. Comments are blanked second, on the already-emptied buffer, where a `//` is unambiguous by construction — every string interior is spaces, every regex is gone, and a `//` in JavaScript is never two divisions. `typescript` is already an extension devDependency (it is what `npx tsc --noEmit` runs) and this module is test-only, so nothing reaches the shipped bundle.

The property review asked for is pinned tree-wide in a new suite, `test/error-source-scan.test.ts`: **injecting a backtick at a position the mask already calls static must not change what the mask calls code, in any of the 125 scanned files.** Plus offset/newline preservation and idempotence, both tree-wide.

### B2 — the catch scan did not see `.catch((e) => …)`

The binding scan matched `catch\s*\(\s*(name)`, and `\s*` does not match `(`. The statement form was read; the promise form was not — in the mechanism round 4 advertised as having replaced spelling with structure.

Not an exotic spelling: Prettier 3 defaults `arrowParens` to `"always"` and this repo overrides nothing, so the parenthesised form is the **only** shape `prettier --write` will ever emit for a single parameter. The tree already writes it at `lib/spa-router.ts:40` and `entrypoints/background.ts:469` — both spelled `err`, so the spelling fallback was quietly covering the gap. `async` and `function` handler heads are read through too.

### B3 — the exemption bite check did not check what it claimed

`every exemption would actually be flagged without it` asserted that the assignment still existed and that `rendersAnErrorValue(entry.expression)` was true. Neither is the property an exemption claims: it never asked whether the site would be **reported**, so it never ran `setsWhiteSpaceDirectly`, `setsWhiteSpaceByClass`, `fromPanelBuilder` or the `cssText` test. Any site already covered by one of those four sailed through — review added a decorative `ui/toast.ts` / `toast` / `message` entry and the suite stayed green, while `ui/toast.ts:78` carries `white-space: pre-line` and is provably not an offender.

The check now re-runs the **real** offender scan with the entry removed and requires the site to appear in the report — the shape the sibling `every exclusion would actually fail without it` on the sweep already had. A fixture test pins it against that exact decorative entry, and asserts both of the things the old check asserted are *true* of it, so the fixture cannot start passing for the wrong reason.

---

## Also closed from the same review

- **N5 — the empty-`Error('')` swallow.** Round 4's reply did not mention it; this closes it. `sfErrorParts()` falls back to `errorText()`, which is `''`, so `setSfError` found no parts and took the branch meant for `renderSfError(null)`. A caught `new Error('')` or a bare `throw new Error()` rendered as **no panel, no `role="alert"`, no text**, from every call site that forwards a caught error. That is #308 itself — a failure the user never sees — arriving through the null contract this helper introduced, and invisible to the sweep because every call site is *correct*. A thrown value always produces a panel now; `null`/`undefined` still render nothing, because a caller saying "there is no error" is a different statement from an error saying nothing. No call site relies on `''` to clear — the five clears all pass `null`.
- **N2 — the newlines guard asked only the spelling question.** Its entry point was `rendersAnErrorValue()`, which takes no names and never calls `errorBoundNames()`, so every structural gain of the shared module reached rules 2 and 3 and stopped at that file's door while the module header claimed one shared definition. It asks `carriesAnError()` with the file's own bindings now. **Measured tree-wide before changing it: zero new sites** — a claim made true rather than a widening.
- **N3 — `wordsOf()` did not split on a digit,** so `err2` was one word and not `err`. A `\d` boundary was added; `err2`, `error1`, `msg2` are pinned.
- **Rule 1 read comments as code.** Not in the review — found while measuring B1. Rule 1 is the one scan that *cannot* use `dynamicParts()`: that mask empties string literals, and a class name is a string literal. (Measured: rule 1 read through the full mask flips exactly one verdict — `ui/panels.ts`, the single file it must trip.) It kept reading raw source, so a commented-out hand-roll reported as a live one. `withoutComments()` is split out and rule 1 reads through that. Latent, not live, and it fired in the safe direction; measured over every scanned file, no rule-1 verdict changes.

---

## Mutation evidence

Every fix reverted in place, the named test observed red, then restored. Same format as rounds 2-4.

| # | mutation | test that goes red | result |
|---|---|---|---|
| B1 | `dynamicParts()` restored to `d42b7c1`'s character scanner | `cannot be blinded by one stray backtick, in any file in the tree` · `drops a comment rather than reading it as code` · `keeps the code that follows an unbalanced backtick` · `binds no phantom name from a comment or a template` · `a stray backtick in a comment cannot blind rule 2` | **5 failed / 32 passed** |
| B2 | `CATCH_BINDING` reverted to `\bcatch\s*\(\s*([A-Za-z_$][\w$]*)` | `rule 2 knows a catch binding in its PROMISE form too` · `rule 3 knows a catch binding in its PROMISE form too` | **2 failed / 26 passed** |
| B3 | review's decorative `ui/toast.ts` entry added to `EXEMPT` | `every exemption would actually be flagged without it` | **1 failed / 13 passed** |
| B3 | …same entry, with `d42b7c1`'s version of that check restored | *(none — it accepts it)* | **13 passed** — reproduces the review finding exactly |
| B3 | the one real `EXEMPT` entry deleted | `every element assigned an error message declares a white-space rule` | **1 failed / 13 passed** — the live entry is load-bearing |
| N5 | `setSfError`'s unconditional clear-on-empty restored | `still renders a panel for a thrown error with no message` | **1 failed / 27 passed** |
| N2 | newlines guard reverted to `rendersAnErrorValue(expression)` | `knows an error by its BINDING here too, not only by its spelling` | **1 failed / 13 passed** |
| N3 | the `\d` boundary removed from `wordsOf()` | `rule 2 knows an error by its binding, not only by its spelling` | **1 failed / 27 passed** |
| rule 1 | `classesByElement` reverted to raw source | `does not read a commented-out hand-roll as a live one` | **1 failed / 28 passed** |

---

## Constructed injection — re-run, not reasoned about

The review's method, against **two files no round has used**: `features/api-name-generator.ts` (a real `.sfdt-console` preview pane) and `features/debug-log-viewer.ts`. 33 probes spliced into the real source of a real file, driving the real exported matchers (`buildsSfErrorPanel`, `rendersErrorIntoConsole`, `passesStringifiedError`) plus a faithful replica of the newlines guard's per-file scan. Run on both sides in the same worktree, base measured by swapping in `d42b7c1`'s scanner.

|  | `d42b7c1` | `ed1a878` |
|---|---|---|
| 28 defect probes | **10 CAUGHT / 18 MISSED** | **28 CAUGHT / 0 MISSED** |
| 5 correct-code probes | 4 declined, **1 false positive** | **5 declined** |

The 18 base misses: all eight backtick-blinder shapes (line comment, three backticks, block comment, JSDoc, regex character class, `` don`t `` typo, defect 40 lines later, blinder + const hop), all nine `.catch((x) => …)` shapes (`(e)`, `(reason)`, `(splat)`, `async (e)`, `function (e)`, interpolated, single-letter `(q)`, numbered `(err2)`, and the rule-3 `renderSfError(e.message)` form), and the combined backtick-then-promise-catch shape. The base false positive was a **commented-out** defect reported as live — gone now.

Two probes were CAUGHT at base and are worth naming so the table is not read as better than it is: *two comments, one backtick each* (the two backticks balance, so the mask closes before the defect) and *`.catch((e) => …)` via a const hop* (`e instanceof Error ? …` is a flatten by construction and needs no binding). Both were already closed.

The regression probes rounds 2-4 closed — template interpolation, bare `${err}`, `catch (splat)`, five const hops, the `let` accumulator, the Prettier trailing-`+` wrap, `renderSfError(String(err))` — are all still CAUGHT.

---

## Gates

Both sides measured in one worktree, same install, `packages/flow-core` built once, `wxt prepare` re-run on each side.

| | `origin/develop` `d42b7c1` | head `ed1a878` |
|---|---|---|
| `extension/ npx vitest run` | 114 files / **2121** passed | 115 files / **2137** passed |
| `extension/ npx tsc --noEmit` | `exit=0` | `exit=0` |
| `extension/ npx eslint .` | `exit=0` | `exit=0` |
| `extension/ npx wxt build` | `Σ Total size: 1.4 MB` | `Σ Total size: 1.4 MB` |
| root `npm test` | 318 files / **5477** passed | 319 files / **5493** passed |
| root `npm run generate:catalogs` | no drift | no drift |

Built manifest `sha256` on both sides: `2a302f6a1b97539b0857f162e24fb498774b602f8840b515d865ecc9e8359c7d` — byte-identical, and the same hash rounds 3 and 4 measured.

Diff is 7 files, all under `extension/`, all test files plus `ui/panels.ts` and the changelog. `git diff d42b7c1..ed1a878 -- generated/ FEATURES.json package.json package-lock.json extension/package.json` is empty.

Prettier: `test/panels.test.ts` and `ui/panels.ts` are **not** Prettier-clean at base, so neither was run through `--write`. Verified instead that `prettier`'s complaint set on each is *identical* before and after this change — the added blocks introduce no new drift. The two base-clean files were formatted normally.

---

## Declined, with reasons

- **N1 — six other DOM APIs (`append`, `appendChild(createTextNode(…))`, `insertAdjacentText`, `replaceChildren`, `Object.assign(pane, {…})`, `innerText`, `pane['textContent']`).** Declined for this round. Review called it "bigger than the destructuring gap" and explicitly would not block on it, and it is correct that `Object.assign` is the sharpest of them since rule 1 already reads that form for classes. But this is a new *category* of match, not a correction to an existing one, and widening rule 2 across seven more APIs tree-wide needs its own false-positive measurement and its own round of adversarial construction. Every hole in this PR's history came from a widening that was argued for rather than measured. It should be its own change.
- **Array and object destructuring (`const { message } = err`, `const [first] = String(err).split('\n')`).** Same reason, and I want to be explicit that I read the review's argument that these are "not in the same category" as the accepted two-hop funnel and that they need no dataflow. That is right. What they do need is a decision about what `const { status } = err` binds, which is a new judgement in the most-reviewed function in this PR's history, made in the same round that replaces the masker underneath it. Two structural changes to `errorBoundNames()` in one round is how the last three rounds each opened a hole.
- **`.then(onOk, (e) => …)` rejection handlers.** Named in the new `CATCH_BINDING` doc comment as open. Finding it means splitting an argument list on the comma between two arbitrary expressions — a much less reliable question than reading forward from the word `catch`, and precisely the sort of pattern that has to guess.
- **N4 — `errorSinks()` over-includes.** Confirmed and left. Over-inclusion only ever adds failures to look at, the tree-wide rule-3 offender set is `[]`, and narrowing it means reading function bodies to prove forwarding. The `toContain` weakness review noted is real but is a test that cannot go wrong in the dangerous direction.
- **N6 — `entrypoints/` is swept but not scanned by the newlines guard,** and `entrypoints/options/main.ts:682` is the #308 defect today. Confirmed still true. It is pre-existing on `develop` and outside this diff; widening `SCANNED_DIRS` on the newlines guard pulls in the options page's whole stylesheet model, which is a different surface with a different contract. Named, not fixed.
- **The two-hop in-file funnel.** Still open, still asserted as a decision by its own test. Unchanged from round 4, and I agree with review's judgement that it is acceptable.
- **No real browser.** happy-dom only, fifth round running. The N5 fix's `role="alert"` landing on the element is a DOM fact confirmed here; whether an AT announces it is not answerable in this environment.

## What I could not verify

- **Whether an unbalanced comment backtick would ever have occurred.** Review could not find one in the tree and neither could I. B1's mechanism and blast radius are measured facts; that one of the 1,819 would eventually be odd remains a judgement. The fix is justified by the mechanism, not by a live firing.
- **Whether `new Error('')` is reachable from a real org.** Not traced. Same position review took: the swallow is confirmed, the path is not.
- **The injections I did not think of.** 33 probes across two fresh files is fewer than round 4's 102. I spent the budget on the two files no round had touched and on the base/head comparison rather than on breadth, and the 18-to-0 delta is the number I would want checked first.

---


scoobydrew83 (commented):
**APPROVE** — no blocking findings. Round 5 holds up under construction. Three of three blocking findings from the #327 review are closed, and I verified each by building the defect rather than by reading the fix; the new parser-based masker survives an independent oracle over the whole tree.

> Filed as a `COMMENT` because GitHub refuses `REQUEST_CHANGES` on a same-account PR. Line 1 is the verdict, per this series' convention.

Independent review at `ed1a878`, base `origin/develop` `d42b7c1`. Own worktree under `.claude/worktrees/`, nothing pushed, `FEATURES.json` untouched, tree clean at exit (`git status --porcelain` empty, all probe files deleted).

Four rounds of this guard have been wrong, and every real hole was found by constructing the #308 defect rather than by reading the regex. So I did not read my way to this verdict either. Every number below is mine, measured on both sides in one worktree; the two that matter most are **56,944** and **0**.

---

# How I looked

**1. An independent oracle for the mask.** `dynamicParts()` decides what every rule can see, and it was rewritten this round on top of the TypeScript AST. Rather than argue about whether the AST walk is right, I asked a *different* TypeScript API the same question and compared. For every one of the 125 scanned files I walked the parse tree, took every leaf token that is **not** a string / template / regex literal, and asserted it survives the mask byte-for-byte:

```
ORACLE identifiers checked: 56944; blanked by the mask: 0
ORACLE non-literal tokens blanked: 0
```

That is exactly the property the #327 review asked to be pinned — *"masking must never remove a code token that exists in the raw source"* — and it holds at token granularity across the entire tree, not on a sample.

**2. A much wider version of this PR's own backtick property.** The shipped `cannot be blinded by one stray backtick, in any file in the tree` probes the **first line comment of each file** — about 125 positions. I injected a backtick at **every offset the compiler itself reports as a comment**, line and block, in all 125 files:

```
probes 5116   failures 3
  features/soql-runner.ts   '//'
  features/soql-runner.ts   '//'
  lib/salesforce-api.ts     '//'
```

All three "failures" are my probe's fault, not the mask's: they are empty `//` comments where my injection point lands *between the two slashes*, turning the comment into a division. 5,113 of 5,113 real injections change nothing the mask calls code. B1 is genuinely closed, and closed more broadly than the PR claims.

**3. Parse health, tree-wide.** `ts.createSourceFile` reports **0 parse diagnostics** across all 125 scanned files, so the AST the mask trusts is the real one everywhere.

**4. 56 constructed probes in two fresh host files**, spliced into real source, run on both sides in the same worktree. Hosts: **`features/data-import.ts`** and **`features/field-impact.ts`** — neither has been used by any round, and neither is on the burned list (`api-name-generator.ts` and `debug-log-viewer.ts` are burned now and I avoided them). Probes are organised by **category**, not by spelling, because the last two rounds' surviving holes were both new categories. Driven through the real exported matchers (`buildsSfErrorPanel`, `rendersErrorIntoConsole`, `passesStringifiedError`) plus a faithful replica of the newlines guard's per-file scan.

**5. Every mutation reproduced**, plus two of my own on `DEFINING_ARTIFACTS`.

**6. Every gate re-measured on both sides**, same worktree, same install, `packages/flow-core` built once, `wxt prepare` re-run per side.

---

# My injection numbers

| | `d42b7c1` | `ed1a878` |
|---|---|---|
| 48 defect probes | **28 CAUGHT / 20 MISSED** | **44 CAUGHT / 4 MISSED** |
| 8 correct-code probes | 7 declined, **1 false positive** | **8 declined, 0 false positives** |

The base false positive is a **commented-out** defect read as live — the PR attributes that to rule 1; on my probe it fires on rule 2 as well, and it is gone at head either way.

**The 20 base misses**, by category: nine masker-blinder shapes (backtick in a line comment, in a block comment, in a regex character class, in a bare regex, a `` don`t `` after a division, a regex with a slash *and* a backtick in the class, the defect 60 lines later, and two combinations), seven `.catch(…)` binding shapes (`(zz)`, `async (zz)`, `function (zz)`, `function named(zz)`, expression-body arrow, newline-before-param, interpolated), the blinder-then-promise-catch combination, `renderSfError(String(zz))` via a promise catch, and the two destructuring shapes.

**The 4 head misses are all declines, and three of them are declines this PR names.** Listed under non-blocking below; one of them is a category the PR does not mention.

For calibration, these were CAUGHT at head and are worth naming so the table is not read as better than it is: `catch (zz)` statement form, unparenthesised `.catch(zz =>)`, typed `catch (zz: unknown)`, nested catch, five const hops, the `let +=` accumulator, the Prettier trailing-`+` wrap, `err.stack`, `err + ''`, `[String(err)].join('')`, `String(err.cause)`, `JSON.stringify(err)`, and `window.addEventListener('error', (ev) => … ev.message)`. Several of those were already closed at base.

---

# Claims I independently confirmed

### B1 — the parser-based masker

Beyond the oracle and the 5,116-position injection: **20 adversarial masker constructions**, all of which keep the defect visible at head. Backtick in a line comment, in a block comment, in a JSDoc fence, in a regex character class, in a bare regex, inside a single-quoted string, inside a double-quoted string, escaped inside a template; a comment inside a `${…}` hole; a string containing `/*`; a template containing `*/`; a regex containing `//`; a regex with `/` *and* `` ` `` in the class; a URL in a template; a division followed by a backtick comment; nested templates with a quoted hole; a tagged template; `satisfies` with generics; an angle-bracket cast; the defect 60 lines after the blinder.

Tree-wide token loss, the measurement round 4 made: **6 tokens differ**, all of them mentions inside comments —

```
features/soql-bulk-delete.ts   renderSfError(   raw=1  masked=0
ui/health-modal.ts             classList.add    raw=24 masked=23   (line 43, a JSDoc mention)
ui/panels.ts                   renderSfError(   raw=2  masked=1
lib/sf-error-guidance.ts       renderSfError(   raw=1  masked=0
lib/ui-controls.ts             .textContent=    raw=4  masked=3
lib/ui-styles.ts               renderSfError(   raw=1  masked=0
```

Round 4 measured five; the sixth is new because comments are now masked too, and it is correct.

**Mutation** — `dynamicParts()` restored to `d42b7c1`'s character scanner:

```
× cannot be blinded by one stray backtick, in any file in the tree
× drops a comment rather than reading it as code
× keeps the code that follows an unbalanced backtick
× binds no phantom name from a comment or a template
× a stray backtick in a comment cannot blind rule 2
  Tests  5 failed | 47 passed (52)
```

Exactly the five named tests. And the tree-wide property is **not vacuous under that mutation**: `firstCommentBody()` requires `masked[at] === ' '`, so at base it finds no probe position at all and the `probed > 100` assertion is what fails first. Good design.

The argument that a newline bound "was not available" checks out: `lib/ui-styles.ts`'s sheet and `features/soql-runner.ts`'s Python template are real multi-line templates, and `identifiersIn(dynamicParts('const py = \`\nquery = """\n\`;'))` is `['const','py']` at head.

### B2 — `.catch((e) => …)`

**Mutation** — `CATCH_BINDING` reverted to `\bcatch\s*\(\s*([A-Za-z_$][\w$]*)`:

```
× rule 2 knows a catch binding in its PROMISE form too
× rule 3 knows a catch binding in its PROMISE form too
  Tests  2 failed | 50 passed (52)
```

And on my own probes, seven `.catch` shapes flip from MISSED at base to CAUGHT at head. The `async` / `function` / named-`function` heads all read through. Confirmed the two live sites (`lib/spa-router.ts:40`, `entrypoints/background.ts:469`) are spelled `err`, so the spelling fallback really was covering the gap.

### B3 — the exemption bite check, **including the counter-mutation**

This is the claim I most wanted to check, because it is the builder asserting the reviewer was right. It reproduces exactly.

*With the reviewer's decorative entry added and `d42b7c1`'s version of the check restored:*

```
 Test Files  1 passed (1)
      Tests  13 passed (13)
```

Green. #327's check does accept a decorative `ui/toast.ts` / `toast` / `message` entry, exactly as the round-4 review reported.

*With the same decorative entry and the new check:*

```
× every exemption would actually be flagged without it
  Tests  1 failed | 13 passed (14)
```

*With the one real entry deleted:*

```
  Tests  1 failed | 13 passed (14)
```

Load-bearing, both directions.

### Golden principle #12 — both lists, both directions, at head

I ran the sibling mutations on the sweep myself:

```
decorative lib/icons.ts added to DEFINING_ARTIFACTS
  × every exclusion would actually fail without it        1 failed | 28 passed (29)
the real DEFINING_ARTIFACTS entry deleted
  × no feature, entrypoint or lib module hand-rolls the error block
  × no console pane is handed a caught error directly     2 failed | 27 passed (29)
```

`DEFINING_ARTIFACTS` is byte-identical to base (one member). `EXEMPT` is one member. `test/` is in neither guard's `SCANNED_DIRS`, so the new `test/error-source-scan.test.ts` is not an artifact of the checks it exercises. #12 satisfied.

### The `withoutComments()` split — "flips exactly one verdict"

Measured directly, by running rule 1 over all 125 files under three treatments:

```
HEAD_withoutComments: ui/panels.ts
RAW_source:           ui/panels.ts
FULL_dynamicParts:    (none)
```

Both claims are exact. Rule 1 genuinely **cannot** use `dynamicParts()` — the full mask empties the string literal that is the class name, and `ui/panels.ts` stops tripping the rule it is excluded for. And `withoutComments()` changes **no** rule-1 verdict versus raw source, so it is the latent fix the PR says it is.

### N2 — "measured tree-wide: zero new sites"

Measured over `features` + `ui`, every `X.textContent =` site, comparing `carriesAnError(expr, bindings)` against `rendersAnErrorValue(expr)`:

```
sites caught by NEW question only: 0
sites caught by OLD question only: 0
```

A claim made true rather than a widening — confirmed. **Mutation** (revert to `rendersAnErrorValue`): `× knows an error by its BINDING here too, not only by its spelling`, `1 failed | 13 passed (14)`.

### N5, N3, rule 1 — mutations

```
setSfError's unconditional clear restored
  × still renders a panel for a thrown error with no message   1 failed | 27 passed (28)
the \d boundary removed from wordsOf()
  × rule 2 knows an error by its binding, not only by its spelling   1 failed | 28 passed (29)
classesByElement reverted to raw source
  × does not read a commented-out hand-roll as a live one      1 failed | 28 passed (29)
```

**9 of 9 mutations reproduce, each red on exactly the named test.**

I also audited the N5 behaviour change against every call site. All nine `setSfError` callers pass a real thrown value; all ten clear paths call `clearSfError()` explicitly; all five pre-built empty panels use `renderSfError(null)`. **No caller passes `''` to clear.** The three funnels that pass a `string` (`rest-explore.ts:222`, `soap-explore.ts:290`, `soql-runner.ts:2132`) are `showError(message: unknown)` wrappers that also set `display: 'block'`, so the new behaviour is strictly better there. The fix is safe.

### The dependency claim, checked at the bundle and not at the manifest

- `typescript: ^5.6.3` is in `extension/package.json` `devDependencies` at head **and at base** — `git diff d42b7c1..ed1a878 -- extension/package.json package.json package-lock.json generated/ FEATURES.json` is empty. It resolves to `extension/node_modules/typescript` 5.9.3 from the extension's cwd (the root hoists 6.0.3; the extension does not pick that up).
- `test/error-source-scan.ts` is imported from `test/` only. No shipped source imports `typescript`.
- Built `.output/chrome-mv3`: **0 occurrences** of `createSourceFile` / `SyntaxKind` anywhere in the bundle.
- Manifest `sha256` **`2a302f6a1b97539b0857f162e24fb498774b602f8840b515d865ecc9e8359c7d`** on both sides, matching the PR and rounds 3 and 4.
- The bundle is **not** byte-identical — nor should it be, and the PR does not claim it is. `content.js` grows by **91 bytes**, and I diffed the two: the delta is exactly the `setSfError` empty-parts branch. Nothing else moved.

### Gates — my numbers, not the table's

| | `origin/develop` `d42b7c1` | head `ed1a878` |
|---|---|---|
| `extension/ npx vitest run` | 114 files / **2121** passed | 115 files / **2137** passed |
| `extension/ npx tsc --noEmit` | `exit=0` | `exit=0` |
| `extension/ npx eslint .` | `exit=0` | `exit=0` |
| `extension/ npx wxt build` | `Σ Total size: 1.4 MB` | `Σ Total size: 1.4 MB` |
| built manifest `sha256` | `2a302f6a…9c7d` | `2a302f6a…9c7d` |
| root `npm test` | 318 files / **5477** passed | 319 files / **5493** passed |
| root `npm run generate:catalogs` | — | no drift (`git status --porcelain generated/` empty) |

**Every number in the PR body reproduces exactly.** `d42b7c1` is an ancestor of `ed1a878`; 7 files changed, all under `extension/`.

**Prettier.** I confirmed your warning first and never ran `--write` on a whole file. `ui/panels.ts` and `test/panels.test.ts` are not clean at base. Piping each through `prettier --parser=typescript` and diffing against itself, the complaint set is **identical** before and after: `ui/panels.ts` 6 hunks both sides (the same `setSfError` signature wrap), `test/panels.test.ts` 7 hunks both sides (the same `const stack` string and the same `Object.assign` object). No new drift. The other four changed files are clean.

---

# Non-blocking

### NB1 — `Promise.allSettled`'s `.reason` is a fourth way to hold a thrown error, and it is named nowhere

My probe B15, MISSED at head:

```ts
const sink = doc.createElement('pre');
sink.className = 'sfdt-console';
const [r] = await Promise.allSettled([go()]);
if (r.status === 'rejected') sink.textContent = String(r.reason);
```

`ERROR_WORDS` deliberately drops `reason`, and the module's comment justifies that as:

> *"None of them needs to be here: as a `catch` binding they are reached by the catch clause…"*

For `.reason` on a settled result that is false — it is a **property name the language mandates**, not a variable name someone chose, and no `catch` binds it. The tree uses `Promise.allSettled` at `features/org-health-checks.ts:132` and `features/soql-bulk-delete.ts:443`.

**No live hole today**, and the reason is a near-miss worth stating: `org-health-checks.ts:136` happens to write `s.reason instanceof Error ? s.reason.message : String(s.reason)`, and the `instanceof Error ?` idiom has `subject: 'always'`, so the binding scan does catch it. Simplify that one line to `String(s.reason)` — an entirely ordinary edit — and the site goes dark. This is the *category* slot where I would look first in round 6, and it is cheaper to close than the destructuring category: the subject test for `String(x)` could accept `x.reason`, or `reason` could go back on the spelling list with the `allSettled` justification written down.

### NB2 — a live #308 defect inside the newlines guard's own `SCANNED_DIRS`, invisible to it

Stronger than N6, because N6 at least sits in a directory the guard admits it does not scan. This one is in `features/`. The chain is measurable end to end:

1. `lib/salesforce-api.ts:657` — ``throw soapError(advice ? `${headline}\n${advice}` : headline, …)``. Two lines: the org's text, then the guidance line.
2. `features/org-health-checks.ts:132-137` —
   ```ts
   const settled = await Promise.allSettled(checks.map((c) => c.run()));
   …
   const reason = s.reason instanceof Error ? s.reason.message : String(s.reason);
   return { …, summary: `Could not run: ${reason}`, findings: [] };
   ```
3. `features/org-health-checks.ts:160-162` —
   ```ts
   const summaryEl = doc.createElement('span');
   summaryEl.className = 'sfdt-muted';
   summaryEl.textContent = check.summary;
   ```
4. `.sfdt-muted` at `lib/ui-styles.ts:243` declares `font` and `color` and **no `white-space`**. Reading the sheet the same way the guard's own `WHITE_SPACE_CLASSES` does, the classes that carry one are exactly `sfdt-cell-pre`, `sfdt-console`, `sfdt-editor-gutter`, `sfdt-editor-hl`, `sfdt-editor-input`, `sfdt-msg`, `sfdt-sf-error-note`, `sfdt-sf-error-text`, `sfdt-table` — and `sfdt-muted` is not among them.

So an org failure in a health check renders its guidance line collapsed onto the org's text — the #308 defect verbatim. The guard cannot see it because `summary` is neither an error spelling nor a local binding: it is a field on a struct written in one function and read in another. That is the **documented-open cross-function funnel**, and the round-4 review judged leaving it open acceptable. I agree it should not block this PR — it is pre-existing on `develop` and outside this diff. But it is now a *second* live instance of that class alongside `entrypoints/options/main.ts:687`, and two live instances is a different argument from a theoretical gap. I would spend round 6 on this and on N6 rather than on more masker work.

(I re-confirmed N6: `entrypoints/options/main.ts:687` still carries `saveStatus.textContent = err instanceof Error ? err.message : String(err)` into a `.status show error` span, and there is **no `white-space` anywhere** in the options page's stylesheet.)

### NB3 — the masker fails open on a parse error, and says nothing

```
dynamicParts('/* never closed\npane.textContent = err.message;')
  →  "               \n                               "
```

`blankLiterals()` never consults `sourceFile.parseDiagnostics`. A file the parser cannot read is silently blanked in full, which is the same *failure shape* as `LITERAL_ONLY` and as #327's backtick — silence, not a wrong answer.

**Unreachable in practice, and I checked rather than assumed:** `.github/workflows/extension.yml` runs `npm run compile --workspace=@sfdt/extension` (`tsc --noEmit`) **before** `npm run test:extension`, and a job step failure stops the job. A non-parsing file cannot reach the guard green. And the tree measures 0 parse diagnostics today.

Still worth one line in the tree-wide suite, because it costs nothing and converts an invisible failure into a loud one — assert `parseDiagnostics` is empty for every scanned file. The guard's entire history is silent failures; the mask should refuse to answer when it does not know.

### NB4 — `CATCH_BINDING` also binds a bare callback reference

`.catch(handleError)` binds `handleError`; `.catch(console.error)` binds `console`; `.catch(reject)` binds `reject`. The doc comment says the pattern reads "the identifier a `catch` binds", and none of those is one. Only ever a false-positive direction, and it does not fire: I grepped the tree for `.catch(<ident>)` and found **none**, and tree-wide the binding scan produces **55 names, zero suspicious** (no `this`, `doc`, `el`, `pane`, …). Worth a sentence in the comment rather than a change.

### NB5 — the shipped backtick property probes one position per file

`firstCommentBody()` returns on its first hit, so each file is probed at exactly one offset (~125 total). Mine probed 5,116 and found nothing, so the weaker version is not hiding anything **today** — but the property is cheap to run at every comment position, and the version that would have caught B1 is the strong one. Consider iterating rather than returning.

### NB6 — the declined destructuring category is narrower than round 4 measured

Worth recording so round 6 does not re-litigate the whole category. At head:

| | |
|---|---|
| `const { message } = err as Error; sink.textContent = message` | **CAUGHT** (by spelling — `message` is an error word) |
| `catch ({ message: zz }) { sink.textContent = zz }` | MISSED |
| `const [first] = String(err).split('\n'); sink.textContent = first` | MISSED |

So the plain destructure — the "far more ordinary authoring" round 4 pointed at — is already covered. Only the **renamed** forms miss. That makes the decline cheaper than the review's list implies, and it makes me more comfortable with it.

### On the declines generally — I checked whether each shape occurs

| declined shape | occurs in the tree today? |
|---|---|
| `.then(onOk, (e) => …)` | **no** — zero matches for a two-argument `.then` with a handler |
| `const { … } = err` / `catch ({ … })` | **no** |
| `const [x] = String(err).split(…)` | **no** (the one array destructure, `lib/apex-log/tokens.ts:63`, is a CSV split) |
| N1's seven DOM APIs, carrying an error | **no** — the two `.append` sites pass *elements*, and `features/ai-assistant.ts:211` is `replaceChildren(renderSfError(message, { doc }))`, which is the correct path |
| N4 `errorSinks()` over-inclusion | confirmed, harmless, rule-3 offender set is `[]` |
| N6 `entrypoints/` | confirmed still live, see NB2 |

**Every decline except NB1 is defensible on "this shape is not in the tree", not only on "widening would need its own measurement".** That is a materially stronger position than the PR body claims for itself, and I think the "one structural change per round" discipline is the right call given this line's history — three rounds in a row shipped a hole through a widening that was argued for rather than measured, and this round shipped none.

---

# What I could not verify

- **No live Salesforce org, no real browser.** happy-dom only, fifth round running. `role="alert"` landing on the element is a DOM fact I confirmed by mutation; whether an assistive technology announces it is not answerable in this environment.
- **NB2's collapse is a reasoned chain, not a screenshot.** I traced `salesforce-api.ts:657` → `org-health-checks.ts:136` → `:162` → `lib/ui-styles.ts:243` in source and read the class out of the sheet the guard itself reads. I did not render the health panel against a failing org and look at it.
- **Whether `new Error('')` is reachable from a real org.** Not traced. The swallow and its fix are confirmed; the path is not — the same position rounds 4 and 5 took.
- **Whether an unbalanced comment backtick would ever have occurred.** I did not find one either. B1's mechanism and blast radius are measured facts; the fix is justified by the mechanism, and after this round the question is moot because the mask no longer has to decide what a backtick means.
- **The injections I did not think of.** 56 probes across two fresh hosts is fewer than round 4's 102, and I spent the rest of the budget on the token-level oracle and the 5,116-position injection instead of on breadth. I would trust the oracle numbers (`0 / 56944`) far more than the probe count: they are a property over the whole tree rather than a sample of it. The probe set is where I expect round 6 to find things, and **NB1 is where I would look first**.
- **Performance under growth.** The extension suite runs 42.9s at head against 40.8s at base — about +5% for parsing every scanned file and every unique expression fragment. The `MASKED` / `NO_LITERALS` / `NO_COMMENTS` caches are bounded at 8192 entries each, so there is no unbounded growth, but I did not profile how the cost scales as the tree does.
- **Anything outside the sweep and `ui/panels.ts`.** The other 5,400 root tests and 2,100 extension tests pass on both sides; I did not review what they assert.

---

Nothing pushed, `FEATURES.json` untouched, all probe files deleted, worktree clean at `ed1a878` with `115 passed (115)` / `2137 passed (2137)`.

Four bad rounds is not evidence the fifth is bad; it is evidence that reading is not enough to tell. I constructed, injected, mutated and measured, and this one holds. Ship it.

---

PR #330, opened 2026-08-06
fix(extension): the guard could not see a settled rejection, and the masker failed open

scoobydrew83:
Round 6 of C-FIX-4, plus the fix for the blocking finding its own review raised. Base is `origin/develop` @ `c7547e7`. Branch `ext/c-fix-4-nb-followups`, head **`ad199f0`**. `FEATURES.json` untouched. Not merged.

Two commits: `b710a54` closed the round-5 non-blocking findings, and **shipped a regression doing it**. `ad199f0` fixes that regression and the four non-blocking items from the round-6 review. The regression is described first, because it is the most important thing on this page.

---

## B1 (round-6 review, BLOCKING) — the fail-closed mask manufactured the condition it refused on

`b710a54` added a refusal on an unterminated literal. `blankLiterals()` blanks a string literal's interior with spaces **while preserving newlines** — that is the offset contract — and a backslash immediately before a line break is a **LineContinuation**: legal ES, legal TS, and `npx tsc --noEmit` and `npx eslint` both exit 0 on it in this repo. Eating the backslash and keeping the newline leaves the newline inside the quotes with nothing holding them open:

```
RAW     const s = 'a\⏎ b';   ← parses, zero diagnostics
MASK1   const s = '  ⏎  ';   ← an unterminated string that was never written
MASK2   REFUSED              ← at c7547e7: MASK2 OK, idempotent
```

The guards mask their own output constantly — `readExpression()` cuts fragments out of the masked buffer and `flattensAnError()` masks them again — so the second pass threw. Review's measurement, one three-line legal file dropped into `features/`: the three guard suites go from **52 passed (52)** at `c7547e7` to **6 failed / 54 passed (60)** at `b710a54`. A regression against the base, triggered by source a contributor could write tomorrow.

**Reproduced exactly before fixing anything**, including the detail that it lands in the `readExpression()` fragment category — the same category `still answers for the FRAGMENTS the guards legitimately hand it` exists to protect and the same category the body claimed to have accounted for.

**The fix.** `blankLiteralInto()` keeps a continuation verbatim, backslash and terminator both, which preserves every offset and every line number exactly as before and leaves a literal that still closes. Escapes are tracked properly on the way through, so an escaped backslash followed by a real newline is **not** read as a continuation — that one is genuinely unterminated in the raw source and is still refused. Regex literals keep plain blanking, since they are blanked whole, delimiters included, and a preserved backslash there would land in code position.

**And the claim is a property now, not a measurement of one day's tree.** The body previously said *"this refuses nothing the guards legitimately ask"*, which was true of that tree and was not a statement about the trigger. `masking its own output is always accepted, and always a no-op` asserts, for legal source: the mask accepts it, the output is the same length and the same number of lines, the mask accepts **its own output**, and the second answer equals the first. It runs over

- a 16-member corpus of legal-but-hostile constructs — six continuation shapes including CRLF and escaped-backslash-then-continuation, a continuation inside a comment, a continuation carrying an escaped quote, quotes inside template holes, nested templates, a regex holding a quote and a slash, a string holding a comment opener, a string holding a backtick, and division-that-is-not-a-regex — **each asserted LEGAL against the compiler first**, so no member can rot into an illegal-input test that passes by being refused;
- every one of the 125 scanned files;
- **every one of the 125 scanned files with a line continuation spliced in**, so the tree-wide half does not rest on today's tree happening not to contain one. Which is exactly how this was got wrong the first time.

Pinned at the guard level too, where it was measured: `a line continuation in the source does not take the rules down` requires all three rules to answer, to still catch the defect below the continuation, and to still not read the continued string as a value.

---

### NB1 — a rejected settlement's reason, a fourth way to hold a thrown error, named nowhere

`reason` is deliberately off `ERROR_WORDS`, justified by *"as a `catch` binding they are reached by the catch clause"* — true of `problem` and `thrown`, **false of this one**. `PromiseRejectedResult.reason` is a property name the LANGUAGE mandates; no author chose it and no catch clause binds it.

Constructed against `c7547e7` first: ten probe shapes, all MISSED; all ten CAUGHT at head. Not a live hole only because `org-health-checks.ts:136` happens to write the `instanceof Error ?` idiom, which matches on `subject: 'always'`.

The fix is structural and scoped by evidence — `Promise.allSettled(`, or the `.status === 'rejected'` narrowing TypeScript requires before `.reason` is reachable. The narrowing is matched on the MASKED buffer and the state name read back out of RAW source at the same offset, which is what the offset contract is for. **The state comparison now includes the closing quote** (round-6 review, item 1): a prefix match let `s.status === 'rejectedByAdmin'` scope `reason` for a whole file. Two probes pin both halves.

Tree-wide, independently measured, `c7547e7` versus `ad199f0` — **byte-identical**:

| | `c7547e7` | `ad199f0` |
|---|---|---|
| distinct binding names | 11 | 11, same set |
| rule 1 / 2 / 3 offenders | `[]` / `[]` / `[]` | `[]` / `[]` / `[]` |
| `carriesAnError` sites, features + ui | 12 | 12, same list |

### NB2 — a LIVE #308 defect inside the newlines guard's own SCANNED_DIRS

`salesforce-api.ts:657` throws two lines; Org Health folds them into `summary`; the span wore only `.sfdt-muted`, which declares no `white-space`. **Fixed in both renderers** — the live rows in `org-health-checks.ts` and the CLI-snapshot rows in `org-health.ts`, the same field rendered the same way — each pinned by a rendered-DOM test that drives a real two-line failure and reads the qualifying classes out of the sheet.

Whether the guard could be taught to see it was **measured, not argued**. The object-literal-property widening was built and run tree-wide: binding names 11 to 30 (gaining `id`, `title`, `status`, `ok`, `T`), rule 3's offender set from `[]` to **seven false positives on correct code** including `setSfError(errorPanel, message, { doc })`, and it claims `check.title` alongside `check.summary` — and it still cannot reach the CLI half, whose other end is a JSON payload from another process. **Declined on the record, no exemption added**, with `the struct-field funnel is open, and that is a MEASURED decision` carrying the numbers.

### NB3 — the masker failed OPEN

`dynamicParts()` handed an unterminated block comment returned a buffer of nothing but spaces, which every rule reads as a file with no code — green and silent, the third time this guard has shipped that shape. Unreachable only because CI happens to run `tsc --noEmit` first.

"Any parse diagnostic" is the wrong trigger, and that was measured: across the whole extension suite **385 masked inputs carry a diagnostic and every one is a legitimate expression fragment**. The trigger is the runaway itself — `isUnterminated`, or a block comment with no close — of which there are **zero** across the same suite. Plus the assertion review asked for verbatim: every scanned file parses with zero diagnostics.

**B1 above is the correction to this fix**, and it is worth being plain about the shape of the mistake: closing a fail-open by adding a refusal, and not checking that the thing doing the refusing could not create the condition itself.

---

## Also closed

- **NB4** — the bare-callback over-claim (`.catch(handleError)` binds `handleError`) documented at the definition and pinned by a decision test. The mutation is the argument for not patching it: making `\(?` mandatory turns **9** tests red.
- **NB5** — the backtick property widened from 125 positions to **917**, now **stratified by comment kind**, with a per-file property that every kind a file contains appears in that file's own sample. **The claim it made is also corrected** (round-6 review, item 3): measured over the tree there are **4,522** line comments, **579** JSDoc and **zero** plain block comments, so `block=0` was a fact about the codebase and not a sampling failure. `commentRanges()`'s block branch is exercised by the JSDoc ones; a plain block comment is covered directly by a unit case, since the tree cannot cover it. The tree-wide composition is now pinned so `block` appearing later is a visible change.
- **Cache bounding**, which fell out of measuring NB5: three 8,192-*entry* caches of whole files reached **1,166 MB** resident under the exhaustive probe; a 4 MB budget each brings it to **372 MB** with identical answers. This is the "performance under growth" item review listed as unprofiled.
- **`ts.createScanner` is the wrong tool for finding comments**, confirmed and worse than first described: driven without the parser it re-enters a template literal after the first hole, and on `ui/workspace-host.ts` it reports **158** comments where the compiler reports **39** — 23 of the phantoms inside the CSS template. Injecting a backtick there terminates the template early, which looks exactly like a mask bug and is not one.

### Why not all 4,899 positions

Built and run, not estimated: **4,899 probes, 0 failures, +40 seconds of wall clock on a 45-second suite** — the run goes to 85s, because each probe re-masks a whole file at ~4.8 ms of parse. At 8 stratified positions per file it is **917 probes**, which vitest's parallelism absorbs: **45.00s at `c7547e7`, 46.98s at `ad199f0`**. Both the number and its cost are written into the test.

---

## Mutation evidence

15 mutations, each reverted in place, the named test observed red, then restored. The three new ones first.

| # | mutation | test that goes red | result |
|---|---|---|---|
| **B1** | the line-continuation fix reverted to plain blanking | `masking its own output is always accepted, and always a no-op` · `a line continuation in the source does not take the rules down` | **2 failed / 61 passed** |
| NB1 | the settled state loosened back to a prefix match | `knows a rejected settlement's reason, which no catch clause binds` | **1 failed / 62 passed** |
| NB5 | the sampling unstratified (one pool, not one per kind) | `cannot be blinded by one stray backtick, in any file in the tree` | **1 failed / 62 passed** |
| NB1 | the settled-reason binding removed | `knows a rejected settlement's reason…` · `rules 2 and 3 know a rejected settlement's reason` | **2 failed** |
| NB1 | the same binding UNSCOPED | `leaves the word 'reason' alone in a file that settles nothing` · `binds a bare callback reference too…` · `binds no phantom name…` · `every element assigned an error message declares a white-space rule` | **4 failed** |
| NB2 | `org-health-checks.ts` back to `'sfdt-muted'` | `keeps the guidance line on its own line when a check's error is two lines` | **1 failed** |
| NB2 | `org-health.ts` back to `'sfdt-muted'` | `keeps a snapshot summary on two lines when the CLI reports a two-line failure` | **1 failed** |
| NB3 | the unterminated-LITERAL refusal removed | `refuses to answer rather than blanking a file it could not read` | **1 failed** |
| NB3 | the unterminated-COMMENT refusal removed | *(same test)* | **1 failed** |
| NB5 | `PROBES_PER_FILE` back to 1 | `cannot be blinded by one stray backtick, in any file in the tree` | **1 failed** |
| NB4 | `CATCH_BINDING`'s `\(?` made mandatory | 9 tests across all three guard suites | **9 failed** |
| #12 | decorative `ui/toast.ts` added to `EXEMPT` | `every exemption would actually be flagged without it` | **1 failed** |
| #12 | the real `EXEMPT` entry broken | that plus `every exemption still matches real source` and the offender scan | **3 failed** |
| #12 | decorative `lib/icons.ts` added to `DEFINING_ARTIFACTS` | `every exclusion would actually fail without it` | **1 failed** |
| #12 | the real `DEFINING_ARTIFACTS` entry broken | that plus rules 1 and 2 | **3 failed** |

**15 of 15 reproduce.** Two rows are corrected from the first version of this body, which under-counted them by running a subset of the suites — review caught both, and every mutation now runs all three guard suites. Golden principle #12: both lists load-bearing in both directions; `test/` is in neither guard's `SCANNED_DIRS`.

---

## Constructed injection — 54 probes, three sides

**35 defect probes, 14 correct-code probes, 5 known-open probes**, spliced into the real source of two hosts no round has used (`features/dependency-explorer.ts`, `features/canvas-search.ts`). Every probe runs against BOTH hosts and only counts if they agree. Driven through the real exported matchers plus a faithful replica of the newlines guard's per-file scan.

| | `c7547e7` | `b710a54` | head `ad199f0` |
|---|---|---|---|
| 35 defect probes | 23 CAUGHT / **12 MISSED** | 30 CAUGHT / 5 REFUSED | **32 CAUGHT / 3 REFUSED / 0 MISSED** |
| 14 correct-code probes | 14 declined, **0 FP** | 11 declined, **3 FP** | 13 declined, **1 FP** |
| 5 known-open probes | 4 open, 1 caught | 4 open, 1 caught | 4 open, 1 caught |

**The middle column is the regression, and it is left in the table on purpose.** At `b710a54` the three line-continuation probes all came back REFUSED — two defect probes and one correct-code probe that a legal file turned into a hard error — and `s.status === 'rejectedByAdmin'` was a false positive. Both are gone at head.

**The one remaining false positive is named rather than hidden.** It is review's OK3: a file containing `Promise.allSettled` where a *domain* `planned.reason` is rendered. NB1's scoping is file-wide, so it claims that too. The first version of this body admitted the over-claim in prose while the table read 0 false positives; the row is in the table now. It is the conservative direction — a site to look at, not a rule going quiet — it is confined to files that handle settlements, and it produces **zero sites in the actual tree** (the tree-wide table above is unchanged). Narrowing it needs receiver-precision in `carriesAnError()`, which is a second structural change to the binding scan and is not going in this round.

**The 12 misses at `c7547e7`:** ten settled-reason shapes plus the unterminated block comment and the unterminated template.

**REFUSED is counted separately and is not a catch.** Of the three, two were MISSED at base; the third (unterminated single-quoted string) was **CAUGHT** at base, because TypeScript terminates a single-quoted string at the newline so the blanking was bounded to that line. Head turns that catch into a refusal — louder, not quieter, but a change.

---

## New open categories the review found — MISSED on both sides, so not regressions, and deliberately not fixed here

Adding a match category on top of a masker fix and a blocking regression is exactly how rounds 2, 3 and 4 each opened the next hole.

| category | in the tree | status |
|---|---|---|
| **`for-of` bindings** — a for-of head whose iterable flattens an error | **263** heads of `for ((const or let) IDENT of …)`, **301** counting every binder form, 34 destructuring | **The largest named gap.** Pinned open by `the for-of binding is open, and it is the LARGEST named gap`, which also shows the same value one line later through a binding the scan does follow is caught — so the gap is the for-of head and nothing else |
| `catch ({ message: zz })` — renamed destructure | 0 | open; probe O2 |
| `const { stack } = err` — destructured non-error word | 0 | open; probe O3, and a shape neither round 4 nor 5 named |
| `const [first] = String(err).split(…)` | 0 | open; probe O4 |
| `switch (s.status)` narrowing | **0 in the tree** | open in principle; no site to construct against |
| `Promise.any` / AggregateError | 2 mentions, 1 real call (`lib/sfdt-bridge.ts:339`) | **could not reproduce as open** — my probe O5 is CAUGHT on both sides, because a `catch` binds the AggregateError. Recorded as unconfirmed rather than claimed |

`errorBoundNames()`'s declaration scan requires an `=` after the name, and a for-of head has none — that is the whole mechanism. It is one category and it wants one widening, measured on its own.

---

## Gates

All three sides measured in one worktree, same install, `packages/flow-core` built once, `wxt prepare` re-run per side.

| | `origin/develop` `c7547e7` | head `ad199f0` |
|---|---|---|
| `extension/ npx vitest run` | 115 files / **2137** passed, 45.00s | 115 files / **2150** passed, 46.98s |
| `extension/ npx tsc --noEmit` | `exit=0` | `exit=0` |
| `extension/ npx eslint .` | `exit=0` | `exit=0` |
| `extension/ npx wxt build` | `Σ Total size: 1.4 MB` | `Σ Total size: 1.4 MB` |
| built manifest `sha256` | `2a302f6a…9c7d` | `2a302f6a…9c7d` |
| root `npm test` | 319 files / **5493** passed | 319 files / **5506** passed |
| root `npm run generate:catalogs` | no drift | no drift |

### Bundle delta, accounted for

`ad199f0`'s build is **byte-identical to `b710a54`** — this commit touches only `test/` and the changelog. Against `c7547e7`:

- `content-scripts/content.js` **633,460 to 633,478, +18 bytes**
- `chunks/workspace-host-*.js` **189,917 to 189,935, +18 bytes**

Both are the same thing: two source sites changed from `'sfdt-muted'` to `'sfdt-muted sfdt-msg'`, **9 characters each, twice, in each of the two bundles that contain both features**. `2 x 9 = 18`, both times. `chunks/app-*.js` (818 B) and `chunks/sidepanel-*.js` (1,181 B) are byte-length identical and differ only in the hashed filename they import (verified by reading the import line out of the built chunk); the two HTML files differ only in that script `src`. `manifest.json` is byte-identical.

`git diff c7547e7..ad199f0 -- generated/ FEATURES.json package.json package-lock.json extension/package.json` is empty. No new dependency.

### Prettier

Never `--write` on a file that is not clean at base.

- Clean at `c7547e7` and still clean, 0 hunks both sides: `test/error-source-scan.ts`, `test/error-source-scan.test.ts`, `test/error-render-newlines.test.ts`, `test/sf-error-panel-contract.test.ts`.
- Not clean at base, complaint set **identical**: `CHANGELOG.md` 57/57, `features/org-health.ts` 6/6, `test/org-health-checks.test.ts` 4/4, `test/org-health.test.ts` 8/8. Three added blocks were hand-formatted to Prettier's own output so they added no new instance of an existing complaint.
- `features/org-health-checks.ts` 8/8, same set — the one textually different hunk is the file's pre-existing whole-function re-indent, which now contains the added comment lines. They follow the file's local four-space indentation, which is what keeps the count at 8.

---

## Declined, with reasons

- **N1 — the seven other DOM APIs.** Review confirmed none of these shapes carries an error anywhere in the tree. Its own change, with its own false-positive measurement.
- **NB6 — destructuring, re-measured at head** rather than re-litigated: `const { message } = err`, `const { message } = zq` and `catch ({ message })` are all CAUGHT by spelling; `catch ({ message: zz })`, the array destructure and `const { stack } = err` miss. Narrower than round 4 measured, and not only the *renamed* forms as round 5 concluded.
- **`.then(onOk, handler)`**, **N4 `errorSinks()` over-inclusion**, **N6 `entrypoints/`**, **the two-hop in-file funnel** — unchanged, all still pinned or named as before.
- **Narrowing NB1's file-scoped over-claim.** Needs receiver-precision in `carriesAnError()`, a second structural change to the binding scan. Owned in the injection table instead.

## What I could not verify

- **NB2's fix is a DOM fact, not a screenshot.** Sixth round with no real browser and no live org.
- **That B1's continuation is the ONLY way the mask can manufacture a token.** The argument is that literal and comment spans come from the SCANNER, so a stray brace makes the parser recover but does not move a quote, and blanking a comment or a regex removes a span rather than creating one. The evidence is the accept-and-idempotence property over the corpus and over the tree with a continuation spliced into every file — which is a much stronger statement than the one this branch shipped the first time, and is still not a proof. **This is the specific thing I would attack next**, because it is where I was wrong before.
- **`Promise.any` / AggregateError as an open category.** Review named it; my probe for it is caught on both sides. Recorded as unconfirmed rather than claimed either way.
- **Whether the `String(s.reason)` simplification would ever have been written.** The hole is measured; the likelihood is a judgement.
- **The injections I did not think of.** 54 probes across two fresh hosts. The `for-of` category, at 263 sites, is where I would look first.

---


scoobydrew83 (commented):
**REQUEST CHANGES** — one blocking finding, constructed. Everything else in this PR reproduced exactly, including the numbers I most expected to be reconstructions after the fact.

> Filed as a `COMMENT` because GitHub refuses `REQUEST_CHANGES` on a same-account PR. Line 1 is the verdict, per this series' convention.

Independent review at `b710a54`, base `origin/develop` `c7547e7`. Own worktree under `.claude/worktrees/review-330`, nothing pushed, `FEATURES.json` untouched, all probe files deleted, tree clean at exit (`git status --porcelain` empty, `115 passed` / `2147 passed`).

Five rounds of this guard were wrong and every real hole was found by constructing the defect. So I did not read my way to this verdict either. The blocking finding below is a three-line legal source file that is **green at `c7547e7` and turns six guard assertions red at `b710a54`**.

---

# BLOCKING

## B1 — the fail-closed mask manufactures the exact condition it refuses on, and refuses on legal source

`blankLiterals()` blanks a literal's interior with spaces **while preserving newlines**. That is safe for every literal kind but one: a string with a **LineContinuation** — `\` immediately before a newline, which is legal ES/TS and which this repo's ESLint and `tsc` both accept (I checked: `npx eslint` exit 0, `npx tsc --noEmit` exit 0 on the file below).

Blanking eats the backslash and keeps the newline, so the mask's own output contains an **unterminated string literal that was not in its input**. The new `isUnterminated` refusal then fires on the mask's own output.

Isolated, at `b710a54`:

```
RAW: "const s = 'a\\\n b';\nconst t = 1;\n"
raw parse diagnostics: 0
MASK1 OK:      "const s = '  \n  ';\nconst t = 1;\n"
MASK2 REFUSED: error-source-scan: refusing to mask source with an unterminated StringLiteral (line 1)…
```

At `c7547e7` the same input is `MASK2 OK, idempotent=true`.

**This is not a double-mask curiosity — it takes the primary guards down.** I dropped one file into `features/`, nothing else:

```ts
// extension/features/zzz-probe-lc.ts
const s = 'a\
 b';
export const t = s;
```

| | `c7547e7` | `b710a54` |
|---|---|---|
| `vitest run test/error-source-scan.test.ts test/sf-error-panel-contract.test.ts test/error-render-newlines.test.ts` | **52 passed (52)** | **6 failed \| 54 passed (60)** |

The six:

```
× every exemption would actually be flagged without it
× the exemption bite check rejects a decorative entry
× no console pane is handed a caught error directly
× nothing hands the renderer an already-stringified error
× is idempotent, in every file
```

And the buffer the mask refused on, from the stack:

```
--- first 200 characters of what was handed to the mask ---
 '  
  '
 ❯ refuseRunaway test/error-source-scan.ts:294:9
```

That is a `readExpression()` **fragment** cut out of an already-masked buffer — precisely the category `still answers for the FRAGMENTS the guards legitimately hand it` exists to protect, and precisely the category the PR body says was accounted for:

> *"re-masking an already-masked fragment turns a blanked regex into `replace( , ' ')` … Measured across the same suite: zero occurrences, so this refuses nothing the guards legitimately ask."*

The measurement is real (I reproduce it below: 393 masked inputs carry a diagnostic across the suite, and the only four inputs carrying an unterminated literal or comment are the four deliberate probes in the new test). But it is a measurement of **today's tree**, and the body states it as a property of the trigger. It is not one. The trigger is evaluated on buffers the mask itself produced, and the mask does not preserve the well-formedness of the literals it blanks.

**Why I am blocking on a false alarm rather than a blind spot.** It cannot make a guard go silent — it is the opposite failure, and that matters. But it is a regression this diff introduces: correct, lint-clean, type-clean source that is green at base is red at head, and the message it prints tells the author *"Fix the source … do not relax this check"* about source that is fine. This series' whole discipline is that the fix must be measured rather than argued, and the general claim behind NB3 is falsifiable in three lines. The fix looks small — keep the LineContinuation backslash when blanking a string interior, or exempt a literal whose interior is already all spaces — and it wants the same treatment everything else here got: build it, then measure the suite on both sides.

If the maintainer judges a loud false alarm on a shape absent from the tree to be acceptable and ships this anyway, that is defensible — but then the NB3 paragraph in the body should be corrected, because as written it is false.

---

# Non-blocking

### NB1's marker matches a PREFIX, not a state name

`handlesSettledResults()` reads the raw literal and asks `startsWith('rejected') || startsWith('fulfilled')`. Constructed:

```
SCOPED   if (deploy.status === 'rejectedByAdmin') note(deploy.reason);
SCOPED   if (order.status !== 'fulfilledAtStore') note(order.reason);
not      if (job.status === 'queued') note(job.reason);
```

A domain status literal that merely *begins* with either word scopes `reason` as an error word for that file's whole body. Zero such literals in the tree today (only the two files the PR names scope), so this changes nothing now; the closing quote is one character away.

Also measured, as declines rather than defects: `` s.status === `rejected` `` (template-quoted) is **not** scoped — 0 occurrences in the tree. `Promise . allSettled (jobs)`, `s?.status`, and double-quoted states all **are** scoped.

### The "11 correct-code probes, 0 false positives" table omits the shape the PR itself admits

The `soql-bulk-delete.ts` over-claim is documented at the definition and in the body — but it is not represented in the correct-code probe set, so the table reads cleaner than the behaviour is. My probe `OK3`:

```ts
const rs = await Promise.allSettled(zqJobs);
const zqPlanned = { reason: 'SKIPPED_LOCKED' };
zqPane.textContent = zqPlanned.reason;   // a domain code, correct code
```

`c7547e7`: declined. `b710a54`: **CAUGHT** — a false positive on correct code. One of eight in my correct-code set, zero at base. It is the admitted trade, not a surprise; the number should just say so.

### NB5's shipped sample contains no plain block comments

Measured on the shipped `PROBES_PER_FILE = 8`: **917 of 4,899** usable positions, `line=784 jsdoc=133 block=0`. The body says the sample "probes block comments, JSDoc and trailing comments — three shapes the header-comment-only version never reached." JSDoc yes, plain `/* … */` no. The exhaustive set reaches them; the sample does not.

### Two mutation rows under-count, in the direction that helps you

Running three files rather than two, `M2` (NB1 unscoped) is **4 red, not 3** — the extra is the tree-wide `every element assigned an error message declares a white-space rule`, i.e. the unscoped widening produces a real tree-wide false positive, not only three unit failures. `M8` (`\(?` mandatory) is **7 red, not 6** — the extra is `knows an error by its BINDING here too, not only by its spelling`. Both make the respective decisions *more* load-bearing than claimed.

### Categories still open — my probes, in files no round has used

Hosts: `features/schema-browser.ts` and `features/trace-flags.ts` (avoiding `rest-explore`, `apex-anonymous`, `api-name-generator`, `debug-log-viewer`, `data-import`, `field-impact`, `dependency-explorer`, `canvas-search`). Every probe run against both hosts; both agreed on all 37.

MISSED at head **and** at base, so not regressions — categories:

| probe | in the tree? |
|---|---|
| `switch (s.status) { case 'rejected': … }` — a narrowing TS accepts, neither marker present | `switch (x.status)` occurs **1×** |
| `if (isRejected(s)) … s.reason` — imported type-guard | the PR names this one itself |
| `const { status, reason } = s; if (status === 'rejected') …` | destructured settled |
| `for (const line of String(err).split('\n')) pane.textContent = line` | **`for (const x of …)` occurs 265×** |
| `for (const sub of err.errors) pane.textContent = sub.detail` | `.errors` occurs 29× |
| `Object.entries(m).forEach(([, v]) => …)` — array-destructured callback param | 13× |
| `Promise.any` → `AggregateError.errors` | 0× |

**The one I would spend round 7 on is the for-of binding**, not destructuring: `for (const X of <expression that flattens an error>)` is a declaration form the scan does not read at all, and the shape occurs 265 times in the tree. It is the same *kind* of gap as NB1 — a binding the language makes in a position the declaration regex does not look at — and it is a much larger surface than `const { stack } = err`.

---

# Claims I independently confirmed

All measured by me, in one worktree, base measured by checking out `c7547e7` in place.

### NB1 — tree-wide invariance

| | `c7547e7` | `b710a54` |
|---|---|---|
| scanned files | 125 | 125 |
| distinct binding names | **11** | **11 — identical set** |
| `carriesAnError` sites, `features/` + `ui/` | **12** | **12 — identical list** |
| files scoped by `handlesSettledResults()` | — | exactly `org-health-checks.ts`, `soql-bulk-delete.ts` |
| rule 1 / 2 / 3 offender sets | `[]` | `[]` (suite green both sides) |

The set both sides: `RegistryLogger attempt captureNote err errorMsg logger message msg reason refusal text`. Diffing the **per-file** lists, the entire tree-wide delta of this change is one line:

```
only-head: features/soql-bulk-delete.ts: err,reason
only-base: features/soql-bulk-delete.ts: err
```

— the admitted `planned.reason` over-claim, and nothing else. `org-health-checks.ts` binds `reason` on **both** sides, which is the near-miss the body describes (at base via the `instanceof Error ?` declaration, at head structurally).

**The raw/masked offset coupling, which is the risky part of the scoping.** Asserted as a property rather than trusted: for all 125 files, every byte of `dynamicParts(source)` is either identical to `source` at the same index, or a space, or a newline. **0 mismatches.** A masked-buffer match offset therefore indexes the raw buffer correctly by construction, and the evidence-must-be-code tests (comment mention, JSDoc mention, SQL template, string) all hold.

### NB2 — the live defect, and the decline

- `lib/ui-styles.ts:670` — `.sfdt-msg { white-space: pre-line; }`. `.sfdt-muted` declares `font` and `color` and no `white-space`. The fix is real. `SFDT_COMPONENT_CSS` is adopted by `ui/shadow-host.ts` and by every entrypoint, so both surfaces get the rule.
- Both rendered-DOM tests go red when their own line is reverted (mutations M3, M4 below), so the class is load-bearing in each test.
- **The declined widening was BUILT and measured — it is not a reconstruction.** I added `/[{,]\s*([A-Za-z_$][\w$]*)\s*:/` to the declaration loop and re-measured:

| | shipped | with the widening |
|---|---|---|
| distinct binding names tree-wide | 11 | **30** — gaining `id`, `title`, `status`, `ok`, `fields`, `manifest`, `onActivate`, `sfdtKind`, `T`, `summary`, `headline`, `detail`, … |
| rule 3 offenders | `[]` | **7, every one correct code** — `features/soql-runner.ts: setSfError(errorPanel, message, { doc })`, two `renderSfError(…)` in `schema-browser.ts`, four `showNlError`/`showError` funnels |
| newlines guard | `[]` | gains `features/org-health-checks.ts:159 (titleEl)` — a TITLE — and `features/soql-runner.ts:3180 (autocompleteTitle)` |

Every number in that table reproduces. `the struct-field funnel is open, and that is a MEASURED decision` also goes red under the widening, so the test genuinely pins the shape.

### NB3 — everything except B1

- The mask **did** fail open at base: `dynamicParts('/* never closed\npane.textContent = err.message;')` returns spaces at `c7547e7`.
- **The refusal is loud.** No consumer wraps `dynamicParts()` / `withoutComments()` in a `try`; the only `catch` is the deliberate one in the new test. B1's repro is the proof — a refusal takes six guard assertions red rather than being swallowed anywhere.
- **"Any diagnostic is the wrong trigger" is measured, and I get the same picture.** Instrumenting `blankLiterals()` across the whole extension suite (2,147 tests): **6,571 masked inputs**, **393** carrying ≥1 parse diagnostic (PR says 385; immaterial), top causes `';' expected` ×129, `Expression expected` ×103, `Argument expression expected` ×84. **Exactly 4 inputs** carry an unterminated literal or comment, and all four are the deliberate probes in `refuses to answer rather than blanking a file it could not read`. Zero legitimate inputs refuse **on today's tree** — see B1 for why that is not the same as the claim made.
- **Tree-wide parse health**: 0 diagnostics across all 125 scanned files, confirmed independently.
- The disclosed `CAUGHT → REFUSED` flip for the unterminated single-quoted string is louder-not-worse, and correctly disclosed. B1 is the case where `CAUGHT → REFUSED` **is** worse, and it is not disclosed.

### Cache bounding — eviction cannot change a verdict

Tested as a property, not by reading `MaskCache`. For all 125 files, compute the mask warm; then between every subsequent call, flood each cache with >4 MB of unique buffers to force `clear()`; compare.

```
EVICTION_DIFFERS (dynamicParts)    0 / 125     (cold and warm-hit both checked)
EVICTION_DIFFERS (withoutComments) 0 / 125
```

The policy is a pure-function memo with a clear-on-budget eviction, so same input → same answer regardless of cache state. Confirmed. (Note in passing: a single entry larger than the 4 MB budget degenerates the cache to one live entry rather than leaking — still bounded.)

### NB5 — the sample, and the `ts.createScanner` trap

- **4,899 usable comment positions** across the tree, exactly as claimed. I ran the **exhaustive** version myself: **4,899 probes, 0 failures, 0 throws** at head (4,888 / 0 at base — the difference is the comments this PR adds).
- The shipped sample is **917**, and it is **not cherry-picked**: the positions are chosen by index (`round(i·(n−1)/7)`), independent of outcome, and the sample includes both the first and the last comment of the file in **124 of 125** files. Since the exhaustive set has zero failures there is nothing to select toward in any case.
- **The `ts.createScanner` trap is real, and worse than described.** On `ui/workspace-host.ts` a bare scanner reports **39** comments where the comment-range API reports **158**, and **23 of the scanner's 39 are positions the API does not report at all** — including CSS comments inside the stylesheet template (`/* Header — a fixed 64px app bar…`, `/* Pill-shaped search, icon inset. */`). Using the trivia API instead was the right call.

### NB4 / NB6 — the declines

`.catch(<identifier>)`: **0 sites** in the tree. Two-argument `.then`: **0**. `catch ({ … })`: **0**. `{ stack }`: **0**. The 12 array destructures in the tree are all `Promise.all` / `split(',')` / `records`, none error-carrying.

NB6's split, re-derived through `rendersErrorIntoConsole` rather than by reading:

```
CAUGHT  const { message } = err as Error
CAUGHT  const { message } = zq
CAUGHT  catch ({ message })
MISSED  catch ({ message: zz })
MISSED  const [first] = String(err).split('\n')
MISSED  const { stack } = err
```

Exactly as the body reports, including `catch ({ message })` — which neither prior round tested — and `const { stack } = err`, which neither prior round named.

### Golden principle #12 — both lists, both directions

`EXEMPT` is one member, `DEFINING_ARTIFACTS` is one member, `test/` is in neither guard's `SCANNED_DIRS` (`features`, `ui`, `entrypoints`, `lib`), so the new tests are not artifacts of the checks they exercise. All four #12 mutations reproduce (M9–M12 below).

### All 12 mutations reproduce

Each applied in place, the named suites run, then restored.

| # | mutation | red |
|---|---|---|
| M1 | settled-reason binding removed | `knows a rejected settlement's reason, which no catch clause binds` · `rules 2 and 3 know a rejected settlement's reason` |
| M2 | `reason` added unconditionally | `leaves the word 'reason' alone…` · `binds a bare callback reference too…` · `binds no phantom name…` · **`every element assigned an error message declares a white-space rule`** |
| M3 | `org-health-checks.ts` → `'sfdt-muted'` | `keeps the guidance line on its own line when a check's error is two lines` |
| M4 | `org-health.ts` → `'sfdt-muted'` | `keeps a snapshot summary on two lines when the CLI reports a two-line failure` |
| M5 | unterminated-LITERAL refusal removed | `refuses to answer rather than blanking a file it could not read` |
| M6 | unterminated-COMMENT refusal removed | *(same test)* |
| M7 | `PROBES_PER_FILE` → 1 | `cannot be blinded by one stray backtick, in any file in the tree` |
| M8 | `CATCH_BINDING`'s `\(?` made mandatory | the 6 named + **`knows an error by its BINDING here too…`** |
| M9 | decorative `ui/toast.ts` in `EXEMPT` | `every exemption would actually be flagged without it` |
| M10 | the real `EXEMPT` entry broken | `every element assigned…` · `every exemption still matches real source` · `every exemption would actually be flagged without it` |
| M11 | decorative `lib/icons.ts` in `DEFINING_ARTIFACTS` | `every exclusion would actually fail without it` |
| M12 | the real `DEFINING_ARTIFACTS` entry broken | `every exclusion would actually fail without it` · `no console pane is handed a caught error directly` · `no feature, entrypoint or lib module hand-rolls the error block` |

**12 of 12, each red on exactly its named test** (M2 and M8 red on one more each, as above).

---

# My injection numbers

37 probes (29 defect, 8 correct-code), each run against **both** hosts, counted only where the two agree — they agreed on all 37. Driven through the real exported matchers (`buildsSfErrorPanel`, `rendersErrorIntoConsole`, `passesStringifiedError`) plus a replica of the newlines guard's per-file scan.

| | `c7547e7` | `b710a54` |
|---|---|---|
| 29 defect probes | **8 CAUGHT / 21 MISSED / 0 REFUSED** | **16 CAUGHT / 9 MISSED / 4 REFUSED** |
| 8 correct-code probes | 8 declined, **0 false positives** | 7 declined, **1 false positive** (`OK3`, above) |

**The 12 base misses that head closes** are, independently of the PR's list: all ten settled-reason shapes (`String(r.reason)`, interpolated, `.toString()`, const hop, `!== 'fulfilled'`, `.filter().map()`, `const { reason } = r`, the Prettier wrap, bare `r.reason`, and rule 3 via `renderSfError`), plus the unterminated block comment and the unterminated template — both `MISSED → REFUSED`. The same 12 the body claims.

**The four REFUSED are not counted as catches.** Three are the disclosed fail-open shapes. The fourth is B1's line-continuation string, which is `CAUGHT` at base and `REFUSED` at head on **legal** source.

Still CAUGHT at head, so the table is not read as better than it is: the class-getter funnel, an arrow returning the flatten, `let` assigned in a `catch`, `let` assigned inside `.catch((e) => …)`, a backtick in a block comment, a backtick in a trailing comment.

---

# Gates — my numbers

Both sides in one worktree, one `npm ci`, `packages/flow-core` built once, `wxt prepare` re-run per side.

| | `c7547e7` | `b710a54` |
|---|---|---|
| `extension/ npx vitest run` | 115 files / **2137** passed, 42.61s | 115 files / **2147** passed, 44.16s |
| `extension/ npx tsc --noEmit` | `exit=0` | `exit=0` |
| `extension/ npx eslint .` | `exit=0` | `exit=0` |
| `extension/ npx wxt build` | `Σ Total size: 1.4 MB` | `Σ Total size: 1.4 MB` |
| built manifest `sha256` | `2a302f6a…9c7d` | `2a302f6a…9c7d` (identical) |
| root `npm test` | 319 files / **5493** passed | 319 files / **5503** passed |
| root `npm run generate:catalogs` | — | no drift (`git status --porcelain generated/` empty) |

**Every number in the PR's gate table reproduces exactly.**

### The bundle delta — the claim this round got right

- `content-scripts/content.js` **633,460 → 633,478 = +18**; occurrences of `sfdt-muted sfdt-msg` go **5 → 7**. 2 × 9 characters.
- `chunks/workspace-host-*.js` **189,917 → 189,935 = +18**; `sfdt-muted sfdt-msg` **3 → 5**. First byte difference: ``s.className=`sfdt-muted`,s.textContent=n.summary`` → ``s.className=`sfdt-muted sfdt-msg`,…``.
- `chunks/app-*.js` **818 → 818** and `chunks/sidepanel-*.js` **1,181 → 1,181** — byte-length identical; the content hash moves only because the imported chunk filename changed (`workspace-host-So0zjZwh.js` → `workspace-host-Dy2N3svI.js`, same length). Read out of the built chunk.
- `app.html` / `sidepanel.html` differ only in that `src`. `manifest.json` byte-identical.
- `git diff c7547e7..b710a54 -- generated/ FEATURES.json package.json package-lock.json extension/package.json` is empty.

### Prettier

Never ran `--write`. Piping each changed file through `prettier --stdin-filepath` and diffing against itself, the hunk counts are **identical on both sides** for every file:

```
CHANGELOG.md 57/57 · features/org-health-checks.ts 8/8 · features/org-health.ts 6/6
test/org-health-checks.test.ts 4/4 · test/org-health.test.ts 8/8
test/error-source-scan.ts 0/0 · test/error-source-scan.test.ts 0/0
test/error-render-newlines.test.ts 0/0 · test/sf-error-panel-contract.test.ts 0/0
```

`org-health-checks.ts`'s hunk count stays at 8 while the diff-line count grows 84 → 96, which is the pre-existing whole-function re-indent now containing the added comment lines — exactly as the body says.

---

# What I could not verify

- **NB2 is still a DOM fact, not a browser.** happy-dom only, sixth round. `.sfdt-msg` declares `white-space: pre-line` and the sheet is adopted by the shadow host and every entrypoint; whether Chrome then puts the guidance on its own line follows from CSS and was not observed. No live org.
- **Whether B1's line-continuation string would ever be written.** It is legal, unlinted and untyped-checked here, and `prettier` preserves it — but it is not idiomatic modern TS and it is absent from the tree. The mechanism is a measured fact; the likelihood is a judgement, and it is the same judgement shape as round 5's unbalanced-backtick question.
- **Whether B1 is the only shape that manufactures an unterminated literal.** I reasoned it is (blanking preserves delimiters and newlines, so only an escape that consumes a newline can break a literal; templates span newlines legally and regexes are blanked whole) and found no second one, but I did not enumerate exhaustively.
- **Whether the two `allSettled` markers are exhaustive.** They are not — my `switch (s.status)`, imported-type-guard and destructured-`status` probes all miss, and none occurs in the tree in an error-carrying position today. I did not build a fix.
- **The 393 vs 385 diagnostic count.** Mine is instrumented at `blankLiterals()`'s cache-miss path; the difference is almost certainly instrumentation placement plus the four deliberate probes. I did not chase it.
- **The injections I did not think of.** 37 probes across two fresh hosts is fewer than round 4's 102 and round 6's 44; the budget went into the exhaustive 4,899-position property, the eviction property, the offset-correspondence property, rebuilding the declined NB2 widening, and the base/head gate matrix. Properties over the whole tree are worth more here than probe count — but the for-of category above is where I expect round 7 to find something, and it is a category, not a spelling.
- **Anything outside the sweep.** The other 5,503 root and 2,147 extension tests pass on both sides; I did not review what they assert.

---

Five bad rounds is not evidence the sixth is bad. This one is very close: NB1's scoping is exactly as narrow as claimed and I could not break its offset coupling, the cache eviction is a genuine no-op on verdicts, the NB2 decline was really built and really measured, the sampling is honest, and every gate and bundle byte reproduces. The one thing that does not hold is the general claim under the fail-closed change — and it took a three-line legal file to falsify it.

---

scoobydrew83 (commented):
**APPROVE** — B1 is fixed, and I could not break the fix. Both corrections you made to my round-6 list are right and I withdraw them. Ship it.

> Filed as a `COMMENT` because GitHub refuses `REQUEST_CHANGES` on a same-account PR. Line 1 is the verdict, per this series' convention.

Re-review at `ad199f0`, base `origin/develop` `c7547e7`, scoped to the delta from `b710a54` plus what the delta could have broken. Own worktree under `.claude/worktrees/review-330b`, nothing pushed, `FEATURES.json` untouched, all probe files deleted, tree clean at exit (`git status --porcelain` empty, `115 passed` / `2150 passed`).

I did not read the fix and find it reasonable. I built the leak you asked me to look for, and it is not there.

---

# The B1 fix — attacked, holds

## The oracle: no byte reaches code position

A masker that now *conditionally preserves* a character it used to always blank has exactly one new failure mode, so I asserted its absence as a property rather than arguing about it. Over **394 inputs** — all 125 scanned files, each again with an LF continuation spliced in, each again with a CRLF continuation spliced in, plus my own 19-member legal-but-hostile corpus:

```
ORACLE inputs=394  nonLiteralTokens=221730  failures=0
  a) changed OUTSIDE a literal interior or comment:  0
  b) code token altered:                             0
  d) a backslash surviving in CODE position:         0
  refused a legal input:                             0
  length / line-count drift:                         0
```

(d) is the one that matters: I parse the **masked** buffer, compute its own literal interiors and comment ranges, and assert every surviving `\` is inside one. **221,730 non-literal tokens survive byte-for-byte** and not one backslash lands outside a literal. The leak does not exist.

## My own hostile corpus — 19 members, legality checked against the compiler first

Built independently of yours and aimed squarely at the new conditional branch. Every legal member **ACCEPTED + idempotent**; every illegal member **REFUSED**:

| construct | diags | mask |
|---|---|---|
| continuation LF · CRLF · **lone CR** · **U+2028** · **U+2029** | 0 | ACCEPTED + idempotent |
| continuation at the end of an interior · as the whole interior | 0 | ACCEPTED + idempotent |
| **TemplateHead** · **TemplateMiddle** · **TemplateTail** continuation | 0 | ACCEPTED + idempotent |
| a continued string inside a template hole | 0 | ACCEPTED + idempotent |
| 3-backslash run then LF (escape + continuation) · 5-backslash run | 0 | ACCEPTED + idempotent |
| continuation followed by `//` · `/*` · `\'` · a backtick, all inside the string | 0 | ACCEPTED + idempotent |
| continuation in a JSDoc line · in a line comment | 0 | ACCEPTED + idempotent |
| **escaped backslash then a real LF** | 3 | **REFUSED** |
| 2-backslash run then a real LF | 2 | **REFUSED** |
| 4-backslash run then a real LF | 3 | **REFUSED** |
| bare LF in a string | 3 | **REFUSED** |

The lone-CR and U+2028/U+2029 cases are the ones I expected to break it — `k += next === '\r' && out[k + 2] === '\n' ? 2 : 1` plus the loop's own `k++` gets all three right. **The escaped-backslash claim is confirmed directly**, on four separate backslash-run lengths.

## The regex claim

Confirmed, and weaker than it needs to be — in your favour:

```
dynamicParts("const r = /['\\/`]/g;\nconst t = 1;\n")  →  "const r =           ;\nconst t = 1;\n"
```

Blanked whole, no backslash survives. But a `RegularExpressionLiteral` **cannot contain a line terminator** — an unterminated regex is refused before blanking — so `blankLiteralInto()` on a regex could never meet a continuation in the first place. See N2 below: removing the carve-out entirely changes nothing.

## The three-line demonstration, re-run

Same legal file, `npx eslint` exit 0, `npx tsc --noEmit` exit 0:

| | `c7547e7` | `b710a54` | `ad199f0` |
|---|---|---|---|
| the three guard suites, with `features/zzz-probe-lc.ts` present | 52 passed (52) | **6 failed / 54 passed (60)** | **63 passed (63)** |

## The property is not vacuous, and not a measurement in a new coat

This is the part I most wanted to break, because my B1 was found precisely by noticing that "measured across the suite: zero occurrences" was a fact about that day's tree wearing the clothes of a property. Four mutations:

| mutation | result |
|---|---|
| **V1** `blankLiteralInto` reverted to `blankInto` | `masking its own output is always accepted, and always a no-op` · `a line continuation in the source does not take the rules down` — **2 red** |
| **V1b** same, with the **splice removed** — corpus alone | **still 2 red** |
| **V1c** same, with the **corpus check removed** — splice alone | **still 2 red** |
| **V2** one corpus member made illegal (`'a\⏎ b'` → a bare newline) | **red**, on the legality assertion |

**Both halves are independently load-bearing**, and the corpus catches the regression with no tree involvement at all. That is a property, not a measurement. V2 shows the corpus cannot rot into a set of illegal inputs that pass by being refused — which is the failure mode the `EXEMPT` bite check exists for, applied here.

---

# The two corrections to my list — you are right on both

### 1. `switch (s.status)` — substantively right, and I withdraw the category

Zero **settled-shaped** switches in the tree: I searched for `switch (<ident>.status)` followed by `case 'rejected'|'fulfilled'` within 200 characters and got **0**. My round-6 listing implied the shape occurs in a relevant form and it does not.

One precision note so the number is on the record: the bare syntactic shape does occur once — `features/soql-runner.ts:1504`, `switch (outcome.status) {`, a domain status on an NL-generation outcome, nothing to do with a settlement. So `switch (<x>.status)` is 1 and settled-shaped is 0. The category stays open as a marker gap (my probe still MISSES it) but it has no tree presence, which is what you claimed.

### 2. `Promise.any` / AggregateError — you are right and I was wrong

I withdraw it. Measured at `ad199f0`:

```
Promise.any → catch (zqAgg) → String(zqAgg.errors[0])                  CAUGHT
Promise.any → catch (zqAgg) → String(zqAgg)                            CAUGHT
Promise.any → catch (zqAgg) → for (const sub of zqAgg.errors) …        MISSED
try/catch   → for (const zl of String(zq).split('\n')) …               MISSED
```

My probe missed **only** because I routed it through a `for-of`. It was the for-of finding wearing an AggregateError costume, not an independent category, and recording it as unconfirmed rather than claiming it was the correct call.

### for-of counts — reconciled, and immaterial

My numbers, on the masked buffer over the same 125 files:

```
for ((const|let|var) IDENT of …)   265        for (var IDENT of …)        0
for ((const|let) IDENT of …)       265        for await (… of …)          0
all binder forms                   303        destructuring binders      38
for-in, any binder                   0        for (IDENT of …) no binder  0
```

You report 263 / 301 / 34. I cannot account for the 2 and the 4 — there are no `var` and no `for await` heads to explain it, so it is regex boundary detail. The decision does not turn on it: it is 260-plus either way, an order of magnitude above anything this guard has closed, and `the for-of binding is open, and it is the LARGEST named gap` pins it. **V7** (teaching the declaration regex `\bof\b`) turns that test red, so the decision is a decision and not a comment.

---

# The other four

### NB5 — my round-6 finding inverts, and you are right

I measured it. Tree-wide comment composition:

```
line = 4585    jsdoc = 585    plain block = 0
```

You report 4,522 / 579 / **0**. My counts run slightly higher because my walk descends into token trivia as well as node trivia; **the load-bearing number is identical and it is zero.** `block=0` was a fact about the codebase, not a sampling failure, and my round-6 note that it contradicted the claim had the causality backwards. What was actually wrong was the test's own comment, and you fixed that. Stratifying anyway was still worth doing: **V6** (reverting to one flat pool) turns `cannot be blinded by one stray backtick, in any file in the tree` red, so the per-file coverage property bites.

### Prefix tightening — pinned, and it works end to end

**V3** (dropping the closing quote from the comparison) → `knows a rejected settlement's reason, which no catch clause binds` red. And in my injection set, the new correct-code probe `if (zqDeploy.status === 'rejectedByAdmin') zqPane.textContent = zqPlan.reason;` is **declined at `ad199f0`** where it was a false positive at `b710a54`. The unbounded half of the over-claim is gone; what is left is the file-scoped half you own.

### OK3 in the table — the right call

Owning **1 false positive** beats narrowing it here. I checked the alternative: making `carriesAnError()` receiver-precise is a second structural change to the binding scan, in the same round as a masker rewrite and a blocking-regression fix — which is the exact stacking that opened the next hole in rounds 2, 3 and 4, and which round 6 demonstrated again by shipping a regression while doing one thing at a time. A documented false positive that produces a site to look at is the cheap direction. Reporting `0` while the comment at the definition said otherwise was the only thing wrong with it, and that is fixed.

### Mutation rows

**M8 = 9** across all three suites at `ad199f0` — confirmed exactly, including the two new tests (`a line continuation in the source does not take the rules down`, `the for-of binding is open…`) joining the seven I measured at `b710a54`.

**M2 = 5, not 4.** The extra is `knows a rejected settlement's reason, which no catch clause binds`, which now also asserts `rejectedByAdmin` is *not* scoped, so unconditional `reason` breaks it too. Over-reproduction, in your favour; the table is one short.

---

# Non-blocking, new at `ad199f0`

### N1 — the escape-pair skip is unpinned

**V4**: delete the `k += 1` skip in `blankLiteralInto()`'s escape branch, so an escaped character is no longer stepped over — **63 passed (63)**, nothing notices.

I believe it is genuinely unobservable rather than untested-and-broken: the case it guards (`'a\\⏎`, where the second backslash would be misread as opening a continuation) is refused by `isUnterminated` before blanking ever runs, and inside a template — where `\\⏎` is legal — both readings produce an equally legal, equally idempotent buffer. So it is correct defensive code. But the PR presents escape tracking as load-bearing to the fix, and nothing in the suite would notice its removal.

### N2 — the regex carve-out is unpinned and unreachable

**V5**: remove the carve-out and run `blankLiteralInto()` on regex literals too — **63 passed (63)**. A regex cannot contain a line terminator, so the branch can never encounter a continuation and behaves identically to `blankInto`. The stated justification — "a preserved backslash would land in code position" — is right about the consequence and reads as though the case were live. Worth one word saying it is defence against a shape the grammar forbids.

### N3 — "diff of the full measurement output empty" is slightly stronger than what is true

Every aggregate is identical to `c7547e7`, verified by me:

| | `c7547e7` | `ad199f0` |
|---|---|---|
| distinct binding names tree-wide | 11 | **11, identical set** |
| `carriesAnError` sites, `features/` + `ui/` | 12 | **12, identical list** |
| rule 1 / 2 / 3 offender sets | `[]` | `[]` |

But a **per-file** dump is not empty — `features/soql-bulk-delete.ts` goes `err` → `err,reason`. That is the over-claim the PR documents, so it is disclosed; the sentence should say "the guard's reported output" rather than "the full measurement output". My full measurement at `ad199f0` is **byte-identical to `b710a54`**, and differs from `c7547e7` in exactly four places, all expected: that one binding line, the settled-scoped file list, the exhaustive probe count (4,888 → 4,899, from the comments this PR adds), and the settled-destructure probe.

---

# Injection — three sides, my probes

42 probes (32 defect, 10 correct-code), each run against both hosts (`features/schema-browser.ts`, `features/trace-flags.ts`), counted only where the two agree — they agreed on all 42.

| | `c7547e7` | `b710a54` | `ad199f0` |
|---|---|---|---|
| defect probes | 8 CAUGHT / 21 MISSED / 0 REFUSED | 16 CAUGHT / 9 MISSED / 4 REFUSED | **20 CAUGHT / 9 MISSED / 3 REFUSED** |
| correct-code probes | 0 false positives | 1 false positive | **1 false positive** (OK3, owned) |

**The 3 REFUSED at head, checked individually — every one is a genuinely unterminated construct**, and none is a second B1:

```
X1  /* never closed …            REFUSED   (unterminated block comment)
X2  const zqT = `never closed;   REFUSED   (unterminated template)
X3  const zqS = 'never closed;   REFUSED   (unterminated string)
X4  const zqS = 'a\⏎ b';         CAUGHT    ← legal; REFUSED at b710a54
X5  const zqS = 'a\␍␊ b';        CAUGHT    ← legal, CRLF
X6  const zqT = `a\⏎ b`;         CAUGHT    ← legal, template
```

The three additional CAUGHTs versus `b710a54` are exactly the three legal continuations. The 9 remaining MISSES are all MISSED at `c7547e7` too — no regressions — and are the marker gaps and the for-of / callback-destructuring categories, all named.

---

# Gates — my numbers

| | `c7547e7` | `b710a54` | `ad199f0` |
|---|---|---|---|
| `extension/ npx vitest run` | 2137 passed | 2147 passed | **2150 passed** (115 files, 45.95s) |
| `extension/ npx tsc --noEmit` | 0 | 0 | **0** |
| `extension/ npx eslint .` | 0 | 0 | **0** |
| `extension/ npx wxt build` | 1.4 MB | 1.4 MB | **1.4 MB** |
| built manifest `sha256` | `2a302f6a…9c7d` | same | **same** |
| root `npm test` | 5493 passed | 5503 passed | **5506 passed** (319 files) |
| `check:all-contracts` | 0 | 0 | **0** |
| `generate:catalogs` | — | no drift | **no drift** |

**Bundle: byte-identical to `b710a54`.** All 20 output files match on sha256 and size, so the `+18 / +18` against `c7547e7` stands unchanged and this round adds nothing to the shipped extension — as it should, since the whole delta is under `test/` plus one changelog line.

**Prettier**: `error-source-scan.ts`, `error-source-scan.test.ts` and `sf-error-panel-contract.test.ts` are 0 hunks on both sides; `CHANGELOG.md` is 57/57. No new drift.

---

# What I could not verify

- **Whether U+2028/U+2029 or a lone CR would ever appear in this tree.** I constructed all three and the mask handles them; none occurs today, and `commentRanges()`'s `//` branch still terminates a line comment at `\n` only, so a line comment ended by U+2028 would over-blank. That is pre-existing on `develop`, outside this delta, and I did not chase it.
- **Whether N1's escape-pair skip is unobservable or merely untested.** I argued it from the grammar and could not construct an input that distinguishes it; that is weaker than a measurement.
- **The 263 vs 265 for-of count.** Reconciled to "regex boundary detail" without finding the two.
- **No browser, no live org.** Sixth round.
- **The 12 round-6 mutations.** I re-ran only M2 and M8, per scope; the other ten I verified at `b710a54` and the delta does not touch them.
- **The injections I did not think of.** 42 probes across two hosts, plus the 394-input / 221,730-token oracle. As last round, I trust the property far more than the probe count.

---

I blocked round 6 on a false alarm the fix introduced, and the response was to make the thing a property with a corpus and a splice that each independently catch the regression, and to correct two of my own findings with measurements rather than argument. Both corrections are right. The one thing left on my list — the for-of binding — is now a named, counted, mutation-pinned decision rather than a surprise. That is what closing a round looks like. Ship it.

---

PR #332, opened 2026-08-06
fix(extension): a green tick after Cancel, and a sweep that catches the next caller (C-FIX-5)

scoobydrew83:
**Item:** C-FIX-5 · the tracked follow-ups from the C-P4-2 bulk-delete review (#326, merged) · Chrome ext
**Base:** `develop` @ `84e6db8`. `FEATURES.json` untouched. No docs-site change is needed — every wording change here either replaces prose that the docs already state correctly, or has no docs-site surface (see §5).

These were raised by round 2's independent review, which declined to block on them and asked they be tracked somewhere that is not the CHANGELOG, on the grounds that "a CHANGELOG sentence is not a tracked fix; it is user-facing prose that scrolls away at release."

**Round 2 of this PR** (`2d943a4`) fixes the one blocking finding from the review of round 1 — the sweep's `await` condition was a formatting fact wearing a control-flow fact's clothes, and went quiet on a behaviour-preserving refactor of the exact regression it exists to catch. §2 is rewritten around it; everything else below is unchanged except the two body corrections the review asked for, marked **[corrected]**.

---

## R2 — BLOCKING: the sweep missed the most likely form of the regression · FIXED

The reviewer regressed B1 on the real `soql-runner.ts` — moving `btn.disabled = true` out of `onConfirmed` back to click time — in two forms. The inline form fired the sweep. The **hoisted** form, a behaviour-preserving refactor that moves the callback to a `const` on the line above, left the sweep **green**, caught only by the three behavioural tests that happen to exist on that one surface. On the next surface, which by construction has no focus test yet, nothing would have caught it.

**Cause, as diagnosed.** The enclosing case required the dialog call to sit lexically inside the `await` expression's TEXT RANGE (`a.start <= call.at && call.at < a.end`). Hoisting moves the text out of the range while the callback still runs during the await.

**Fix — the reviewer's condition, taken verbatim:**

```
scan.awaits.some((a) => a.inFn === from && a.start > disable.at)
```

This is the control-flow fact the header always argued for: *this body disabled a control and then suspended on something*. Both halves are now pinned, and the second half was not:

| Mutation | Result |
| --- | --- |
| **M-J** back to the await text range (the reported hole) | CAUGHT — `detects the B1 shape`, which now carries the inline form **and two hoisted forms** (the deps object hoisted, and the callback itself hoisted) |
| **M-K** drop the `> disable.at` half, leaving "awaits anything" | **SURVIVED on first run** — nothing in the file declined a body whose `await` precedes the disable. Fixed by adding that shape (an async render function that fetches, then disables, then registers a listener) to `declines a listener registered in the same body`, plus its mirror with the await moved *after* the disable, which must be caught. M-K now fails. |
| **M-C2** drop the await condition entirely | CAUGHT (3) — incl. the tree-wide test |

**Proof the fix is what closes it, not something else.** With the hoisted regression live on the real `soql-runner.ts` **and** the condition reverted to the text range, the tree-wide sweep test passes — the reported hole reproduced exactly. With the new condition it fails. Against the shipped tree, both regression forms on the real file now fail the sweep plus the three behavioural tests (4 failures each).

**Offender set re-measured on both sides — unchanged in all three shapes:**

```
no await condition,   develop-shape debug-log-viewer  → 2  (debug-log-viewer:484, flow-version-manager:187)
new await condition,  develop-shape debug-log-viewer  → 1  (debug-log-viewer:484)
new await condition,  as shipped                      → 0
```

`flow-version-manager.ts:187` is still declined, now for the reason that actually applies rather than a lexical accident: `ensureToolbarButton()` awaits nothing at all.

**What it over-approximates, named in the file rather than discovered later:** a body that disables a control, awaits something unrelated, and only *then* registers a listener would be flagged. That fires in the safe direction — a site to look at, not a rule that goes quiet — and there is no such site in the tree.

### R2 — the two secondary spellings: caught, not named

`el['disabled'] = true` and `el.setAttribute('disabled', …)`. Both are caught rather than named as open, which is only defensible because the false-positive cost is a measured **zero**: there is no occurrence of either in the 125 scanned files, so widening the scan moved no verdict. `the two rarer disable spellings cost nothing, and that is measured` asserts both halves — the tree-wide zero, and that the scan really does catch them (so the zero is a measurement, not a second way of saying it is blind). The AST↔mask agreement check now counts **all three** spellings, so widening the scan cannot quietly narrow the agreement. **M-L** (stop catching the bracket form) and **M-M** (stop catching `setAttribute`) both fail it.

`btn.disabled = busy` stays out and is named: it is a state sync — `flow-version-manager.ts` writes `toolbarBtn.disabled = count === 0` on every repaint — and claiming it would flag correct code across the tree.

---

## 1. The time-sensitive one — a green tick after someone presses Cancel · FIXED

`reportDeleteOutcome` took the else branch for a cancelled run with zero failures: a **green** `sfdt-toast--success` reading *Deleted 25 Account record(s)*, **no report panel at all**, and `recordActivity` logging `status: 'success'`. So round 1's N6 — `formatBulkDeleteReport`'s *"Canceled before the remaining rows were attempted."* being unreachable in production — did not fall out when the signal was wired. It moved.

The branch is now on `outcome.canceled`, before the failure count:

- the toast is `warning`, reading `Stopped after 25 of 60 Account record(s)` (`· N failed` when there were failures);
- the report **is** shown, so the canceled line is reachable in production for the first time;
- `recordActivity` records `action: 'Bulk delete (stopped)'` and `status: 'failed'`. `ActivityStatus` is a two-value union (`'success' | 'failed'`) and widening it is a schema change in `lib/activity-log.ts` I did not want inside this item — of the two, a run the user stopped part-way is not the one that means "did what you asked", and the action text carries the nuance.

```
M1  outcome.canceled branch short-circuited to false
    × warns rather than congratulating, and shows the report it used to hide
M2  status reverts to `failures.length > 0 ? 'failed' : 'success'`
    × does not record the cancelled run as a success
```

## 2. The structural one — the contract sweep · SHIPPED

`test/confirm-dialog-trigger-contract.test.ts`, in the shape of `test/sf-error-panel-contract.test.ts`.

**The rule.** Flag a `.disabled = true` that lexically **precedes** a `confirmDialog(…)` call reached from the same function body. Three things it is careful about:

| | Why it is in the rule |
| --- | --- |
| **Order** | The correct fix writes the same two tokens the other way round (`await confirmDialog(…)` then `if (!ok) return;` then `btn.disabled = true`). A rule that ignored position would flag the fix it is asking for. |
| **Nesting direction** | `soql-runner.ts` hands `runBulkDelete` two SIBLING callbacks — `confirm:` opens the dialog, `onConfirmed:` disables the trigger — and an object literal has no meaningful order, so both arrangements must be clean. A disable in the **enclosing** body with the dialog down in a callback is the B1 defect exactly, so an enclosing function counts and a sibling does not. |
| **`await`** | The enclosing case additionally requires the body to `await` something **after** the disable. See R2 above for what this condition was, why it was wrong, and what it is now. |

**Why the compiler, and why the mask is still there.** All three questions are about SCOPE, which is the thing a regex cannot answer; hand-rolled brace matching is the move that has holed the sibling guard three times, twice inside the machinery a fix itself added. `ts.createSourceFile()` does not guess. What the AST alone does not give is a **fail-closed** answer, so `dynamicParts()` from `test/error-source-scan.ts` is called first — it refuses, loudly, on an unterminated literal or block comment, which is the shape that makes a scanner blank a region and go silently green. I did not touch that module (C-FIX-6 owns it); it is imported and reused. Its output is also **cross-checked against the AST's findings on every scanned file**, so a blinded walk in either mechanism goes red rather than quiet.

**Golden principle #12: there are no exclusions, and that is a result.** `ui/confirm-dialog.ts` writes the banned shape verbatim in its own caller contract — `btn.disabled = true;` on one line, `await confirmDialog({ … });` on the next — in a doc comment, and it is the file the whole rule is about. Through the AST that comment is not code and through the mask it is blanked prose, so the one file that would plausibly need an exemption needs none. `the dialog's own caller contract is prose, not a violation` pins that against the real file; `there is nothing to exempt` pins the empty list; and `every exclusion would actually fail without it` is present and vacuous by construction, so the first entry has to earn itself.

### Mutation evidence — the sweep

Every hole was CONSTRUCTED and watched, never read out of the regex.

| Mutation | Result |
| --- | --- |
| **M-C2** drop the await condition entirely | CAUGHT (3) |
| **M-J** the await TEXT RANGE instead of the control-flow fact | CAUGHT (2) — the R2 blocking finding |
| **M-K** drop the after-the-disable half | CAUGHT (1) — after the shape it declines was added; it survived before that |
| **M-D** ignore lexical order | CAUGHT — `declines the correct order — disabled AFTER the dialog answers` |
| **M-E** same-body only, drop the enclosing case | CAUGHT (2) |
| **M-F** drop the local-wrapper fixed point | CAUGHT (3) |
| **M-G** find the sites by raw text instead of through the AST | CAUGHT (6) — incl. **`the dialog's own caller contract is prose, not a violation`** and the AST/mask agreement. Read as raw text, `ui/confirm-dialog.ts` is the loudest offender in the tree; this is the load-bearing proof that the masking is not decoration. |
| **M-H** drop the fail-closed `dynamicParts()` gate | CAUGHT — `refuses to answer on source no scanner can read` |
| **M-I** an AST walk that stops at arrow functions (a blinded parser) | CAUGHT (4) — including the AST/mask agreement, which is the only assertion that can see this class at all: the tree-wide rule stays **green** under it. |
| **M-L / M-M** stop catching the bracket form / `setAttribute('disabled', …)` | CAUGHT — `the two rarer disable spellings cost nothing, and that is measured` |
| **M-A2** revert `debug-log-viewer.ts` to the develop shape | CAUGHT (3) — the sweep plus both new behavioural tests |
| **B1-INLINE / B1-HOISTED** on the real `soql-runner.ts` | CAUGHT (4 each) — the sweep plus three behavioural tests, in **both** forms |

The declining side is measured too: sibling callbacks in **both** literal orders, the disable-after-the-dialog fix, the listener-in-the-same-body shape (against the real `flow-version-manager.ts`) in both a sync and an async body, and comments, strings, template literals and five backtick blinders each carrying the real defect beneath them, with a control case so they cannot pass for the wrong reason. Four shapes are named as open rather than patched: a disable hidden behind a helper, a computed value instead of the literal `true`, a callback in a body that awaits nothing, and a dialog reached through an imported wrapper — each with its nearest closed neighbour asserted.

## 3. The deferred one — `debug-log-viewer.ts` · FIXED, not exempted

The reviewer's condition was that it be tracked; the tracking row calls the sweep "making it visible instead of remembered", and fixing was preferred. It is fixed.

Moving one line was genuinely not available: `deleteAll()` disabled the button across an async `fetchAllLogIds()` that paginates the org's whole ApexLog table and needs a busy affordance. So the phases are separated, which is the same move `onConfirmed` was for the bulk delete:

- `countAllLogIds()` owns the busy state and **ends it before returning**, so nothing can open a dialog with the trigger disabled;
- `deleteLogs()` owns the destructive phase and disables there, past the confirm;
- re-entrancy is a `deleteInFlight` flag rather than `disabled`, for the reason the dialog's caller contract gives;
- and one thing the deferral write-up did not mention: **disabling the focused element hands focus to `document.body` in a real browser**, so the busy affordance would have re-created B1 by itself. `countAllLogIds` remembers focus against the same expression the dialog reads (`doc.activeElement === deleteBtn`) and hands it back — but only when the disable is what lost it, so a user who tabs away during the count is not yanked back.

happy-dom does not model blur-on-disable (and makes `blur()` a no-op once disabled), so the test does that step by hand and says so; without it that restore is unobservable here and the bug would exist in Chrome only.

```
M-A2  revert to the develop shape
      × no caller disables its trigger on the way into the dialog        (the sweep)
      × returns focus to the Delete button when the confirm is cancelled (C-FIX-5)
      × keeps the trigger busy across the count, and focusable once the dialog is up
M-B2  the focus restore never fires
      × keeps the trigger busy across the count, and focusable once the dialog is up
```

The two behavioural tests are deliberate: the sweep catches the SHAPE, and a shape rule can be satisfied by moving a disable into a helper without fixing anything.

## 4. Six mutation survivors — six assertions, each mutation-proved

| Survivor | Assertion | Mutation that now fails |
| --- | --- | --- |
| reusable for a second delete | `a second delete works once the first has finished` — and it refuses the first at the dialog, because even a run that deleted nothing must release the latch | **M9** `deleteInFlight = true` in the `finally` |
| failed and timed-out rows survive pruning | `leaves the failed and the timed-out rows in the table` — asserted against the table's own text, since the report panel names those Ids too | **M7** prune `deletedIds` ∪ `failures` |
| the trigger IS disabled once deleting starts | the existing test at `:297`, whose name promised it — see §5 | **M10** `onConfirmed` no longer disables |
| the Cancel control IS visible while deleting | `keeps the Cancel control visible while the delete is running`, sampled from inside `apiRequest` | **M3** `display = 'inline-block'` removed |
| a new query aborts a running delete | `a new query stops the delete, and says so somewhere that is not the table` | **M5** the abort call removed (25 becomes 60) |
| rows without an Id survive pruning | `leaves rows that carry no usable Id exactly where they were` — reachable because `planBulkDelete` SKIPS an unusable Id rather than refusing the set | **M8** prune filter inverted to `id !== null && …` |

## 5. Smaller items

**The tooltip (`soql-runner.ts`).** *"Downloads a backup CSV"* claimed the part the extension structurally cannot observe, on the one control whose entire safety story is the backup. Now: *"Generates a backup CSV of the 2 rows and hands it to your browser — SFDT cannot confirm it reached your disk — then asks you to type …"*, matching the dialog, `features.mdx` and `privacy.mdx` verbatim in claim. Pinned by `the toolbar tooltip claims only what the extension can observe`, which also asserts the old string is **gone**; **M13** restores it and the test fails. No docs-site change: a `title` attribute has no docs surface and the honest wording is already what the site says.

**`extension/CHANGELOG.md` duplicated sections.** `[Unreleased]` carried Added at 7 and 48, Fixed at 17 and 37, Changed at 32 and 43, with no blank line before the second Fixed. Merged into one section per type in the house order the released sections use (Added, Changed, Fixed). Bullet counts before and after: Added 7+1 = 8, Changed 3+2 = 5, Fixed 5+6+3 = 14 — 27 in, 27 out. One bullet is reworded (the #326 one that said `debug-log-viewer.ts:481` "is **not** fixed here", since it now is); the sweep bullet is further reworded in `2d943a4` to describe the corrected await condition. No section moves.

**The status line above the NEW result set.** After a delete aborted by a new query, `Deleted 25 of 60 · re-run the query to refresh the table` landed above rows it was not about, advising the user to do the thing they had just done — and `onProgress` could land a `Deleting… 25 of 60` on top of the new query's own count after that. The three stop reasons are now distinguished (`canceled`, `superseded`, `closed`) because `signal.aborted` says THAT a run stopped and never why. On `superseded` nothing is written against the table at all — no status line, no report panel (which `execute()` clears for exactly this reason), no pruning of a result set the Ids no longer belong to — and the user is told through a toast plus the activity log, so nothing goes unreported: *Delete stopped by the new query — 25 of 60 Account record(s) had already been deleted*. **M4** (report it like a user cancel) and **M6** (progress keeps writing) both fail the test.

**`test/soql-runner-bulk-delete.test.ts:297`** was named *"…and disables it only once deleting starts"* and asserted only the first half. Both halves now, with the second sampled at the first DELETE. **M10** fails it.

**`1 failed — 1 failed.`** The summary reads `${count} ${label}` under a lead-in that already gives the count, and `unknown`'s label was the word *failed*. It is `unclassified` now, and when that is the only kind the breakdown is dropped entirely. `2 failed — 1 timed out, 1 unclassified.` is still produced for a mixed set, asserted in the same test. **M11** and **M12** both fail it.

## Explicitly not reopened

`confirm-dialog`'s `trim()` and the backup-on-disk guarantee — both accepted by the reviewer, both untouched. The docs-site "44 features" collision was resolved by sfdt-site#48 and never shipped.

## Gates — actual output

Base is `develop` @ `84e6db8`, built and run separately in the same worktree.

| Gate | `develop` 84e6db8 | This PR 2d943a4 | Δ |
| --- | --- | --- | --- |
| ext `npx vitest run` | 115 files / 2150 pass | **116 / 2178 pass** | +1 file / +28 tests |
| ext `npx tsc --noEmit` | clean | **clean** | — |
| ext `npx eslint .` | clean | **clean** | — |
| ext `npx wxt build` | exit 0, Σ 1.4 MB | **exit 0, Σ 1.41 MB** | see below |
| root `npm test` | 319 / 5506 pass | **320 / 5534 pass** | +1 / +28 |
| root `npm run check:all-contracts` | exit 0 (all six) | **exit 0 (all six)** | — |
| root `npm run generate:catalogs` | no drift | **no drift** | — |

```
```

**Manifest sha256 — identical on both sides:**

```
84e6db8  2a302f6a1b97539b0857f162e24fb498774b602f8840b515d865ecc9e8359c7d
2d943a4  2a302f6a1b97539b0857f162e24fb498774b602f8840b515d865ecc9e8359c7d
```

No `package.json` or lockfile change anywhere in the diff. No new manifest permission and no new host.

**Bundle delta, accounted for:**

```
content-scripts/content.js   633.48 kB → 634.33 kB   (+0.85 kB)
chunks/node-graph-*.js       240.74 kB → 241.61 kB   (+0.87 kB)
```

Both carry the same three feature files, so the same delta appears twice. It is the new runtime code and nothing else — the `DeleteStop` wiring and the cancelled/superseded branches in `reportDeleteOutcome`, the longer tooltip string, the report's only-unclassified branch, and the `debug-log-viewer` split. **[corrected]** Counted per chunk with a binary-safe grep: `Delete stopped by the new query` ×1, `Counting logs` ×1, and `SFDT cannot confirm it reached your disk` **×2** — my new tooltip plus the pre-existing dialog copy, not ×1 as round 1 of this body said. The new test file is test-only and reaches no bundle; `typescript` was already an extension devDependency, which is what `npx tsc --noEmit` runs. Round 2 (`2d943a4`) touches only a test file and the CHANGELOG, so the bundle is byte-identical to round 1.

**Prettier. [corrected]** **Seven** already-unclean files, not six — `CHANGELOG.md` is one of them. All seven were already Prettier-unclean at `84e6db8`, and the complaint set is the same set before and after, verified by running `--check` against the base blobs. Nothing was `--write`-formatted. The one **new** file is Prettier-clean, matching its neighbours `test/sf-error-panel-contract.test.ts` and `test/error-source-scan.ts`, which both are.

## Coordination with C-FIX-6

`test/error-source-scan.ts` and the error-panel guard tests are **not modified** — read and imported only. The sweep depends on exactly one export, `dynamicParts`, and uses it for the fail-closed refusal and the tree-wide agreement check; it does not use `readExpression` or `withoutComments`, because the AST answers those questions directly. **No change is needed there for this PR.** One observation for that branch rather than a request: the `for-of` gap it is closing has no analogue here — this rule binds nothing and follows no names, so widening `errorBoundNames()` cannot change any verdict in `confirm-dialog-trigger-contract.test.ts`. `extension/CHANGELOG.md` is the only shared file, and this PR owns it.

## What I could not verify

1. **No delete has run against a real Salesforce org**, and no debug-log delete either. All of the above is happy-dom plus a mocked `SalesforceApiClient`.
2. **No unpacked-build pass in a real Chrome.** Manifest byte-identity and bundle string inspection only.
3. **Blur-on-disable is asserted as happy-dom plus a hand-written step**, not as Chrome. The `hadFocus` restore in `debug-log-viewer.ts` is correct for the specified browser behaviour and its test simulates that behaviour; it has not been watched in a browser.
4. **Real assistive-technology behaviour.** Every focus finding here is a DOM `activeElement` assertion; no screen reader was involved.
5. **The sweep's open shapes are named, not closed** — a disable behind a helper, a computed value, a callback in a body that awaits nothing, an imported wrapper. Each is asserted as open so the boundary cannot drift silently, but each is a real way to write the bug. The R2 finding is the reason to read that list as a live risk rather than a formality: the `await` condition looked principled and was not, and only a constructed refactor showed it.
6. **`ActivityStatus` remains two-valued.** A stopped run records `failed`, which is the closer of the two and not a precise description. Widening the union is a schema change and did not belong in this item.


scoobydrew83 (commented):
## Independent review — C-FIX-5 · **VERDICT: REQUEST CHANGES**

> Posted as a `COMMENT` review only because GitHub refuses `REQUEST_CHANGES` on a PR owned by the same account. **Treat this as REQUEST CHANGES** — there is **one** blocking finding, and it is a one-line fix I have already proven green.

Reviewed from my own detached worktree at `e22e34a`, with `develop` (`84e6db8`) built and run **in the same worktree** so both sides of every gate are mine. I did not review by reading: I reproduced all 21 of your mutations, ran 17 of my own probes against the new sweep, and constructed the B1 regression **on the real `soql-runner.ts`** twice — once the way you tested it, and once with a behaviour-preserving refactor.

Everything you claim is true except one thing, and that one thing is the thing this PR is for.

---

## The block

### B1 — the sweep does not fire on the most likely form of the regression it exists to catch

Round 2's ask was *"what would actually catch the next caller is a contract sweep"*. It catches the next caller only if the next caller writes the callback **inline**. Hoist it to a `const` — a refactor that changes no behaviour, changes no order, and is not in your named-open list — and the guard goes silent.

I regressed B1 on the real file, both ways:

```
MUTATION: btn.disabled = true moved out of onConfirmed and back to click time
          (round 2's "B1 regressed" mutation, on features/soql-runner.ts)

  (a) confirm: (plan, phrase) => confirmDialog({ … })      ← inline, as shipped
      × no caller disables its trigger on the way into the dialog        ← SWEEP FIRES
      × returns focus to the Delete button when the dialog is cancelled
      × returns focus to the Delete button when … dismissed with Escape
      × leaves the trigger enabled while the dialog is up, and disables it only once deleting starts

  (b) const askToConfirm = (plan, phrase) => confirmDialog({ … });
      …
      confirm: askToConfirm,                                ← hoisted, same behaviour
      × returns focus to the Delete button when the dialog is cancelled
      × returns focus to the Delete button when … dismissed with Escape
      × leaves the trigger enabled while the dialog is up, and disables it only once deleting starts
      ✓ no caller disables its trigger on the way into the dialog        ← SWEEP GREEN
```

In (b) only the three **behavioural** tests catch it — the ones that exist on this one surface because #326 wrote them. On the next surface, which by construction has no focus test, **nothing** catches it. That is the sweep's entire purpose, and it is the exact failure mode the header says this codebase cannot see: it fails green.

**Why.** The enclosing case requires the dialog call to sit *lexically inside the text range of an `await`*:

```ts
return scan.awaits.some((a) => a.inFn === from && a.start <= call.at && call.at < a.end);
```

The header argues for something weaker and correct — *"an awaited expression is evaluated NOW, which is the grammatical version of 'this callback runs as part of this function'"*. But `a.start <= call.at && call.at < a.end` is not that. It asks whether the callback's **source text** happens to be nested inside the `await` expression, which is a formatting fact, not a control-flow one. A `const` on the line above moves the text out of the range while the callback still runs during the await.

Reduced, so it can be re-checked in isolation — this is my probe P2, and it is declined:

```ts
async function start(btn) {
  btn.disabled = true;
  const ask = () => confirmDialog({ doc });
  await runBulkDelete(rows, { confirm: ask });
}
// disablesTriggerBeforeDialog(src) => []
```

**The narrowest fix, measured.** Ask whether the body genuinely suspends after the disable, rather than where the text sits:

```ts
return scan.awaits.some((a) => a.inFn === from && a.start > disable.at);
```

I ran this. It is **strictly better on every axis you measured**:

| | shipped condition | `a.start > disable.at` |
| --- | --- | --- |
| clean tree, 125 files | green | **green** |
| the sweep's own 16 tests | 16 pass | **16 pass** — incl. `declines a listener registered in the same body`, both sibling orders, `declines the correct order`, and all four named-open shapes |
| `flow-version-manager.ts:187` (the false positive the `await` condition exists for) | declined | **declined** — `ensureToolbarButton()` contains no `await` at all, so it never reaches the check |
| develop-shape offender set | `debug-log-viewer.ts:484` | **`debug-log-viewer.ts:484`** — identical |
| B1 regressed, callback inline | fires | **fires** |
| B1 regressed, callback hoisted | **misses** | **fires** — `features/soql-runner.ts: btn.disabled = true (line 1864) before confirmDialog(` |

So the `await` condition is not the narrowest fix for the false positive. It is a *convenient* one that also hides the real case, which is precisely the question I was asked to answer, and the answer is the uncomfortable one.

**Either** take the one-liner (16/16 still green, tree still clean, false positive still declined), **or** — if you want to keep the stricter reading for a reason I have not seen — add the hoisted-callback shape to `these are the shapes it does not close, and that is a decision` with its nearest closed neighbour, so the boundary is honest. Right now §2 says *"Four shapes are named as open … so 'open' describes a boundary rather than a rule that never fires"*, and there is a fifth, more reachable than at least two of the four.

**Two smaller members of the same family**, neither in the named list, both declined today (probes P5, P16):

```ts
btn.setAttribute('disabled', '');   await confirmDialog({ doc });   // => []
btn['disabled'] = true;             await confirmDialog({ doc });   // => []
```

Lower priority — I grepped `features/ ui/ lib/ entrypoints/` and **neither idiom appears anywhere in the tree**, so these are hypothetical in a way P2 is not. Naming them costs a line each.

---

## Everything else — confirmed

### Gates: re-measured on both sides, in one worktree

| Gate | `develop` 84e6db8 | PR e22e34a | claimed | mine |
| --- | --- | --- | --- | --- |
| ext `npx vitest run` | 115 files / 2150 | 116 / **2177** | 2150→2177, 116 files | ✅ exact |
| ext `npx tsc --noEmit` | clean | clean | clean | ✅ |
| ext `npx eslint .` | clean | clean | clean | ✅ |
| ext `npx wxt build` | exit 0, Σ 1.4 MB | exit 0, Σ **1.41 MB** | exit 0 | ✅ |
| root `npm test` | 319 / 5506 | 320 / **5533** | 5506→5533 | ✅ exact |
| root `check:all-contracts` | — | exit 0, all six | exit 0 | ✅ |
| root `generate:catalogs` | — | no drift | no drift | ✅ |

**Manifest sha256 identical on both sides** — `2a302f6a1b97539b0857f162e24fb498774b602f8840b515d865ecc9e8359c7d`. Permissions unchanged (`storage, clipboardWrite, cookies, contextMenus, sidePanel`), 10 host permissions, no new host. No `package.json` or lockfile change anywhere in the diff. `FEATURES.json` not in the diff.

**Bundle delta accounted for, independently:**

```
content-scripts/content.js   633.48 kB → 634.33 kB   (+0.85 kB)
chunks/node-graph-*.js       240.74 kB → 241.61 kB   (+0.87 kB)
```

Both chunks carry the same three feature files, so the same delta twice. String census in the PR build confirms the source of it, and confirms it is only these files:

```
"Delete stopped by the new query"           content=1  node-graph=1
"Counting logs"                             content=1  node-graph=1
"Bulk delete (stopped)"                     content=1  node-graph=1
"unclassified"                              content=1  node-graph=1
"SFDT cannot confirm it reached your disk"  content=2  node-graph=2
```

One correction: the last is **2**, not "exactly once" — one is the new tooltip, one is the pre-existing dialog copy from #326. The delta is still +1. Nothing else moved.

**Prettier.** Confirmed by running `--check` on both sides. The complaint set is byte-identical:

```
BASE == PR:  CHANGELOG.md, features/debug-log-viewer.ts, features/soql-bulk-delete.ts,
             features/soql-runner.ts, test/debug-log-viewer.test.ts,
             test/soql-bulk-delete.test.ts, test/soql-runner-bulk-delete.test.ts
```

Nothing was `--write`n. The new file is Prettier-clean. That is **seven** already-unclean files, not six — the PR body undercounts by one (`CHANGELOG.md`). Trivial.

### Coordination with #331 — clean

`test/error-source-scan.ts` and the error-panel guard tests are **not in this diff**. `extension/CHANGELOG.md` is the only file both branches touch, as stated. I also read #331's change to `error-source-scan.ts`: it is comment-only around `blankLiteralInto`/`blankLiterals`, no behavioural change to `dynamicParts`, so #332's import cannot be affected whichever lands first.

### The offender set — reproduced exactly

Not taken on trust. I ran the shipped rule against a `develop`-shape `debug-log-viewer.ts`, with and without the `await` condition:

```
develop shape, no await condition  →  2 offenders
    features/debug-log-viewer.ts:    deleteBtn.disabled = true (line 484) before confirmDialog(
    features/flow-version-manager.ts: btn.disabled = true (line 187) before handleBulkDelete(

develop shape, WITH await condition →  1 offender
    features/debug-log-viewer.ts:484

PR head, WITH await condition       →  0 offenders
```

**The false positive is genuine.** `ensureToolbarButton()` (`features/flow-version-manager.ts:179-191`) sets the initial state of a freshly created `<input type=button>`; `updateToolbar()` at `:109` sets `toolbarBtn.disabled = count === 0` on every refresh, so the trigger is enabled whenever the user can click it, and the handler runs in a different turn. Flagging it would mean editing correct code. Your reading is right; only the *fix* for it is too broad (B1 above).

### Sweep mutations — all reproduced

| Mutation | Claimed | Mine |
| --- | --- | --- |
| M-C drop the `await` condition | CAUGHT (3) | ✅ **3** — `no caller disables…`, `declines a listener registered in the same body`, `these are the shapes it does not close` |
| M-D ignore lexical order | CAUGHT | ✅ `declines the correct order — disabled AFTER the dialog answers` |
| M-E same-body only | CAUGHT | ✅ `detects the B1 shape — disabled in the body, dialog down in a callback` |
| M-F drop the wrapper fixed point | CAUGHT (2) | ✅ **2** — `finds every caller…`, `follows a local wrapper…` |
| M-G raw text instead of AST | CAUGHT (5) | ✅ **4** on my construction — incl. **`the dialog's own caller contract is prose, not a violation`** and the AST/mask agreement. Count differs because M-G is not a uniquely specified mutation; the load-bearing assertion fires either way, so the masking claim holds |
| M-H drop the fail-closed gate | CAUGHT | ✅ `refuses to answer on source no scanner can read` |
| M-I a walk that stops at arrow functions | CAUGHT (3) | ✅ **3** — and I confirm the specific claim: `no caller disables its trigger…` stays **green** under it; the AST↔mask agreement is the only thing that sees it |
| M-A2 revert `debug-log-viewer.ts` | CAUGHT (3) | ✅ **3** — the sweep plus both new behavioural tests |
| M-B2 focus restore never fires | CAUGHT (1) | ✅ `keeps the trigger busy across the count, and focusable once the dialog is up` |

M-G is the good one. Read as raw text `ui/confirm-dialog.ts` really is the loudest offender in the tree, and the AST + mask together make it prose. **The empty exclusion list is a genuine result.**

### Golden principle #12 — holds, and the scope is not gerrymandered

`SCANNED_DIRS = ['features','ui','entrypoints','lib']` is **byte-identical to the sibling guard's** (`sf-error-panel-contract.test.ts:97`), 125 files. `test/` is excluded for the same stated reason it is there. The only other `.ts` outside the set are `mockups/mock.ts` and `scripts/check-version-match.ts` — I grepped both for `confirmDialog` and `disabled = true`: **neither appears**, so nothing is hidden by the scope. The sweep does not flag its own checker (not scanned), its fixtures (inline strings), or its docs (`ui/confirm-dialog.ts`, verified in-scan and verified clean). Zero exclusions is true rather than arranged.

### Runtime mutations M1–M13 — all 13 reproduced, each failing exactly the named assertion

```
M1  canceled branch → false          × warns rather than congratulating, and shows the report it used to hide
M2  status reverts                   × does not record the cancelled run as a success
M3  Cancel never shown               × keeps the Cancel control visible while the delete is running
M4  superseded reported as a cancel  × a new query stops the delete, and says so somewhere that is not the table
M5  abort removed from runQuery      × …  "expected 60 to be 25"      ← your claimed number, verbatim
M6  progress keeps writing           × a new query stops the delete, and says so …
M7  prune deletedIds ∪ failures      × leaves the failed and the timed-out rows in the table
M8  prune filter inverted            × leaves rows that carry no usable Id exactly where they were
M9  deleteInFlight = true in finally × a second delete works once the first has finished
M10 onConfirmed no longer disables    × leaves the trigger enabled while the dialog is up, and disables it only once deleting starts
M11 unclassified → failed             × does not summarise an unclassified failure as "failed" twice
M12 drop the only-unclassified case   × does not summarise an unclassified failure as "failed" twice
M13 restore the old tooltip           × the toolbar tooltip claims only what the extension can observe
```

**The six survivors are genuinely dead.** M9, M7, M10, M3, M5, M8 each go red now, and each on the assertion written for it — not on an incidental one. `:297` in particular: the mutation that survived a whole round because the test's name promised something it did not check now fails that same test.

### §1 — the cancel affect, and `status: 'failed'`

I traced the consumers rather than accepting the reasoning. `ActivityStatus = 'success' | 'failed'` (`lib/activity-log.ts:39`), validated at `:88`. **Exactly one downstream reader exists:**

```
ui/workspace-host.ts:600  const pillClass = entry.status === 'success' ? 'sfdt-success' : 'sfdt-error';
ui/workspace-host.ts:601  const label     = entry.status === 'success' ? 'Success' : 'Failed';
```

That is the whole blast radius. **No filter, no count, no aggregation, no telemetry** reads the field — I grepped every `.status` read in `features/ ui/ lib/ entrypoints/` and everything else is a different `status` (HTTP, org-health bands, flow metadata, the options page's own CSS class).

So a user-initiated cancel renders as a red **Failed** pill next to the action text **"Bulk delete (stopped)"**. Imprecise, and you say so. But it does not corrupt anything, and the alternative is worse: a green **Success** pill for a run the user aborted at 25/60 would reintroduce, in the activity feed, the exact affect bug this PR removes from the toast — and it would contradict the warning toast written in the same function. Given a two-valued union, this is the right one. **Not a finding.** Your item 6 in "what I could not verify" is the honest framing and I would leave it there.

### §3 — `debug-log-viewer.ts`, and the focus claim

**No window exists where the destructive button is live during the count.** `deleteBtn.disabled = true` is the first statement of `countAllLogIds()`, synchronous, ahead of `await fetchAllLogIds()`. The re-enable is in that function's `finally`, so it is complete before the promise resolves and before `deleteAll()` can reach `confirmDialog`. The gap between the re-enable and the modal's scrim is a microtask, and `deleteInFlight` covers it regardless. This is the reason the fix was deferred, and the split genuinely closes it rather than trading it away.

**The focus claim is correct and it is load-bearing.** Disabling the focused element does hand focus away in a real browser — HTML's unfocusing steps run when a focused element stops being a focusable area, and focus lands on the body. So without the `hadFocus` restore the dialog would capture `activeElement === body`, restore to `body`, and the busy affordance would have re-created B1 on its own. M-B2 proves it is not decoration: removing only the restore turns `the count did not cost the trigger its focus` red. Good catch, and it is a catch — the deferral write-up in #326 did miss it.

The restore is correctly conditional (`hadFocus && activeElement is null-or-body`), so a user who tabs away during the count is not yanked back, and it is remembered against `doc.activeElement`, the same expression `confirm-dialog.ts:65` reads. Self-consistent.

### §5 — the smaller items

**The three stop reasons, and `superseded` writing nothing.** I walked every exit of `reportDeleteOutcome` asking your question — can a destructive operation's outcome now go unreported? **No.** On `superseded` the toast fires **and** `recordActivity` still runs; only the two table-anchored channels (status line, report panel) are suppressed, which is right, because `execute()` has already replaced the table and cleared the panel. Skipping the prune is right for the same reason.

The three early returns above it (`ineligible`, `backup-failed`, `not-confirmed`) each write `status.textContent = prevStatus` **unconditionally**, which under `superseded` would stamp the old table's status line over the new query's. I checked reachability rather than flagging it: `superseded` requires `deleteController`, set immediately before `runBulkDelete`, and everything from there to the dialog opening is microtasks (`backup` is synchronous), so no user input can land in it; once the dialog is up it is modal with a full-viewport scrim, and the only keyboard run path is `textarea`-bound (`soql-runner.ts:3362`) and therefore outside the focus trap. **Not reachable, so not a finding** — but it is the one place a future change (a non-modal confirm, a global run hotkey, a palette entry) would turn into the same bug, and a `stop === 'superseded'` guard on those three `prevStatus` writes would cost three tokens.

`close()` → `stop = 'closed'` takes the cancel branch and writes a report panel into a torn-down view plus a warning toast. Writing to a detached node is inert and the toast is correct information. Fine.

**CHANGELOG.** Verified by parsing both sides rather than counting by eye:

```
BASE [Unreleased]: headings [Added, Fixed, Changed, Fixed, Changed, Added]  →  27 bullets (Added 8, Fixed 14, Changed 5)
PR   [Unreleased]: headings [Added, Changed, Fixed]                         →  32 bullets (Added 9, Fixed 18, Changed 5)
```

Of the 27, **26 survive byte-identical, 0 changed section, 1 is reworded** — the `debug-log-viewer.ts:481` bullet, exactly as you say. Your body's "nothing dropped or reworded" and "the #326 bullet … is corrected" are two sentences in tension; the second is the true one and the rewrite is right (the old text now says the opposite of the truth). One section per type, house order. ✅

**The tooltip.** M13 confirms the old string is asserted gone, not merely replaced. I grepped `sfdt-site/content/` for `Downloads a backup` and for the tooltip text: **no hit**, so "a `title` attribute has no docs surface" checks out.

**`1 failed — 1 failed.`** M11 and M12 both red, and the mixed case `2 failed — 1 timed out, 1 unclassified.` is asserted in the same test so the special case cannot swallow the breakdown.

### My own probes against the sweep — 17 run

Beyond P2/P5/P16 above, the sweep behaved correctly on everything I threw at it:

```
P3  disable inside an if-block, dialog after         CAUGHT
P4  disable in the body, dialog in an awaited IIFE   CAUGHT
P9  class method                                     CAUGHT
P10 module top level (no function at all)            CAUGHT
P11 this.deleteBtn.disabled = true                   CAUGHT  (receiver text preserved in the message)
P14 confirmDialog?.({ … })                           CAUGHT
P17 disable inside a for-of, dialog after            CAUGHT
P7  disable inside a nested helper                   declined — the named "hidden behind a helper" gap
P6  btn.disabled = !!rows.length                     declined — the named "computed value" gap
P8  const p = confirmDialog(…); disable; await p     declined — contrived, not worth naming
P12 wrapper fixed point over-includes upward         by design, and in the safe direction
```

The fixed point on the real files resolves to `{confirmDialog}`, `{confirmDialog, deleteAll}`, `{confirmDialog, confirmModal, handleBulkDelete}` — the wrapper hop is live in the tree exactly as claimed, and the over-inclusion only ever adds sites to look at.

---

## What I could NOT verify

1. **No delete has run against a real Salesforce org**, and no debug-log delete either. happy-dom plus a mocked `SalesforceApiClient` throughout.
2. **No unpacked-build pass in a real Chrome.** Manifest byte-identity, bundle string census, and static inspection only.
3. **The focus claim in §3 is a real-browser fact asserted from happy-dom, and it stays that way after my review.** I agree with it on the specification, and M-B2 proves the restore is load-bearing *given* that behaviour — but happy-dom does not model blur-on-disable, the test writes that step by hand, and I did not watch it in a browser either. If the claim is wrong about Chrome, the `hadFocus` branch is dead code rather than a bug; if it is right, it is the fix. Neither of us has seen it.
4. **Shadow-DOM retargeting.** The injected UI mounts in a shadow root, where `doc.activeElement` returns the host, not the button. Both `confirm-dialog.ts:65` and the new `hadFocus` check read the same expression, so they cannot disagree — but whether either sees the trigger in the real mounted surface is untested on both sides of this PR, and would make the whole focus-restore story a no-op rather than a regression.
5. **Real assistive-technology behaviour.** Every focus finding here is a DOM `activeElement` assertion; no screen reader.
6. **My B1 finding is a source-level demonstration**, not a user report. The hoisted-callback shape is not in the tree today; the tree is clean under both the shipped rule and my proposed one. What I have shown is that the guard would not catch it *tomorrow*.
7. **The named-open shapes remain open under my proposed one-liner too** — it closes the hoisted callback and nothing else. A disable behind a helper, a computed value, an un-awaited callback and an imported wrapper are all still real ways to write the bug.

---

Take the one-line condition (or name the fifth shape) and I will re-approve on sight. The rest of this — the cancel affect, the `debug-log-viewer` split, the focus catch you found that #326's write-up missed, six dead survivors, and an exclusion list that is empty for a reason you can demonstrate — is the strongest work in this series, and the M-G result in particular is the first time the masking has been shown to be load-bearing rather than inherited. Not merging; that is the owner's call.

---

scoobydrew83 (commented):
## Round-2 review — C-FIX-5 · **VERDICT: APPROVE**

> Posted as a `COMMENT` review only because GitHub refuses `APPROVE` on a PR owned by the same account. **Treat this as APPROVE.** It holds.

Fresh detached worktree at `2d943a4`. Scoped to the delta (`extension/CHANGELOG.md` + `test/confirm-dialog-trigger-contract.test.ts`) and to everything the delta could have moved. **My block is fixed, and it is fixed for the right reason** — I reproduced the hole under the reverted condition and watched it close.

No new blocking finding. One nit worth writing down and one correction to a justification, neither of which should hold the merge.

---

## 1. The block is closed, and the proof is the right proof

**Your reproduction, reproduced.** Hoisted B1 regression live on the real `soql-runner.ts`, condition reverted to the text range:

```
extension/test/confirm-dialog-trigger-contract.test.ts -t "no caller disables"
    Tests  1 passed | 16 skipped (17)          ← the tree-wide sweep PASSES
```

That is my reported hole, on the real file, exactly. Flip the condition back to `a.start > disable.at` and it fails. This is the distinction I care about: you did not just make a test go red, you showed the test goes green under the old condition and red under the new one **with the same defect present**. That is the fix, not something adjacent to it.

**Both regression forms, on the real file, 4 failures each:**

```
(a) callback INLINE                          (b) callback HOISTED to a const
  × no caller disables its trigger …           × no caller disables its trigger …
  × returns focus … when cancelled             × returns focus … when cancelled
  × returns focus … dismissed with Escape      × returns focus … dismissed with Escape
  × leaves the trigger enabled while the       × leaves the trigger enabled while the
    dialog is up, and disables it only           dialog is up, and disables it only
    once deleting starts                         once deleting starts
```

**New mutations:**

| Mutation | Claimed | Mine |
| --- | --- | --- |
| **M-J** text range | fails `detects the B1 shape` | ✅ **2** — `detects the B1 shape` **and** `declines a listener registered in the same body` |
| **M-K** awaits anything | fails `declines a listener…` | ✅ **1**, exactly that one — the `asyncRender` decline is the *only* thing standing between M-K and green |
| **M-C2** condition dropped | 3 fail | ✅ **3** |

**Offender set unchanged in all three shapes**, re-measured rather than carried over:

```
develop shape + condition dropped   →  2   debug-log-viewer.ts:484, flow-version-manager.ts:187
develop shape + shipped condition   →  1   debug-log-viewer.ts:484
PR head       + shipped condition   →  0
```

### The `flow-version-manager` decline is not accidental — I made it fire

You asked to be checked on this and it is the right thing to check. I inserted **one `await` after the disable** in the real `ensureToolbarButton()` (made it async, `void`-called at its one call site) and changed nothing else:

```
features/flow-version-manager.ts: btn.disabled = true (line 187) before handleBulkDelete(
  × no caller disables its trigger on the way into the dialog
```

So the decline is genuinely "this body awaits nothing", the reason you state, and not a coincidence of names or shape.

### The M-K decline/mirror pair — attacked as asked

This is where a rule can look discriminating while keying on something else, so I did not accept the A/B. **Every ordering permutation**, one body, three statements:

```
await, disable, listen            => 0   ← the decline case
disable, await, listen            => 1   ← the mirror
await, disable, listen, await     => 1   ← an await on both sides: correctly caught
disable, listen, await            => 1   ← listener registered before the await: correctly caught
listen, disable, await            => 0   ← lexical order declines it (see the nit)
listen, await, disable            => 0
await, await, disable, listen     => 0   ← two awaits, both before
disable, listen  (no await)       => 0   ← the named-open shape
```

Monotone in exactly the one variable, with no cliff anywhere else.

**And the decline is not keyed on incidentals.** Same ordering, everything else changed — all still `0`:

```
renamed every identifier                          => 0
no `void`, a bare function reference              => 0
function expression instead of an arrow           => 0
setTimeout instead of addEventListener            => 0
```

**And the mirror goes through the branch under test.** The dialog-reaching call sits inside the listener arrow, so `innermost(call)` is the arrow and not `render` — it cannot be caught by the same-body `from === opened` path. Confirmed structurally, and M-E (drop the enclosing case) now fails `declines a listener registered in the same body` too, which it did not at `e22e34a`.

M-K surviving your own round and then being closed is the best single thing in this delta. A condition that had quietly widened to "this body awaits at all" would have passed every other assertion in the file.

## 2. The two rarer spellings — caught, and the zero is a real measurement

Both closed, and the precision is better than it needed to be:

```
btn['disabled'] = true                        => 1   caught
btn.setAttribute('disabled', '')              => 1   caught  (double-quoted)
btn.setAttribute('disabled', '')              => 1   caught  (single-quoted)
btn.setAttribute(`disabled`, '')              => 1   caught  (no-substitution template)
setAttribute in the enclosing body,
  dialog in a hoisted callback                => 1   caught  (composes with the new condition)

btn.setAttribute('aria-disabled', 'true')     => 0   correctly NOT a disable
btn.removeAttribute('disabled')               => 0   correctly NOT a disable
```

Still open and unnamed, both marginal enough that I would not add them: a computed attribute name (`setAttribute(NAME, …)`) and `toggleAttribute('disabled')`.

**One precision note on the mutation evidence.** M-L (drop element access) and M-M (drop `setAttribute`) each fail **`the two rarer disable spellings cost nothing, and that is measured`** — and **not** the AST↔mask agreement. They cannot fail the agreement: there are zero occurrences in the tree, so removing a spelling from the AST changes no count on either side. The three-regex widening of the agreement is correct but **vacuous today**; the measurement test is the one with teeth. That is exactly the structure you built it for — the assertion exists so the zero is a measurement rather than blindness — so this is a confirmation, not a finding. Worth stating plainly so nobody later mistakes the agreement for the guard on those two.

## 3. All 11 prior sweep mutations still bite — and several bite harder

Re-run after the reformat, not spot-checked:

```
M-D  ignore lexical order        × declines the correct order                     (1, was 1)
M-E  same-body only              × detects the B1 shape · declines a listener…    (2, was 1)
M-F  drop the fixed point        × finds every caller · declines a listener… ·
                                   follows a local wrapper                        (3, was 2)
M-G  raw text instead of AST     × no caller disables · finds every caller ·
                                   the dialog's own caller contract is prose ·
                                   AST/mask agreement · the two rarer spellings…  (5, was 4)
M-H  drop the fail-closed gate   × refuses to answer on source no scanner can read (1)
M-I  walk stops at arrows        × finds every caller · detects the B1 shape ·
                                   declines a listener… · AST/mask agreement      (4, was 3)
M-A2 revert debug-log-viewer     × 3, unchanged
M-B2 focus restore never fires   × 1, unchanged
M-J / M-K / M-C2                 × 2 / 1 / 3
```

Nothing weakened. The widened probe set picked up four mutations it did not previously see.

## 4. Gates — re-measured, both sides

| Gate | claimed | mine |
| --- | --- | --- |
| ext `npx vitest run` | 116 / **2178** | ✅ 116 files / 2178 pass |
| root `npm test` | 320 / **5534** | ✅ 320 / 5534 pass |
| ext `npx tsc --noEmit` | clean | ✅ exit 0 |
| ext `npx eslint .` | clean | ✅ exit 0 |
| ext `npx wxt build` | exit 0, Σ 1.41 MB | ✅ exit 0, Σ 1.41 MB |
| `check:all-contracts` / `generate:catalogs` | exit 0 / no drift | ✅ / ✅ |

**Bundle byte-identity — stronger than claimed and verified as such.** I did not just compare the manifest hash; I diffed the whole build tree against my round-1 `.output/chrome-mv3` from `e22e34a`:

```
diff -r --brief <e22e34a build> <2d943a4 build>   →   no differences
manifest sha256  2a302f6a1b97539b0857f162e24fb498774b602f8840b515d865ecc9e8359c7d   (unchanged)
content.js 634.33 kB · node-graph 241.61 kB   (unchanged)
```

Every emitted file is byte-identical. Round 2 touches a test file and prose only, and the artifact proves it.

**Body corrections both verified.** Seven already-unclean Prettier files including `CHANGELOG.md`, complaint set byte-identical to base, new file clean. Binary-safe re-count in the shipped chunks: `"SFDT cannot confirm it reached your disk"` **2×** per chunk (new tooltip + pre-existing dialog copy), `"Delete stopped by the new query"` **1×**, `"Counting logs"` **1×**.

**CHANGELOG.** One bullet reworded — the sweep description, now saying "flagged when that body goes on to `await` something — a control-flow fact, deliberately not 'the call sits inside the await's text'". 32 bullets both sides, same three headings, house order. Accurate to what shipped.

---

## Nits

### N1 — the same hoist, placed *above* the disable, is still declined

The order rule keys on where the dialog-reaching call site sits textually. Move the `const` up two lines and the sweep goes quiet again:

```ts
async function start(btn) {
  const deps = { confirm: () => confirmDialog({ doc }) };   // hoisted ABOVE
  btn.disabled = true;
  await runBulkDelete(rows, deps);
}
// => []          (the same thing hoisted BELOW the disable => 1, pinned by this PR)
```

Same bug, same `await`, same callback running with the trigger disabled — only the declaration moved.

**I am not blocking on this, and I want to be explicit about why.** It is a property of the **lexical order** rule, which predates this fix, is untouched by it, and which I probed and accepted in round 1 (my P8). It is not something the new condition introduced or worsened, and calling it a block after you took my condition verbatim would be moving the goalposts.

Since I have it measured, the refinement if you ever want it: **apply the order check only to the same-body case**, where the dialog really is evaluated at that point. A callback's *declaration* position says nothing about when it runs, so for the enclosing case the await position is the whole story:

```ts
const opened = innermost(scan, call.at);
if (from === opened) return disable.at < call.at;   // order matters only here
if (!encloses(from, opened)) return false;
return scan.awaits.some((a) => a.inFn === from && a.start > disable.at);
```

Measured: **all 17 of the sweep's tests still green**, plus all 24 of my probe assertions (the correct-order decline, both sibling orders, the real `flow-version-manager.ts`, the whole M-K permutation table), tree still clean over 125 files, and the hoist-above case goes `0 → 1`. Hoisting to *module* scope stays open either way — that one genuinely needs cross-scope reasoning. Follow-up sized, or a line in the named-open list. Your call, not a condition.

### N2 — the `btn.disabled = busy` justification is right but the stated reason is not the measured one

I agree with the decision. I do not agree with *"claiming it would flag correct code all over the tree"*, because I measured it and it does not — today.

Widening the RHS from `TrueKeyword` to "anything but `false`":

```
125 files · 55 literal-true disables · 21 computed-RHS disables in 11 files
tree-wide offender test under the widened rule:  STILL PASSES — zero new offenders
what actually breaks: the AST↔mask agreement (the mask regexes still only match `= true`)
                      and the named-open assertion for the computed shape
```

The 21 are all state syncs of exactly the kind you name — `toolbarBtn.disabled = count === 0`, `restBtn.disabled = sosl`, `deployBtn.disabled = fields.length === 0 || !hasValid`, and `confirmBtn.disabled = typedInput!.value.trim() !== word` in the dialog itself. **The better reason to keep it out is unknowability, not current noise**: `x.disabled = expr` is not knowably a disable — the same statement enables the control half the time — so a rule that flagged it would be asserting something it cannot determine, and the day one of those 21 lands in a body that disables-then-awaits-then-reaches a dialog there would be no way to tell it from a real defect. Same conclusion, a reason that survives the tree changing. Worth a sentence swap in the doc comment.

### N3 — the declared over-approximation, quantified

You name it: a body that disables, awaits something unrelated, and only then registers a listener is flagged. I confirmed it fires on real code — the `ensureToolbarButton()` experiment above *is* that false positive, manufactured. For scale: **50 of the tree's disables have a later `await` somewhere in the same file**; none currently combines with an enclosing dialog-reaching call, which is why the tree is at zero. It fires in the safe direction and produces a site to look at, so I would take this trade every time — it is the one I proposed. Recording the number so the next round has it.

---

## What I could NOT verify

Unchanged from round 1, and the standing two still stand:

1. **No delete has run against a real Salesforce org**, and no debug-log delete. happy-dom plus a mocked `SalesforceApiClient` throughout.
2. **No unpacked-build pass in a real Chrome.** Byte-identity of the whole build tree against round 1 is stronger evidence than I had before, but it is still static.
3. **The blur-on-disable claim in §3 remains a real-browser fact asserted from happy-dom.** Untouched by this round, and neither of us has watched it in a browser.
4. **Shadow-DOM retargeting** of `doc.activeElement` — untested on both sides, and it would make the focus-restore story a no-op rather than a regression.
5. **Real assistive-technology behaviour.** DOM `activeElement` assertions only.
6. **N1 is a source-level demonstration, not a live defect.** The tree is clean under the shipped rule and under my refinement; what I showed is a shape the guard would not catch tomorrow, in a family that also still contains the module-scope hoist, a disable behind a helper, a computed RHS, an un-awaited callback, and an imported wrapper. A source sweep is a net, not a proof.
7. **I did not re-verify** the items I confirmed at `e22e34a` and the delta cannot reach: the 21 original mutations, `status: 'failed'` and its single consumer, the `debug-log-viewer` split, `superseded`, principle #12's scope, and the CHANGELOG bullet parse. The bundle byte-identity is my evidence that the runtime half is untouched.

---

**Approving.** You took the condition verbatim, proved it was the fix by reproducing the hole under the old one, went looking for what your own change might have widened, found M-K yourself, and closed the two spellings I had written off as not worth naming. The M-K pair in particular is the thing that makes this a control-flow rule rather than a rule that happens to be right — a decline case whose mirror is caught, attacked from eight orderings and four rewrites, is a real discriminator. Not merging; that is the owner's call.

---

